PR #32, opened 2019-01-07
feat: project module

### LINKED ISSUE #4: Working Project module
The business logic domain for the current opened project:

### Settings reducer
- Project name
- Parcels
(And any other scene.json fields that are relevant)

### Scene reducer
- Scene tree (simplified nodes)

NicoSantangelo:
Closes #4


src/modules/project/reducer.ts
```diff
@@ -0,0 +1,36 @@
+import { loadingReducer, LoadingState } from 'decentraland-dapps/dist/modules/loading/reducer'
+import { ModelById } from 'decentraland-dapps/dist/lib/types'
+import { Project } from 'modules/project/types'
+import { CreateProjectAction, CREATE_PROJECT } from 'modules/project/actions'
+
+export type ProjectState = {
```
belohlavek: Shouldn't this go in `project/types.ts`?

NicoSantangelo: We only keep `XXXState` types on the reducer for two things:
- It's close to it's use
- Circular dependencies!

belohlavek: Ok, I feel I asked about it before

PR #52, opened 2019-01-14
feat: UNDO/REDO keyboard shortcuts

### LINKED ISSUE #6: Undo/Redo
We need the client to report the simplified node that was changed together with a path to that specific node. Then, we can store the reference to previous redux states to perform time travel undo/redo.

NicoSantangelo:
Closes #6


src/modules/editor/sagas.ts
```diff
@@ -1,14 +1,36 @@
...
+  const shortcuts = getKeyboardShortcuts()
+  yield put(bindKeyboardShortcuts(shortcuts))
+}
+
+function* handleUnbindEditorKeyboardShortcuts() {
+  const shortcuts = getKeyboardShortcuts()
+  yield put(unbindKeyboardShortcuts(shortcuts))
+}
+
+function getKeyboardShortcuts(): KeyboardShortcut[] {
```
belohlavek: Should this be in a separate file? `shortcuts.ts` or..... `helpers.ts` :P

PR #71, opened 2019-01-23
feat: camera

### LINKED ISSUE #56: Basic editor camera orientation
We should change the basic editor camera to look up from above. No need for complex rotation features for this one.

belohlavek:
5c48ad256710f034264437

- Rotate camer (drag)
- Zoom (wheel or two finger gesture or pinch)
- Move (arrow keys)

Closes #56


src/components/Preview/Preview.tsx
```diff
@@ -4,6 +4,8 @@ import './Preview.css'
 
 const editorWindow = window as EditorWindow
 
+const CONTENT_SERVER = 'https://content.decentraland.today/contents/'
```
NicoSantangelo: maybe we should use an env var here?

src/modules/editor/sagas.ts
```diff
@@ -106,6 +106,8 @@ function* handleUpdateEditor(action: UpdateEditorAction) {
     }
   }
 
+  console.log(msg)
```
NicoSantangelo: Whops

PR #77, opened 2019-01-28
feat: remove flickering

### LINKED ISSUE #76: Disable editor update flickering

menduz:
- There is now a scene file we build and use as a middleman between the editor and the client
- Updates are handles by this scene, which takes entities and components and spawns them without needing to reboot the scene (no flicker)
- We don't need to "generate js" anymore (at least until export to cli/deployment), we just send an object to the scene
- Said scene file needs to be built beforehand as it is loaded using `raw-loader`. Basically it needs to be a separate build.

Closes #76


package.json
```diff
@@ -79,12 +79,12 @@
...
   "scripts": {
     "start": "npm run build:scene && concurrently -n \"scene,builder\" \"decentraland-compiler buildScene.json --watch\" \"node scripts/start.js\"",
-    "build": "npm run build:scene && node scripts/build.js",
     "build:scene": "decentraland-compiler buildScene.json",
+    "build": "npm run build:scene && node scripts/build.js",
     "build:ci": "npm run lint && npm run build:scene && node scripts/build.js",
     "test": "node scripts/test.js",
     "lint": "tslint --project tsconfig.json -e '*.json' -c tslint.json 'src/**/*.ts{,x}'",
-    "bundle:svg": "npx svg-sprite -D src/components/Icon --css-dimensions \"\" --cs icons.svg --css --css-dest . --css-bust false --css-prefix \".Icon.%s\" --css-render-css src/icons/*.svg"
+    "bundle:svg": "npx svg-sprite -D src/components/llllIcon --css-dimensions \"\" --cs icons.svg --css --css-dest . --css-bust false --css-prefix \".Icon.%s\" --css-render-css src/icons/*.svg"
```
menduz: @belohlavek is this a typo?

belohlavek: yup

PR #81, opened 2019-01-28
style: colors

cazala:
This fixes the background color and uses the new decentraland-ui which already has the right primary color. Removes temporarily the line from the navbar via CSS until we can do that from decentraland-dapps. It also uses the prevew background for the placeholder while the client is loading.


belohlavek:
Looking good, @nicosantangelo please check this one

nicosantangelo:
💯

Issue #58, opened 2019-01-18
Delete action button


belohlavek:
The topbar should include a button that deletes the current selected item

Issue #57, opened 2019-01-18
Duplicate Action button


belohlavek:
The topbar should include a button that duplicates the current selected item

Issue #46, opened 2019-01-14
Topbar modes buttons


belohlavek:
<img width="200" alt="screen shot 2019-01-14 at 10 01 53" src="https://user-images.githubusercontent.com/1307029/51114239-70dc1f80-17e3-11e9-88b1-fd34e201b9c9.png">

Should include:
- Select mode
- Translation mode
- Rotation mode

Depends on #7

PR #83, opened 2019-01-30
feat: toggle preview and sidebar

### LINKED ISSUE #48: Topbar toggle buttons
screen shot 2019-01-14 at 10 11 42

These two buttons should: 
- Toggle play/edit mode
- Toggle sidebar visibility

All functionality implementation is required for this issue as buttons should already be implemented by #46 and the sidebar should be already implemented by #30, lastly the play mode should already be implemented by #12

cazala:
I added the editor mode (move/rotate), editor actions (undo/delete/duplicate) and toggle actions (preview/sidebar) to the `<TopBar>` component, and I added an editor state/reducer.

The move/rotate buttons are wired but the sagas still need to interact with the client (i left a TODO there)

The delete/duplicate are disabled until we can get the selected entity to the editor state (I left an action and I wired the state so the buttons are disabled)

Undo button works ✅ 

Toggle preview works ✅ 

Toggle sidebar works ✅ 

Closes #48 

Opens the door for #46 #57 #58 we just need to wire some stuff to the client

preview-sidebar


src/modules/editor/actions.ts
```diff
@@ -51,3 +51,35 @@ export const editorRedo = () => action(EDITOR_REDO, {})
...
+
+export const toggleSidebar = (enabled: boolean) => action(TOGGLE_SIDEBAR, { enabled })
+
+export type ToggleSidebarAction = ReturnType<typeof toggleSidebar>
+
+// Select Entity
+
+export const SELECT_ENTITY = 'Select entity'
+
+export const selectEntity = (entityId: string) => action(SELECT_ENTITY, { entityId })
```
nicosantangelo: Is it going to be possible to deselect an entity?

src/components/ViewPort/ViewPort.tsx
```diff
@@ -2,10 +2,24 @@ import * as React from 'react'
 import { Props } from './ViewPort.types'
 import Preview from 'components/Preview'
 import './ViewPort.css'
+import Icon from 'components/Icon'
 
 export default class ViewPort extends React.PureComponent<Props> {
+  handleClose = () => {
+    const { onTogglePreview } = this.props
+    onTogglePreview(false)
+  }
```
nicosantangelo: Maybe we can add a newline after this `}` to keep it consistent

src/components/ViewPort/ViewPort.tsx
```diff
@@ -2,10 +2,24 @@ import * as React from 'react'
 import { Props } from './ViewPort.types'
 import Preview from 'components/Preview'
 import './ViewPort.css'
+import Icon from 'components/Icon'
```
nicosantangelo: Import order ⚠️

src/components/ViewPort/ViewPort.tsx
```diff
@@ -2,10 +2,24 @@ import * as React from 'react'
...
 
 export default class ViewPort extends React.PureComponent<Props> {
+  handleClose = () => {
+    const { onTogglePreview } = this.props
+    onTogglePreview(false)
```
nicosantangelo: Is this `false` necessary? I think the container does not receive the value and hardcodes `false`

Maybe we can change the name on the container to `onClosePreview`?

src/components/TopBar/TopBar.types.ts
```diff
@@ -1,8 +1,30 @@
...
+  setMode,
+  togglePreview,
+  toggleSidebar,
+  SetModeAction,
+  TogglePreviewAction,
+  ToggleSidebarAction,
+  editorUndo,
+  EditorUndoAction
+} from 'modules/editor/actions'
+import { Dispatch } from 'redux'
```
nicosantangelo: Import order ⚠️

src/components/SideBar/ItemDrawer/ItemDrawer.css
```diff
@@ -47,6 +47,6 @@
 }
 
 .ItemDrawer .overflow-container {
-  height: 100%;
+  height: calc(100% - 54px);
```
nicosantangelo: What is this `54` accounting for? should we have a CSS variable in case it changes?

src/components/EditorPage/EditorPage.container.ts
```diff
@@ -2,10 +2,15 @@ import { connect } from 'react-redux'
 
 import { loadAssetPacksRequest } from 'modules/assetPack/actions'
 import { bindEditorKeyboardShortcuts, unbindEditorKeyboardShortcuts, startEditor, closeEditor } from 'modules/editor/actions'
-import { MapDispatch, MapDispatchProps } from './EditorPage.types'
+import { MapStateProps, MapDispatch, MapDispatchProps } from './EditorPage.types'
 import EditorPage from './EditorPage'
+import { RootState } from 'modules/common/types'
+import { isSidebarOpen, isPreviewing } from 'modules/editor/selectors'
```
nicosantangelo: Import order ⚠️

src/modules/editor/actions.ts
cazala: Good question, I don't know how the client works on that, but I suppose I could let the `entityId` also be `null` so we can `selectEntity(null)`

src/components/ViewPort/ViewPort.tsx
cazala: It's because the type of `onTogglePreview` is `typeof togglePreview`:

```
export type Props = {
  isPreviewing: boolean
  onTogglePreview: typeof togglePreview
}
```

That's the pattern we've been using on this project from what I saw (always typing callbacks with the `typeof` of some action)

I could add a default value of `false` to `togglePreview` so we don't have to pass that `false` and then I can rename it to `onClosePreview`

src/modules/editor/actions.ts
nicosantangelo: Yes! or we could have a `deselectEntity()` with no params that calls `action(DESELECT_ENTITY, { entityId: null })`

src/components/SideBar/ItemDrawer/ItemDrawer.css
cazala: It's the "Item drawer" header + list/grid switcher. I could store that height in a variable but I don't think we'd use it anywhere else than on this line.

src/components/ViewPort/ViewPort.tsx
nicosantangelo: Makes sense!, wasn't sure looking at the code but I thought that was the reason Typescript didn't yell about the `false`

I think adding a default is an option, but maybe we could make an exception here and use something like:

`onClosePreview: () => ReturnType<typeof togglePreview>`

for the type. What do you think?

cazala: done ✅

src/modules/editor/actions.ts
cazala: done ✅

PR #94, opened 2019-01-31
refactor: make use of the Link component

nicosantangelo:


src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -1,33 +1,42 @@
...
   }
 
   render() {
-    const { project } = this.props
+    const { project, onClick } = this.props
 
-    return (
-      <div className="ProjectCard Card" onClick={this.handleOnClick}>
-        {/*<img src={project.thumbnail} alt={project.title} width={248} height={184} />*/}
+    // <img src={project.thumbnail} alt={project.title} width={248} height={184} />
```
cazala: Why is this commented?

nicosantangelo: It was there just as a reminder that we needed to add a thumbnail in the future, but I can remove it!

cazala: cool, leave it 👍

PR #106, opened 2019-02-01
feat: derive categories from state

### LINKED ISSUE #61: Refactor category domain
Categories can be derived from already existing state and will make filtering/searching easier

cazala:
Closes #61 

In this PR I modified the `getCategories` selector to derive the `FullCategory[]` array from the state instead of storing it in the category reducer. I also added the ability to filter by `assetPackId` and by a search string that is compared against asset name, category or tags.

I left the redux actions in case we want to wire it to the UI in the future.


src/modules/category/selectors.ts
```diff
@@ -1,24 +1,74 @@
...
+import { getData as getAssetPacks } from 'modules/assetPack/selectors'
 import { getData as getAssets } from 'modules/asset/selectors'
+import { AssetState } from 'modules/asset/reducer'
+import { AssetPackState } from 'modules/assetPack/reducer'
+import { Asset } from 'modules/asset/types'
 
 export const getState: (state: RootState) => CategoryState = state => state.category
 
-export const getData: (state: RootState) => CategoryState['data'] = state => getState(state).data
+export const getSelectedAssetPackId = (state: RootState) => getState(state).selectedAssetPackId
```
nicosantangelo: I don't get why is this here, wouldn't this be a responsibility of `assetPack`?

nicosantangelo: I don't get why is this here, wouldn't this be a responsibility of assetPack?

src/modules/category/selectors.ts
```diff
@@ -1,24 +1,74 @@
...
 
-export const isLoading: (state: RootState) => boolean = state => getState(state).loading.length > 0
+export const getSearch = (state: RootState) => getState(state).search
 
-export const getError: (state: RootState) => CategoryState['error'] = state => getState(state).error
+const isSearchResult = (asset: Asset, search: string) => {
```
nicosantangelo: I don't get why is this here, it's not checking _only_ the category, so wouldn't this be a responsibility of `assetPack`? or maybe `ui/search` ?

src/modules/category/selectors.ts
```diff
@@ -1,24 +1,74 @@
...
+            }
+          }
+          // add asset to category
+          categories[asset.category].assets.push(asset)
+        }
+      }
+    }
+
+    // convert map to array
+    return Object.keys(categories).map(categoryName => categories[categoryName])
```
nicosantangelo: `Object.values`?

src/modules/category/selectors.ts
belohlavek: If i recall correctly the idea was to leave the possibility to choose a single asset pack at a time, while `null` would merge all assets packs together by category.

This is one of those things that might get rolled back in the future depending on the direction we take. What do you think?

cazala: Maybe `category` is not the right name anymore, but the responsibility of this module is to generate the `FullCategory[]` array that we use to render the item drawer. 

We could rename it to `ui` or `sidebar` or something else, but i think the selected asset pack id and search string belong to a separate module rather that `assetPack`.

src/modules/category/selectors.ts
cazala: https://github.com/decentraland/builder/pull/106#discussion_r253470051

src/modules/category/selectors.ts
nicosantangelo: > We could rename it to ui or sidebar or something else, but i think the selected asset pack id and search string belong to a separate module rather that assetPack.

Agreed

> Maybe category is not the right name anymore, but the responsibility of this module is to generate the FullCategory[] array that we use to render the item drawer.

Think so, I think the only problem here (if there's one) is the name of the domain. I like both UI and sidebar

cazala:
> I'm not sure I understand the search and selected actions, but super good otherwise

@nicosantangelo this is how the search action would work if wired to an input (not part of this PR, just to illustrate)

search

And the `selectAssetId` could be connected to a dropdown to filter by asset pack (ie. All,  Space, Western, Medieval, etc)

src/modules/category/selectors.ts
cazala: refactored into `ui.sidebar`

nicosantangelo:
Awesomeeeee

PR #113, opened 2019-02-04
feat: wait for editor to be ready

### LINKED ISSUE #102: Dispatch action on engine init
Will help to know when the scene is ready to be interacted with

belohlavek:
kapture 2019-02-04 at 15 46 01

Closes #102

### DIFF
--- a/src/components/Preview/Preview.tsx
+++ b/src/components/Preview/Preview.tsx
@@ -10,7 +10,7 @@ export default class Preview extends React.Component<Props, State> {
   canvas = React.createRef<HTMLDivElement>()
 
   componentDidMount() {
-    this.startEditor()
+    this.startEditor().catch(error => console.error('Failed to start editor', error)
   }
 
   async startEditor() {
--- a/src/components/Preview/Preview.tsx
+++ b/src/components/Preview/Preview.tsx
@@ -20,9 +20,6 @@ export default class Preview extends React.Component<Props, State> {
       const canvas = await editorWindow.editor.getDCLCanvas()
 
       if (this.canvas.current && canvas) {
-        this.setState({
-          isLoading: false
-        })
         this.canvas.current!.appendChild(canvas)
         editorWindow.editor.resize()
         this.props.onOpenEditor()
--- a/src/components/Preview/Preview.tsx
+++ b/src/components/Preview/Preview.tsx
@@ -1,48 +1,47 @@
 import * as React from 'react'
 import { Loader } from 'decentraland-ui'
 
-import { EditorWindow } from './Preview.types'
+import { EditorWindow, Props, State } from './Preview.types'
 import './Preview.css'
 
 const editorWindow = window as EditorWindow
 
-export default class Preview extends React.Component {
-  private canvas = React.createRef<HTMLDivElement>()
+export default class Preview extends React.Component<Props, State> {
+  canvas = React.createRef<HTMLDivElement>()
 
   componentDidMount() {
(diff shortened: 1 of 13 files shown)


src/modules/editor/selectors.ts
```diff
@@ -5,3 +5,4 @@ export const getEditorMode = (state: RootState) => getState(state).mode
 export const isSidebarOpen = (state: RootState) => getState(state).sidebar
 export const isPreviewing = (state: RootState) => getState(state).preview
 export const getSelectedEntityId = (state: RootState) => getState(state).selectedEntityId
+export const getReadyState = (state: RootState) => !getState(state).isReady
```
nicosantangelo: I find this naming confusing. It's circumstance but having a `getState(state).isReady` that returns the `ReadyState` is too many states for me.
Plus, we're trying to append `is` to boolean functions, so do you think we can rename it to `isReady`?

src/modules/editor/reducer.ts
```diff
@@ -37,32 +50,50 @@ export const editorReducer = (state = INITIAL_STATE, action: EditorReducerAction
...
+        isReady: true
+      }
+    }
+
+    case CLOSE_EDITOR: {
+      return {
+        ...INITIAL_STATE
+      }
+    }
+
```
nicosantangelo: I think this is fine but this would be the only file with newlines after the `}`. I'm annoying but I think it might be better to remove them for consistency sake

src/modules/editor/actions.ts
```diff
@@ -108,3 +116,11 @@ export const RESET_CAMERA = 'Reset camera'
 export const resetCamera = () => action(RESET_CAMERA, {})
 
 export type ResetCameraAction = ReturnType<typeof resetCamera>
+
+// Editor ready
+
+export const EDITOR_READY = 'Editor ready'
+
+export const editorReady = () => action(EDITOR_READY, {})
```
nicosantangelo: I think this name is backwards, the verb should be first, like `closeEditor`

src/components/Preview/Preview.tsx
```diff
@@ -1,48 +1,47 @@
...
-import { EditorWindow } from './Preview.types'
+import { EditorWindow, Props, State } from './Preview.types'
 import './Preview.css'
 
 const editorWindow = window as EditorWindow
 
-export default class Preview extends React.Component {
-  private canvas = React.createRef<HTMLDivElement>()
+export default class Preview extends React.Component<Props, State> {
+  canvas = React.createRef<HTMLDivElement>()
```
nicosantangelo: Shouldn't we initialize the state here?

src/components/Preview/Preview.tsx
```diff
@@ -1,48 +1,47 @@
...
-  moveCanvas = (canvas: HTMLCanvasElement) => {
-    if (this.canvas.current && canvas) {
-      this.canvas.current.appendChild(canvas)
-      editorWindow.editor.initEngine()
-      editorWindow.editor.resize()
-    }
-  }
-
   render() {
+    const { isLoading } = this.props
```
nicosantangelo: I'm not sure I understand the Props/State for this Component. If we're getting isLoading as prop, why are we updating the state on `startEditor`?

belohlavek: due to the enablement of dani's chot mode

src/components/Preview/Preview.tsx
belohlavek: nope

PR #115, opened 2019-02-04
feat: reset/duplicate/delete item

### LINKED ISSUE #46: Topbar modes buttons

### LINKED ISSUE #57: Duplicate Action button

### LINKED ISSUE #58: Delete action button

### LINKED ISSUE #99: Undo button should be transform reset button
screen shot 2019-02-01 at 10 23 51

Originally in the designs this was meant to be a en entity reset button for that entity's Transform.

We need to discuss wether or not we need an undo button, and if so we should then add a redo button.

cazala:
Closes #46 
Closes #57 
Closes #58 
Closes #99 

I refactored the scene actions so they are all handled by sagas and in the end they dispatch a `provisionScene` action with the whole scene. I did this so in order to have a single action that writes the scene to the app state (this is useful for undo/redo since we just have to track one action) also it makes it easier to listen for changes and send them to the `ecsScene`, we don't have to listen to several actions.

I also did some renaming: 

+ Set Mode -> Set Gizmo

+ Add Asset -> Add Item (to match all the other item-related actions like reset/duplicate/delete, also it feels more accurate since adding an asset would be adding a new model to the available ones, rather than adding an instance of that model to the scene)

### DIFF
--- a/src/components/Modals/ShortcutsModal/ShortcutsModal.tsx
+++ b/src/components/Modals/ShortcutsModal/ShortcutsModal.tsx
@@ -74,6 +74,15 @@ export default class ShortcutsModal extends React.PureComponent<Props> {
               </div>
 
               <div className="shortcut">
+                <div className="name">{t('shortcuts_modal.basic_controls.reset')}</div>
+                <div className="keybinding">
+                  <Chip text={this.getMainShortcutKey()} />
+                  <span className="plus">+</span>
+                  <Chip text="s" />
+                </div>
+              </div>
+
+              <div className="shortcut">
                 <div className="name">{t('shortcuts_modal.basic_controls.delete')}</div>
                 <div className="keybinding">
                   <Chip text="Delete" />
@@ -112,14 +121,14 @@ export default class ShortcutsModal extends React.PureComponent<Props> {
               <div className="shortcut">
                 <div className="name">{t('shortcuts_modal.edit_controls.move')}</div>
                 <div className="keybinding">
-                  <Chip text="m" />
+                  <Chip text="w" />
                 </div>
               </div>
 
               <div className="shortcut">
                 <div className="name">{t('shortcuts_modal.edit_controls.rotate')}</div>
                 <div className="keybinding">
-                  <Chip text="r" />
+                  <Chip text="e" />
                 </div>
               </div>
             </div>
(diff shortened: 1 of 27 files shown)


src/components/Modals/ShortcutsModal/ShortcutsModal.tsx
```diff
@@ -73,6 +73,15 @@ export default class ShortcutsModal extends React.PureComponent<Props> {
                 </div>
               </div>
 
+              <div className="shortcut">
+                <div className="name">{t('shortcuts_modal.basic_controls.reset')}</div>
+                <div className="keybinding">
+                  <Chip text={this.getMainShortcutKey()} />
+                  <span className="plus">+</span>
+                  <Chip text="s" />
```
belohlavek: CTRL + S reset? Feels like save

src/modules/editor/sagas.ts
```diff
@@ -78,12 +77,32 @@ function* handleUnbindEditorKeyboardShortcuts() {
...
-      callback: () => store.dispatch(setMode('move'))
+      combination: ['w'],
+      callback: () => store.dispatch(setGizmo(Gizmo.MOVE))
+    },
+    {
+      combination: ['e'],
+      callback: () => store.dispatch(setGizmo(Gizmo.ROTATE))
+    },
+    {
+      combination: ['command+s'],
```
belohlavek: missing bindings for windows

src/modules/editor/sagas.ts
```diff
@@ -78,12 +77,32 @@ function* handleUnbindEditorKeyboardShortcuts() {
...
+      callback: () => store.dispatch(resetItem())
+    },
+    {
+      combination: ['command+d'],
```
belohlavek: missing bindings for windows

src/modules/editor/sagas.ts
belohlavek: also see comment about ctrl-s being save in other apps :P

src/modules/translation/languages/en.json
```diff
@@ -32,8 +32,9 @@
       "title": "Basic Controls",
       "undo": "Undo",
       "redo": "Redo",
-      "duplicate": "Duplicate object",
-      "delete": "Delete object",
+      "reset": "Reset item",
+      "duplicate": "Duplicate item",
+      "delete": "Delete item",
```
belohlavek: also update "move object" and "rotate object" below

src/modules/translation/languages/en.json
```diff
@@ -32,8 +32,9 @@
       "title": "Basic Controls",
       "undo": "Undo",
       "redo": "Redo",
-      "duplicate": "Duplicate object",
-      "delete": "Delete object",
+      "reset": "Reset item",
```
belohlavek: May be a better name? @nicosantangelo 
`Reset item transform`
`Reset item rotation`

PR #131, opened 2019-02-05
fix: scrolling on firefox

### LINKED ISSUE #130: Firefox sidebar scrolling

belohlavek:
Firefox:
kapture 2019-02-05 at 18 58 09

Chrome:
kapture 2019-02-05 at 18 59 14

Closes #130

### DIFF
--- a/src/components/EditorPage/EditorPage.tsx
+++ b/src/components/EditorPage/EditorPage.tsx
@@ -15,14 +15,11 @@ export default class EditorPage extends React.PureComponent<Props> {
   componentWillMount() {
     this.props.onLoadAssetPacks()
     this.props.onBindKeyboardShortcuts()
-    document.body.classList.add('lock-scroll')
-    document.body.scrollTop = 0
   }
 
   componentWillUnmount() {
     this.props.onUnbindKeyboardShortcuts()
     this.props.onCloseEditor()
-    document.body.classList.remove('lock-scroll')
   }
 
   handleToolClick = (toolName: ToolName) => {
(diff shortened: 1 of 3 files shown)


src/components/EditorPage/EditorPage.tsx
```diff
@@ -15,14 +15,11 @@ export default class EditorPage extends React.PureComponent<Props> {
   componentWillMount() {
     this.props.onLoadAssetPacks()
     this.props.onBindKeyboardShortcuts()
-    document.body.classList.add('lock-scroll')
```
nicosantangelo: We need to remove the class from the css file too!

src/components/EditorPage/EditorPage.tsx
```diff
@@ -15,14 +15,11 @@ export default class EditorPage extends React.PureComponent<Props> {
...
-    document.body.scrollTop = 0
```
nicosantangelo: I'm not sure if this was preventive or a fix for a particular case. It might still be useful cc: @cazala

belohlavek: Yep that one came in the search PR and was partially removed in the latest fix PR. Now that scrolling should be fixed we can validate that this solution works, previously it jumped around and left the page scroll locked incorrectly (probably due to the other overflow issues).

cazala: Yes, the purpose of the `.lock-scroll` class  was to disable the "elastic scroll" effect on macOS while zooming in and out in the editor (you can see it happening in the second gif on this PR)

belohlavek:
No more bouncing and no more pinch to zoom on chrome and firefox (at least on osx). Safari still sucks.

PR #151, opened 2019-02-07
feat: project thumbnails

### LINKED ISSUE #78: Take screenshots for project thumbnails
on every autosave -> WAMMY -> `canvas.toDataURL()` -> BOOM ->  save within project in redux store -> BANGGg -> read from store and insert in  `` -> KABLAMM

cazala:
screen shot 2019-02-08 at 2 56 47 pm

Closes #78

### DIFF
--- a/src/modules/editor/sagas.ts
+++ b/src/modules/editor/sagas.ts
@@ -225,7 +225,7 @@ function* handleScreenshot(_: ScreenshotAction) {
     if (screenshot) {
       const thumbnail = yield call(() => resizeScreenshot(screenshot, 246, 182))
       if (thumbnail) {
-        const currentProject = yield select(getCurrentProject)
+        const currentProject: Project | null = yield select(getCurrentProject)
         if (currentProject) {
           yield put(editProject(currentProject.id, { thumbnail }))
         }
--- a/src/modules/editor/sagas.ts
+++ b/src/modules/editor/sagas.ts
@@ -21,9 +21,9 @@ import {
   selectEntity,
   EDITOR_REDO,
   EDITOR_UNDO,
-  SCREENSHOT,
-  ScreenshotAction,
-  screenshot
+  TAKE_SCREENSHOT,
+  TakeScreenshotAction,
+  takeScreenshot
 } from 'modules/editor/actions'
 import { PROVISION_SCENE, updateMetrics, updateTransform, DUPLICATE_ITEM, DROP_ITEM, DropItemAction, addItem } from 'modules/scene/actions'
 import { bindKeyboardShortcuts, unbindKeyboardShortcuts } from 'modules/keyboard/actions'
@@ -61,7 +61,7 @@ export function* editorSaga() {
   yield takeLatest(RESET_CAMERA, handleResetCamera)
   yield takeLatest(DUPLICATE_ITEM, handleDuplicateItem)
   yield takeLatest(DROP_ITEM, handleDropItem)
-  yield takeLatest(SCREENSHOT, handleScreenshot)
+  yield takeLatest(TAKE_SCREENSHOT, handleScreenshot)
 }
 
 function* handleBindEditorKeyboardShortcuts() {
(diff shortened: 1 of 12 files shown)


src/modules/editor/sagas.ts
```diff
@@ -205,3 +213,20 @@ function* handleDropItem(action: DropItemAction) {
     yield put(addItem(asset, position))
   }
 }
+
+function* handleScreenshot(_: ScreenshotAction) {
+  try {
+    const screenshot = yield call(() => editorWindow.editor.takeScreenshot())
```
nicosantangelo: **¯\_(ツ)_/¯**

If you want to have one less `if` statement below, you could do

```const [screenshot, currentProject] = yield all([call(() => editorWindow.editor.takeScreenshot()), select(getCurrentProject)])```

It's not _exactly_ the same, as the project could be undefined after asking for the thumb, but in this case I don't see that happening.

File: src/modules/editor/sagas.ts [hunk 2]
```diff
@@ -205,3 +213,20 @@ function* handleDropItem(action: DropItemAction) {
...
+    if (screenshot) {
+      const thumbnail = yield call(() => resizeScreenshot(screenshot, 246, 182))
+      if (thumbnail) {
+        const currentProject = yield select(getCurrentProject)
```
nicosantangelo: `currentProject` is `any` here. We may want to add a `currentProject: Project | null`
We need a better way to type `yield select` tho

src/modules/editor/sagas.ts
```diff
@@ -205,3 +213,20 @@ function* handleDropItem(action: DropItemAction) {
     yield put(addItem(asset, position))
   }
 }
+
+function* handleScreenshot(_: ScreenshotAction) {
+  try {
+    const screenshot = yield call(() => editorWindow.editor.takeScreenshot())
+    if (screenshot) {
+      const thumbnail = yield call(() => resizeScreenshot(screenshot, 246, 182))
+      if (thumbnail) {
```
nicosantangelo: If this is false, shouldn't we retry?

cazala:
> I can see the gizmos on the screenshot, is that ok?

@nicosantangelo yes, that's expected

src/modules/editor/sagas.ts
```diff
@@ -205,3 +213,20 @@ function* handleDropItem(action: DropItemAction) {
...
+        const currentProject = yield select(getCurrentProject)
+        if (currentProject) {
+          yield put(editProject(currentProject.id, { thumbnail }))
+        }
+      }
+    }
+  } catch (e) {
```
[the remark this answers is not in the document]
cazala: Since there's a `delay(500)` before dispatching the screenshot action, it could happen that the user leaves the editor page before the screenshot is taken and potentially throw, that's why I wrapped it in a try/catch

src/modules/editor/sagas.ts
cazala: These `if`s shouldn't be false unless the canvas has been detached/unmounted, in which case we can't take a screenshot, so i think it's not worth it implementing a retry logic. For most cases this should work fine.

src/modules/editor/sagas.ts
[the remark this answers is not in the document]
nicosantangelo: gotcha!

src/modules/editor/sagas.ts
nicosantangelo: Perfect

src/modules/editor/utils.ts
```diff
@@ -106,3 +109,52 @@ export function getKeyboardShortcuts(): KeyboardShortcut[] {
...
+  ctx.drawImage(img, 0, 0, width, height)
+
+  // encode image to data-uri with base64 version of compressed image
+  return canvas.toDataURL()
+}
+
+export function resizeScreenshot(screenshot: string, maxWidth: number, maxHeight: number) {
+  return new Promise<string | null>(resolve => {
+    const img = new Image()
+    img.onload = function resizeImage() {
```
belohlavek: Can't this be anonymous?

cazala:
aoe

PR #154, opened 2019-02-08
fix: reset camera on multi-parcel projects

### LINKED ISSUE #125: Fix reset camera
- Multiparcels should be centered properly
- Beta angle should be reset

belohlavek:
kapture 2019-02-08 at 9 59 25

Closes #125


src/modules/editor/sagas.ts
```diff
@@ -191,9 +191,13 @@ function handleZoomOut() {
   editorWindow.editor.setCameraZoomDelta(5)
 }
 
-function handleResetCamera() {
+function* handleResetCamera() {
+  const project: Project = yield select(getCurrentProject)
+  const x = (project.parcelLayout.rows * 10) / 2
```
cazala: We should use the constant PARCEL_SIZE here

Issue #141, opened 2019-02-07
Submit contest modal


belohlavek:
Should have name, description, email and eth address fields. It should show up after accepting terms or instantly if the terms were accepted previously.

PR #159, opened 2019-02-11
feat: add project to contest

### LINKED ISSUE #100: "Participating in contest" badge
screen shot 2019-02-01 at 10 27 22

### LINKED ISSUE #140: Remove email field from contest term modal
The modal should only serve the purpose of accepting the terms and conditions for the contest

nicosantangelo:
Closes #100 #140 #141


src/components/HomePage/HomePage.types.ts
```diff
@@ -4,8 +4,11 @@ import { NavigateToAction } from 'decentraland-dapps/dist/modules/location/actio
 import { ProjectState } from 'modules/project/reducer'
 import { createProjectFromTemplate, CreateProjectFromTemplateAction } from 'modules/project/actions'
 
-export type Props = {
+export type DefaultProps = {
   projects: ProjectState['data']
+}
+
+export type Props = DefaultProps & {
```
cazala: i think this should be `Partial<DefaultProps>`, since they are not actually required

cazala:
🥇 

Just one thing: if I go to "Update Your Entry" or "Add To Contest" and I change the values of the form, and then I hit cancel, the changes are still saved, as if I clicked on submit.

Nice to have: if the users is signed in, it would be nice to pre-populate the "Eth Address" field with his current account address

src/components/HomePage/HomePage.types.ts
nicosantangelo: [Explained here!](https://github.com/decentraland/builder/pull/160#issuecomment-462733310)

nicosantangelo:
> Just one thing: if I go to "Update Your Entry" or "Add To Contest" and I change the values of the form, and then I hit cancel, the changes are still saved, as if I clicked on submit.

Whops, on it. I'll fix conflicts too

> Nice to have: if the users is signed in, it would be nice to pre-populate the "Eth Address" field with his current account address

I thought of this too, but because we won't add "wallet" support until after the contest is finished I don't think we'll be able to do it

nicosantangelo:
Updated!

PR #160, opened 2019-02-11
refactor: standarize default props

nicosantangelo:


cazala:
I don't want to be a pain in the ass 😅 I know for connected components it's the same, but for non-connected or partially connected components, default props should be used as `Partial<DefaultProps>` when `&`ed with `Props`, otherwise they will be **required** when you try to use that component from jsx.

Also, that way we would be following our standards: https://github.com/decentraland/standards/blob/master/standards/react-redux.md#4-component-types

nicosantangelo:
That's no longer the case! Typescript 3 and up [added support for defaultProps](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#support-for-defaultprops-in-jsx) (yes, a language has lib support 🙃 )

We might want to update the standards tho

cazala:
cool ⭐️

PR #167, opened 2019-02-12
feat: ground picker

### LINKED ISSUE #109: Allow setting ground types
screen shot 2019-02-01 at 18 36 01

belohlavek:
kapture 2019-02-12 at 14 46 24

Note: there are 2 grounds that will be removed from the list, so don't worry about the order.
Note: some grounds appear out of bounds. A fix from Shibu's side is incoming as well.
Note: a new image for empty ground is coming from mat soon.

How it works:

When the save button is pressed two actions are dispatched: 

- edit project
- set ground

The set ground action will check wether you are creating, updating or deleting the ground and provision the scene accordingly.

To make this possible the `provisionScene` action now writes a scene (either creating or updating one by id). This means there is no more need for the `create scene` action. Also entities can have their gizmos disabled by setting the `disableGizmos` flag to true on their definition.

Grounds themselves are picked from the asset list that comes from the asset pack. All assets under the `ground` category will be considered valid grounds and will be handled as such.

Closes #109


belohlavek:
Undo breaks it

belohlavek:
Need to move ground card to be a top level component

belohlavek:
need to revert the removal of create project

cazala:
sabe 🍻

Issue #165, opened 2019-02-12
Textures count doesn't go down when deleting items


cazala:
Textures count doesn't go down when deleting items:

![textures count](https://user-images.githubusercontent.com/2781777/52643332-8d8b7600-2ebb-11e9-9d69-2f85462ef179.gif)

PR #168, opened 2019-02-13
fix: texture count

menduz:
Related: https://github.com/decentraland/client/pull/553

Closes: #165


belohlavek:
Waiting for the client pr to be merged so we can also update the ECS version

PR #201, opened 2019-02-19
fix: restore gizmos after exiting preview mode

### LINKED ISSUE #196: Preview breaks gizmos
- Enter into preview mode

- Exit preview mode

- Gizmos don't work anymore on already placed models (new models added after this come with gizmos)

cazala:
Closes #196

### DIFF
--- a/src/ecsScene/scene.ts
+++ b/src/ecsScene/scene.ts
@@ -50,7 +50,7 @@ function handleExternalAction(message: { type: string; payload: Record<string, a
           entity.remove(Gizmos)
         } else {
           const staticEntities = engine.getComponentGroup(StaticEntity)
-          if (!staticEntities.hasEntity(entity.uuid)) {
+          if (!staticEntities.hasEntity(entity)) {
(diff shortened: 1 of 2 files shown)


belohlavek:
Shouldn't this be #196 ?

src/ecsScene/scene.ts
```diff
@@ -50,7 +50,7 @@ function handleExternalAction(message: { type: string; payload: Record<string, a
           entity.remove(Gizmos)
         } else {
           const staticEntities = engine.getComponentGroup(StaticEntity)
-          if (!staticEntities.hasEntity(entity.uuid)) {
+          if (!staticEntities.hasEntity(entity)) {
```
nicosantangelo: _Question_: Should this be here? I think it's a change on `decentraland-ecs`? but I don't see an update

belohlavek: There is an update but my branch has the latest one, Juanca updated but removed the updated version so we can use mine. I think we should revert this.

PR #209, opened 2019-02-20
fix: prevent reporting each time props are received when more than one metric is exceeded

cazala:
When two or more metrics where exceeded (say, `triangles` and `terxtures`) this component would report a `Metric exceeded` event on every call to `componentWillReceiveProps` (that's a lot) and it was polluting the analytics on segment.

This PR fixes that.

### DIFF
--- a/src/components/EditorPage/Metrics/Metrics.tsx
+++ b/src/components/EditorPage/Metrics/Metrics.tsx
@@ -15,7 +15,7 @@ export default class Metrics extends React.PureComponent<Props, State> {
   }
 
   analytics = getAnalytics()
-  metricExceeded: keyof SceneMetrics | null = null
+  metricsExceeded: (keyof SceneMetrics)[] = []
 
   componentWillMount() {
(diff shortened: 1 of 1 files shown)


src/components/EditorPage/Metrics/Metrics.tsx
```diff
@@ -26,20 +26,17 @@ export default class Metrics extends React.PureComponent<Props, State> {
   }
 
   componentWillReceiveProps(nextProps: Props) {
-    let isMetricExceeded = false
     for (let key in nextProps.metrics) {
       const metric = key as keyof SceneMetrics
       if (nextProps.metrics[metric] > nextProps.limits[metric]) {
-        if (!this.metricExceeded || this.metricExceeded !== metric) {
-          this.metricExceeded = metric
+        if (!this.metricsExceeded.includes(metric)) {
```
nachomazzara: I think that `includes` does not work with an array of objects, only values. Maybe we can use a `find` or `any`

Issue #195, opened 2019-02-19
Custom Layout picker broken with invalid input


belohlavek:
Playing around with big numbers and symbols like `-` breaks the component (rendering gigantic grids sometimes, or freezing the UI otherwise)

PR #211, opened 2019-02-20
feat: add a basic layout picker to the edit modal

### LINKED ISSUE #67: Add option to edit project layout
**Edit**:

We need to investigate the possibility of adding an option to allow the user to edit the scene size. Say you pick a 2x3 scene but decide you want it to be 2x2 later, you need to redo the entire scene.

nicosantangelo:
**OUTDATED GIF, grounds live on the sidebar**

editmodal

- Renames parcelLayout to layout
- Keeps a consistent rows x cols throughout the code

Closes #67 #195


src/modules/storage/migrations.ts
```diff
@@ -0,0 +1,25 @@
+import { utils } from 'decentraland-commons'
+
+export const migrations = {
+  2: (data: any) => {
+    type V1Project = { parcelLayout: { rows: number; cols: number } }
```
belohlavek: Is this necessary at this stage?

src/modules/scene/utils.ts
```diff
@@ -40,3 +40,18 @@ export function snapToGrid(position: Vector3, grid: number = 0.5): Vector3 {
     z: position.z - (position.z % grid)
   }
 }
+
+export function clearGround(groundId: string, entities: Scene['entities']): Scene['entities'] {
```
belohlavek: I would rename this to something like `getEntitiesWithoutGrounds` or something better, clearGround just sounds mutate-y

src/modules/translation/languages/en.json
```diff
@@ -134,13 +135,16 @@
...
     "title": "Custom",
-    "max_area_error": "Your scene can't have more than {area} parcels",
     "up_to": "Up to {max} parcels"
   },
-  "layout_picker": {
-    "rows": "rows",
-    "columns": "columns"
+  "project_layout_picker": {
+    "max_area_error": "Your scene can't have more than {area} parcels",
+    "min_area_error": "Your scene can't it's too small"
```
belohlavek: Should be  "Your scene is too small"

belohlavek:
kapture 2019-02-28 at 0 31 04

Project thumbnail is not updated after resizing. Should we care about this?

belohlavek:
Placed items are temporarily lost after resizing

kapture 2019-02-28 at 0 40 58

belohlavek:
Placing items, resizing and undoing breaks the app. Also for some reason this is not caught by the error boundary 🤔 
`Uncaught TypeError: Cannot read property 'ComponentRemoved' of undefined`
kapture 2019-02-28 at 0 43 47

belohlavek:
Can we avoid showing an error whe the fields are empty? It's good enough if not tho

belohlavek:
kapture 2019-02-28 at 0 58 16

This seems correct, just a heads up (undoing after resize changes ground coverage to the previous size)

belohlavek:
screen shot 2019-02-28 at 00 48 38

Found this while playing around with the modal

belohlavek:
Exiting early after resizing fails to compute grounds

kapture 2019-02-28 at 1 07 56

src/modules/storage/migrations.ts
nicosantangelo: Not really but I didn't want to break any test projects. Removing it is 💯 in my opinion, but you'll need to delete your LocalStorage or the builder will explode

nicosantangelo:
@belohlavek  feedback:
- [ ] Code comments
- [ ] Project thumbnail is not updated after resizing. Should we care about this?
- [ ] Placed items are temporarily lost after resizing
- [ ] Exiting early after resizing fails to compute grounds

--

`onCloseModal`: This is an error on decentraland-dapps, it should be fixed by now, but it's not from this project. Maybe an `npm install` is needed

src/modules/translation/languages/en.json
```diff
@@ -134,13 +135,16 @@
...
     "title": "Custom",
-    "max_area_error": "Your scene can't have more than {area} parcels",
     "up_to": "Up to {max} parcels"
   },
-  "layout_picker": {
-    "rows": "rows",
-    "columns": "columns"
+  "project_layout_picker": {
+    "max_area_error": "Your scene can't have more than {area} parcels",
+    "min_area_error": "Your scene it's too small"
```
belohlavek: is!

cazala: I'd rather get rid of the min error message all together:

min-erroir

it feels better if the modal doesn't resize, and I think is clear enough that you can't submit with an empty (or zero or negative) field

nicosantangelo: oh! ok
I'll do that

Issue #242, opened 2019-02-26
Review mobile page


belohlavek:
- [x] Don't redirect

TBD:
- Should we show a success text after adding your email? Which text?
- Should we try to detect desktop devices (on MobilePage) so we can show a tailored message instead of asking for an email address?

belohlavek:
https://github.com/decentraland/builder/issues/235

PR #245, opened 2019-02-26
feat: don't redirect for mobile pages

nicosantangelo:
Addresses the redirect from #242 
There are still things TBD


src/routing/Routes.tsx
```diff
@@ -29,8 +29,7 @@ export class Routes extends React.Component {
       <>
         <Responsive maxWidth={1024}>
           <Switch>
```
belohlavek: Is the switch needed?

nicosantangelo: Nope! good catch. Updated

PR #264, opened 2019-02-28
chore: added more events to segment

### LINKED ISSUE #249: segment: track OPEN_EDITOR and send along projectId
segment: track OPEN_EDITOR and send along projectId

cazala:
Added more events as requested by @dnul 

Closes #249

### DIFF
--- a/src/modules/analytics/sagas.ts
+++ b/src/modules/analytics/sagas.ts
@@ -74,7 +74,8 @@ function* handleSubmitProject(action: SubmitProjectSuccessAction) {
     projectId: project.id,
     email,
     ethAddress,
-    numEntities: Object.keys(scene.entities).length
+    numEntities: Object.keys(scene.entities).length,
+    layout: project.layout
   })
 }
 
--- a/src/modules/analytics/sagas.ts
+++ b/src/modules/analytics/sagas.ts
@@ -19,47 +19,35 @@ import { getScene } from 'modules/scene/selectors'
 
 export function* segmentSaga() {
   yield takeLatest(OPEN_EDITOR, handleOpenEditor)
-  yield takeLatest(ADD_ITEM, handleAddItem)
-  yield takeLatest(DUPLICATE_ITEM, handleDuplicateItem)
-  yield takeLatest(SET_GROUND, handleSetGround)
+  yield takeLatest(ADD_ITEM, handleNewItem)
+  yield takeLatest(DUPLICATE_ITEM, handleNewItem)
+  yield takeLatest(SET_GROUND, handleNewItem)
   yield takeLatest(DELETE_ITEM, handleRemoveItem)
   yield takeLatest(SUBMIT_PROJECT_SUCCESS, handleSubmitProject)
   yield takeLatest(TOGGLE_SNAP_TO_GRID, handleToggleSnapToGrid)
 }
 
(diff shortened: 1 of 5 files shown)


src/modules/editor/sagas.ts
```diff
@@ -194,7 +194,7 @@ function* handleOpenEditor() {
   yield renderScene()
 
   // Enable snap to grid
-  yield put(toggleSnapToGrid(true))
+  yield handleToggleSnapToGrid(toggleSnapToGrid(true))
```
nicosantangelo: I don't get this change, why are you avoiding the dispatch?

src/modules/analytics/sagas.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  getAnalytics().track('New item', { projectId: project.id })
+}
+
+function* handleSetGround(_: SetGroundAction) {
+  const project: ReturnType<typeof getCurrentProject> = yield select(getCurrentProject)
+  if (!project) return
+
+  getAnalytics().track('New item', { projectId: project.id })
+}
```
nicosantangelo: Not sure about this one, but you could have a `handleNewItem` which handles `ADD_ITEM, DUPLICATE_ITEM and SET_GROUND`

```ts
yield takeLatest(ADD_ITEM, handleNewItem)
yield takeLatest(DUPLICATE_ITEM, handleNewItem)
yield takeLatest(SET_GROUND, handleNewItem)```

File: src/modules/analytics/sagas.ts [hunk 3]
```diff
@@ -0,0 +1,89 @@
...
+    layout: project.layout
+  })
+}
+
+function* handleToggleSnapToGrid(action: ToggleSnapToGridAction) {
+  if (!action.payload.enabled) {
+    const project: ReturnType<typeof getCurrentProject> = yield select(getCurrentProject)
+    if (!project) return
+
+    getAnalytics().track('Enable precision', { projectId: project.id })
```
nicosantangelo: Not sure either but you could add a `track` util function to analytics/utils to avoid having to repeat `getAnalytics().track()`

```ts
export function track(...) {
  getAnalytics().track(...)
}

File: src/modules/analytics/sagas.ts [hunk 4]
```diff
@@ -0,0 +1,89 @@
...
+
+  const scene: ReturnType<typeof getScene> = yield select((state: RootState) => getScene(state, project.sceneId))
+  if (!scene) return
+
+  const { email, ethAddress } = action.payload.contest
+  getAnalytics().track('Submit project', {
+    projectId: project.id,
+    email,
+    ethAddress,
+    numEntities: Object.keys(scene.entities).length,
```
nicosantangelo: I'm sure dnul asked for this name and I'm just being annoying, but maybe `entitiesAmount` ? feel free to disregard

src/modules/editor/sagas.ts
cazala: To avoid calling all those mapStateToProps and mapDispatchToProps :)

src/modules/analytics/sagas.ts
cazala: oh yea, this was due to a refactor 😅 mala mia. I was sending also the asset, and the way to get the asset was different for each action. Then I simplified it and now the 3 actions could be handled by the same handler. I'll fix it.

PR #269, opened 2019-03-01
feat: tutorial modal

### LINKED ISSUE #234: Implement tutorial walkthrough

belohlavek:
kapture 2019-03-01 at 12 43 05

Closes #234


cazala:
My 2 cents: I would go start with drag and drop from the sidebar, then placing items and camera controls. It is the order in which the user will actually do that stuff.

cazala:
ie:

1. Search, drag and drop:

step-1

2. Placing stuff

step-2

3. Arrows/orbits (there's no "gizmo" word on the UI)

step-3

4. Camera

step-4

belohlavek:
I think there's a discussion about including limits as the 4th one and gizmos being mentioned in the first one.

twaldorf:
That order makes more sense for a mini-tutorial, I like mentioning the orbits instead of using the `gizmo` word.

As the interface gets easier to use, we might in the future rely on this more to help people get through whatever funnel we're optimizing, which could mean having the above collapsed into a single interface/workflow slide and an export/deployment slide or something. Just thoughts.

twaldorf:
I'll revise that copy and run it through Eric. I think we should keep the copy up on the top of the slides and not try to overlay it on the gifs.

belohlavek:
Mat's feedback is that we should make only 1 line per slide and keep them overlayed as we have it now

twaldorf:
Ace, him and I are working through it.

twaldorf:
gifs!
tutorial-add
tutorial-camera-optimized
tutorial-movement
tutorial-orbitsarrows

PR #319, opened 2019-03-09
chore: chinese translation

oscnet:


cazala:
Not merging this yet because we still have changes to do to the english version

cazala:
I'm closing this because the translations have been added in this commit https://github.com/decentraland/builder/commit/c2db453d872e2501cf23e6381f2ea29668990117

I used these translations but there were some gaps that I filled with google translate.

@oscnet could you please take a look at [this file](https://github.com/decentraland/builder/blob/master/src/modules/translation/languages/zh.json ) and if you find something wrong send a PR? Thank you ❤️

PR #376, opened 2019-03-20
[DO NOT MERGE] chore: warn users away from stg

### LINKED ISSUE #375: Warn users away from .zone
Users are spending time in the test deployment, making and submitting scenes. To prevent this, we should add a banner, as mentioned by Dani, to the top of the .zone version.

Copy incoming

belohlavek:
Screen Shot 2019-03-20 at 10 31 34

Closes #375


belohlavek:
Do not merge, chinese users are using this env

PR #383, opened 2019-03-21
chore: bump to decentraland-ui@2.0.0

cazala:
This PR bumps the Builder to use `decentraland-ui` v2.0.0

It is currently using a release candidate, we need to release https://github.com/decentraland/ui/pull/32 in order to merge this PR


cazala:
Both `<Row>` and `<Column>` are just layout components that align items to center, and if provided the `center` prop, they also justify to center.

`<Center>` center also aligns and justifies its children to center, but it occupies the whole parent (it uses `position: absolute` and top-left-right-bottom to 0px. It is meant of centering something inside a container. `<Row>` and `<Column>` are just shorthands for `display: flex; position: relative; flex-flow: column|center` which I added because we use those three rules alone a lot just for layouting, and it feels more declarative than having `<div>`s

`<Close>` is an X with a circle around it, it can be used as the `closeModal` prop for modals, and if provided the `small` prop, it can be used to close `<Popup>`s

nicosantangelo:
Awesome!

PR #393, opened 2019-03-22
feat: Optimize color of `Submit to contest` button

### LINKED ISSUE #386: Optimize color of `Submit to contest` button
The colorful `Submit to content` button may be more effective if it matches all the other buttons in the flow. We could change the button to be the gradient red from the tutorial flow. The effect of this would be a reduction in the dropoff from the editor to contest submission, where we lose 80% of users.

abarmat:
Closes #386

### DIFF
--- a/src/components/ContestButton/ContestButton.css
+++ b/src/components/ContestButton/ContestButton.css
@@ -9,8 +9,7 @@
 .ContestButton.ui.button.mini:active,
 .ContestButton.ui.button.mini:focus,
 .ContestButton.ui.button.mini:hover {
-  background-image: url('../../images/contest-button.png');
-  background-size: 100%;
+  background-color: var(--primary);
   transform: translateY(0px);
 }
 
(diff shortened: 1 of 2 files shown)


belohlavek:
@abarmat Make sure to remove the unused image (if truly unused)

src/components/ContestButton/ContestButton.css
```diff
@@ -9,8 +9,7 @@
 .ContestButton.ui.button.mini:active,
 .ContestButton.ui.button.mini:focus,
 .ContestButton.ui.button.mini:hover {
-  background-image: url('../../images/contest-button.png');
-  background-size: 100%;
+  background-color: #ff2d55;
```
belohlavek: This color should be taken from the CSS variable

belohlavek:
Are we naming it Add to Contest again?

src/components/ContestButton/ContestButton.css
abarmat: Will do

PR #413, opened 2019-03-28
fix: vertical snap to bounds

### LINKED ISSUE #412: Vertically out of bounds objets are not clamped

belohlavek:
Closes #412

### DIFF
--- a/src/modules/scene/sagas.ts
+++ b/src/modules/scene/sagas.ts
@@ -15,7 +15,8 @@ import {
   DuplicateItemAction,
   DeleteItemAction,
   SET_GROUND,
-  SetGroundAction
+  SetGroundAction,
+  FIX_CURRENT_SCENE
 } from 'modules/scene/actions'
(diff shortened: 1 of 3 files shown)


src/modules/scene/sagas.ts
```diff
@@ -280,3 +282,24 @@ function* handleSetGround(action: SetGroundAction) {
...
+
+function* handleFixCurrentScene() {
+  const scene: Scene = yield select(getCurrentScene)
+  const bounds: ReturnType<typeof getCurrentBounds> = yield select(getCurrentBounds)
+  const componentIdKeys = Object.keys(scene.components)
+  let components = { ...scene.components }
+
+  for (let key of componentIdKeys) {
+    const component = scene.components[key] as ComponentDefinition<ComponentType.Transform>
+    if (component.type === ComponentType.Transform && bounds) {
```
cazala: instead of adding the `&& bounds` here, we could add an `if (!bounds) return` before the for loop

PR #471, opened 2019-04-24
feat: support for server with multiple asset packs

### LINKED ISSUE #460: Load item drawer info to state using the multi asset-pack definition file

cazala:
Closes #460 

Use the following variable in your `.env`

```
REACT_APP_ASSETS_URL=https://builder-packs-stg.now.sh
```

This PR changes the logic of loading asset packs to do the following:

1. Load the asset pack manifest

2. Get the user selection of asset packs (by default this is all the available asset packs)

3. For every selected asset pack:

  3.1. Check if the asset pack is not in state and it has not already been loaded

  3.2 If not loaded, it fetches it from the server, otherwise uses the one already in state

4. All the non-selected asset packs are added to the state and marked as not loaded (so we can render a UI for the user to select them)

5. All the asset packs that were not in the available asset pack since last time the user visited are marked as new

The selection of asset packs are stored in each project. 

The record of available asset packs and newly added asset packs is stored under the `sidebar` domain. The available asset packs slice is stored in local storage so we can tell which assets have been added since the last visit.

I also refactored the `SELECT_ASSET_PACK` action into  `TOGGLE_ASSET_PACK` which allows us to enabled/disable them individually and in bulk.

### DIFF
--- a/src/modules/assetPack/sagas.ts
+++ b/src/modules/assetPack/sagas.ts
@@ -10,8 +10,8 @@ import { getData as getAssetPacks } from 'modules/assetPack/selectors'
 import { getData as getAssets } from 'modules/asset/selectors'
 import { getSelectedAssetPackIds, getAvailableAssetPackIds } from 'modules/ui/sidebar/selectors'
 import { BaseAssetPack, FullAssetPack } from 'modules/assetPack/types'
+import { setAvailableAssetPacks, setNewAssetPacks } from 'modules/ui/sidebar/actions'
 import { api } from 'lib/api'
-import { setAvailableAssetPacks } from 'modules/ui/sidebar/actions'
 
 export function* assetPackSaga() {
   yield takeLatest(LOAD_ASSET_PACKS_REQUEST, handleLoadAssetPacks)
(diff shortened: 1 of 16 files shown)


src/modules/ui/sidebar/selectors.ts
```diff
@@ -101,6 +95,6 @@ export const getSideBarCategories = createSelector<
       }
     })
 
-    return categoryArray
+    return [...categoryArray]
```
belohlavek: Doesn't the filter already create a new array?

belohlavek: Haha copy pasting :P

src/modules/assetPack/sagas.ts
```diff
@@ -1,39 +1,87 @@
...
   loadAssetPacksFailure,
   LoadAssetPacksRequestAction
 } from 'modules/assetPack/actions'
-import { RemoteAssetPack, FullAssetPack } from 'modules/assetPack/types'
+import { getData as getAssetPacks } from 'modules/assetPack/selectors'
+import { getData as getAssets } from 'modules/asset/selectors'
+import { getSelectedAssetPackIds, getAvailableAssetPackIds } from 'modules/ui/sidebar/selectors'
+import { BaseAssetPack, FullAssetPack } from 'modules/assetPack/types'
 import { api } from 'lib/api'
+import { setAvailableAssetPacks } from 'modules/ui/sidebar/actions'
```
belohlavek: Import order for my OCD's sake

belohlavek:
ui/sidebar domain or assetpack domain?

src/scenarios/helpers/assetPacks.ts
```diff
@@ -24,10 +24,13 @@ export function makeFakeAssetPack(assetCount: number = 1): FullAssetPack {
...
   const categories = ['nature', 'furniture', 'items', 'misc']
 
-  let out: { id: string; version: number; title: string; assets: Asset[] } = {
+  let out: FullAssetPack = {
     id,
-    version: 1,
     title: 'Default Asset Pack',
+    thumbnail: '',
+    url: '',
+    isNew: false,
```
belohlavek: We should have another field in the ui.sidebar domain where new packs are stored (only runtime tho)

cazala:
@belohlavek feedback applied 🤸‍♂️

PR #536, opened 2019-07-12
fix: dapper signature

belohlavek:


cazala:
You can use the helper included with web3: `web3.toHex("hola") // "0x686f6c61"`

belohlavek:
@cazala this won't be a real fix tho, we shouldn't merge it

cazala:
This should work now with the version of the content server in staging.

PR #555, opened 2019-07-22
chore: update banners

abarmat:


src/components/SideBar/AssetPackList/AssetPackList.tsx
```diff
@@ -1,34 +1,26 @@
...
 import SidebarCard from '../SidebarCard'
 import { Props } from './AssetPackList.types'
 import './AssetPackList.css'
 
-const ethereum = (window as any)['ethereum']
-
 export default class AssetPackList extends React.PureComponent<Props> {
   handlePromoClick = () => {
-    this.props.onOpenModal('AdBlockModal', { origin: 'Item catalog Dapper CTA' })
+    window.open('https://avatars.decentraland.org?utm_source=builder&utm_campaign=catalog')
```
nicosantangelo: Hardcoded URL ⚠️

abarmat: Where do you think it might be a good place to set this constant?

src/modules/analytics/campaigns.ts
```diff
@@ -1,3 +1,4 @@
 export const Campaign = {
-  DAPPER: 'dapper'
+  DAPPER: 'dapper',
```
belohlavek: is this still used?

src/components/SideBar/AssetPackList/AssetPackList.tsx
abarmat: Addressed 👍

src/modules/analytics/campaigns.ts
abarmat: Will remove that

abarmat: Done

PR #586, opened 2019-08-09
feat: remote thumbnails

### LINKED ISSUE #543: Cloud save thumbnails
- On save send project without thumbnail field
- On save send thumbnail to /media endpoint on a separate request
- On get project, the thumbnail is included pointing to s3 bucket
- On change, the local thumbnail field is updated with base64 image

belohlavek:
Closes #543


src/modules/media/sagas.ts
```diff
@@ -67,7 +67,7 @@ export function* handleTakePictures() {
   canvas.classList.remove('recording')
   yield call(() => editorWindow.editor.resize())
 
-  yield put(recordMediaSuccess({ ...shots, thumbnail }))
+  yield put(recordMediaSuccess({ ...shots, preview: thumbnail }))
```
belohlavek: check this

src/lib/api/builder.ts
```diff
@@ -144,16 +148,27 @@ export class BuilderAPI extends BaseAPI {
...
     await this.request('put', `/projects/${project.id}/manifest`, { manifest }, authorize())
   }
 
+  async saveThumbnail(project: Project) {
+    const blob = dataURLToBlob(project.thumbnail)
+    const formData = new FormData()
+    if (blob) {
+      formData.append('thumbnail', blob)
+      await this.request('post', `/projects/${project.id}/media`, formData, {
+        ...authorize()
```
cazala: this could just be `authorize()` without the `{... }`. as in `saveProject()`

src/modules/common/types.ts
```diff
@@ -20,6 +20,7 @@ import { DeploymentState } from 'modules/deployment/reducer'
 import { MediaState } from 'modules/media/reducer'
 import { AuthState } from 'modules/auth/types'
 import { SyncState } from 'modules/sync/types'
+import { Omit } from 'react-redux'
```
cazala: import order kkkkkkkkkkkkkkkkkkKKKKKAAAAAAHHHHHHH 

❤️

belohlavek: JAJAJAJAJ

src/modules/sync/sagas.ts
```diff
@@ -71,6 +73,7 @@ export function* syncSaga() {
   yield takeLatest(DEPLOY_TO_LAND_SUCCESS, handleDeployToLandSuccess)
   yield takeLatest(CLEAR_DEPLOYMENT_SUCCESS, handleClearDeploymentSuccess)
   yield takeLatest(MARK_DIRTY, handleMarkDirty)
+  yield takeLatest(EDIT_PROJECT_THUMBNAIL, handleSaveThumbnail)
```
cazala: to keep naming consisten this handler should be `handleEditProjectThumbnail`

cazala: and also, when I create a new project, the `POST .../media` goes out before the first `PUT .../manifest` returning a 404 because the project doesn't exist yet.

Should we instead hook this handler to the `SAVE_PROJECT_SUCCESS` rather than the `EDIT_PROJECT_THUMBNAIL` to make sure this doesn't happen?

src/modules/media/reducer.ts
```diff
@@ -23,7 +23,7 @@ export type MediaReducerAction = RecordMediaRequestAction | RecordMediaSuccessAc
 export const mediaReducer = (state = INITIAL_STATE, action: MediaReducerAction): MediaState => {
   switch (action.type) {
     case RECORD_MEDIA_SUCCESS: {
-      const { north, east, south, west, thumbnail } = action.payload.media
+      const { north, east, south, west, preview: thumbnail } = action.payload.media
```
cazala: shouldn't this be just `prevew` and also on line 39 below?

src/modules/sync/sagas.ts
belohlavek: SAVE_PROJECT_SUCCESS won't cotain the latest thumbnail, as a matter of fact it would contain the remote preview.png url for the first save

I'll keep thinking about how to solve both cases

belohlavek: PROVISION_SCENE triggers handleSceneChange
handleSceneChange triggers SAVE_SCREENSHOT
SAVE_SCREENSHOT triggers EDIT_PROJECT_THUMBNAIL
(thumbnail updated at this point)

PROVISION_SCENE triggers SAVE_PROJECT_REQUEST
(thumbnail not updated at this point)

PR #615, opened 2019-08-14
fix: publish to LAND while logged out

### LINKED ISSUE #614: Publish to LAND doesn't work when logged out
Publish to LAND doesn't work when logged out because it's hitting some endpoints (`GET /deployments` and `POST /media`) that shouldn't hit while logged out.

cazala:
Fixes #614

### DIFF
--- a/src/components/Modals/DeployModal/DeployToLand/LandAtlas/LandAtlas.tsx
+++ b/src/components/Modals/DeployModal/DeployToLand/LandAtlas/LandAtlas.tsx
@@ -25,8 +25,10 @@ export default class LandAtlas extends React.PureComponent<Props, State> {
 
   componentDidMount() {
     this.mounted = true
-    const { onFetchDeployments } = this.props
-    onFetchDeployments()
+    const { isLoggedIn, onFetchDeployments } = this.props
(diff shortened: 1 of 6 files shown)


src/components/Modals/DeployModal/DeployToLand/LandAtlas/LandAtlas.tsx
```diff
@@ -25,8 +25,10 @@ export default class LandAtlas extends React.PureComponent<Props, State> {
 
   componentDidMount() {
     this.mounted = true
-    const { onFetchDeployments } = this.props
-    onFetchDeployments()
+    const { isLoggedIn, onFetchDeployments } = this.props
+    if (isLoggedIn) {
+      onFetchDeployments()
```
belohlavek: this now only happens when logged in?

cazala: yes, if you are not logged in, you aready have your deployments in memory

cazala: they come as `occupiedParcels`

PR #616, opened 2019-08-15
fix: thumbnails for duplicated scenes

### LINKED ISSUE #613: Duplicated scene has no thumbnail
Repro:

While logged in and in dashboard:
- Duplicate a scene (it will have a thumbnail)
- Refresh the page (thumbnail is gone)

belohlavek:
Add thumbnail to the duplicated project

If the image is referenced remotely, fetch it and convert it to base64

On project save success the new media should be saved separately

Closes #613

### DIFF
--- a/src/modules/project/sagas.ts
+++ b/src/modules/project/sagas.ts
@@ -118,7 +118,7 @@ function* handleDuplicateProject(action: DuplicateProjectAction) {
 
   let thumbnail = project.thumbnail
 
-  if (thumbnail && !thumbnail.startsWith('data:image/png;')) {
+  if (thumbnail && thumbnail.startsWith('http')) {
     thumbnail = yield call(() => getImageAsDataUrl(project.thumbnail))
   }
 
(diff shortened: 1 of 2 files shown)


src/modules/project/sagas.ts
```diff
@@ -116,8 +116,14 @@ function* handleDuplicateProject(action: DuplicateProjectAction) {
 
   const scene = yield getSceneByProjectId(project.id)
 
+  let thumbnail = project.thumbnail
+
+  if (thumbnail && !thumbnail.startsWith('data:image/png;')) {
```
cazala: should we use `startsWith('http')` to keep it consitent? (https://github.com/decentraland/builder/blob/master/src/components/ProjectCard/ProjectCard.tsx#L72) 

Also if we are going to `fetch` it should be an http/s endpoint

belohlavek: doh, yep

PR #623, opened 2019-08-15
feat: add localstorage toast

cazala:
This PR adds back the localstorag toast, makes it appear after the user has 5 items in the scene, and changes the copy to prompt the user to sign in.

Screen Shot 2019-08-15 at 6 59 48 PM

### DIFF
--- a/src/components/EditorPage/EditorPage.tsx
+++ b/src/components/EditorPage/EditorPage.tsx
@@ -7,6 +7,7 @@ import LoadingPage from 'components/LoadingPage'
 import TopBar from 'components/TopBar'
 import ViewPort from 'components/ViewPort'
 import SideBar from 'components/SideBar'
+import LocalStorageToast from 'components/LocalStorageToast'
 import Tools from './Tools'
 import Metrics from './Metrics'
 import ItemDragLayer from './ItemDragLayer'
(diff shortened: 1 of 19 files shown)


src/components/EditorPage/EditorPage.tsx
```diff
@@ -83,6 +86,14 @@ export default class EditorPage extends React.PureComponent<Props, State> {
       return <NotFoundPage />
     }
 
+    const showLocalStorageToast =
+      !isLoggedIn && // user is not logged in
+      currentScene !== null && // the scene is loaded and there are at more transfrom components than the threshold
+      Object.values(currentScene.components).reduce<number>(
+        (total, component) => (component.type === ComponentType.Transform ? total + 1 : total),
+        0
+      ) > TOAST_ITEMS_THRESHOLD
```
belohlavek: Do we want to interate over all the components on every `render()`?

As an alternative, instead of using the amount of transform components as  the THRESHOLD, we could use the metric's bodies. If we chose this option we need to take into account the following:

- Some items do not have collision meshes and represent a single body ⚠️
- Most items have 2 bodies
- Each ground "tile" represents 1 body (3x3 scene = 9 bodies)

belohlavek: If we use bodies and we keep `5` as the TOAST_ITEMS_THRESHOLD, then 5 non-collidable plants would trigger it, or 6 collidable items

`bodies - (rows * cols) > TOAST_ITEMS_THRESHOLD `

cazala: what about creating a selector shoudShowToast, and we can get rid of the currentScene and isLoggedIn props

PR #625, opened 2019-08-16
feat: empty projects ui

belohlavek:


src/components/HomePage/HomePage.css
```diff
@@ -39,6 +39,20 @@
   position: relative;
 }
 
+.HomePage .empty-projects {
+  display: flex;
+  width: 100%;
+  height: 200px;
+  border-radius: 12px;
+  border: 1px solid var(--bluish-steel);
```
cazala:
```suggestion
  border: 1px solid var(--divider);
```

src/components/HomePage/HomePage.css
```diff
@@ -39,6 +39,20 @@
...
+  align-items: center;
+  justify-content: center;
+  color: var(--bluish-steel);
```
cazala:
```suggestion
  color: var(--secondary-text);
```

src/components/HomePage/HomePage.css
cazala: I would use the `divider` variable, that is meant for those kind of lines, it's a little darker:

Screen Shot 2019-08-16 at 3 23 59 PM

src/components/HomePage/HomePage.css
cazala: let's use `secondary-text`, which is the same color (for this theme) but is more semantically correct

PR #648, opened 2019-08-28
feat: AssetImporter component

### LINKED ISSUE #636: AssetImport Component
Screen Shot 2019-08-22 at 15 36 31

Screen Shot 2019-08-22 at 15 37 11

You will need to abstract a FileInput component to share the drop-zone logic

AssetImport Props:
- onAssets((assets: Asset[], errors: Record<string, string>))

belohlavek:
Kapture 2019-09-09 at 9 07 07

Closes #636


belohlavek:
Missing extension whitelist
Missing translations
Use PureComponent

src/components/AssetImporter/AssetImporter.tsx
```diff
@@ -0,0 +1,257 @@
...
+      thumbnail: '',
+      url: '',
+      isLoaded: false,
+      assets
+    })
+  }
+
+  render() {
+    const { files } = this.state
+    const fileKeys = Object.keys(files)
```
cazala: Here we are doing an `Object.keys` and then an `Object.values` below (for the `items` prop). What about just using the `items` and drop the `Object.keys`? Like:

```tsx
    const items = Object.values(files)
    const buttonText = items.length > 1 ? t('asset_pack.import.action_many', { count: items.length }) : t('asset_pack.import.action')
    const hasCorrupted = items.some(item => !!item.isCorrupted)
    const canImport = items.length > 0 && !hasCorrupted
```

And then just pass `items` to `<FileImport>`

belohlavek: 👌

src/components/AssetImporter/AssetImporter.tsx
```diff
@@ -0,0 +1,257 @@
...
+    const manifestRaw = zip.file(ASSET_MANIFEST)
+    let manifestParsed: Asset | null = null
+
+    if (manifestRaw) {
+      const content = await manifestRaw.async('text')
+      manifestParsed = JSON.parse(content)
+    }
+
+    if (manifestParsed && !manifestParsed.name) {
+      throw new Error('Invalid project')
```
cazala: maybe this should be `throw new Error(t('asset_pack.import.invalid'))` ?

src/components/AssetImporter/AssetImporter.tsx
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    const fileNames: string[] = []
+
+    zip.forEach(fileName => {
+      if (fileName === EXPORT_PATH.MANIFEST_FILE) {
+        throw new Error(t('asset_pack.import.invalid'))
+      }
+
+      if (fileName !== ASSET_MANIFEST || !basename(fileName).startsWith('.')) {
```
cazala: `&&`

PR #665, opened 2019-09-09
feat: asset pack editor

### LINKED ISSUE #639: AssetPackViewer component
Screen Shot 2019-08-22 at 16 08 25

belohlavek:
Kapture 2019-09-09 at 9 20 59

Depends on #664 

Closes #639


src/components/AssetPackEditor/AssetPackEditor.tsx
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+
+  handleThumbnail = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const { assetPack, onChange } = this.props
+    const { files } = e.target
+
+    if (files) {
+      const file = files[0]
+      if (file.size > MAX_THUMBNAIL_SIZE) {
```
cazala: II got an exeption on this line because `file` was `undefined`. I don't know how to reproduce, but that `if (files) { ` above is not stopping an empty array from going through:

```
const files = []
if (files) {
 console.log('boom') // 💥
}
```

We should make it `if (files.length > 0) {`

src/modules/translation/languages/es.json
```diff
@@ -373,9 +373,23 @@
       "action": "Continuar",
       "action_skip": "Saltar paso",
       "errors": {
-        "name_length": "El nombre del Asset solo puede contener un máximo de {count} caracteres",
+        "name_length": "El nombre del Asset sólo puede contener un máximo de {count} caracteres",
```
cazala: 🧐

belohlavek: aca hay que cambiar algo? :P

cazala: jaja no

PR #679, opened 2019-09-18
fix: progress bar and requests

belohlavek:
- Progress bar fix
- Copy fix
- Styles fix (error and overflow container on AssetPackEditor)
- Parallel requests fix
- Track new asset packs on state when created

### DIFF
--- a/src/modules/asset/reducer.ts
+++ b/src/modules/asset/reducer.ts
@@ -64,16 +64,18 @@ export const assetReducer = (state = INITIAL_STATE, action: AssetReducerAction):
     }
     case SAVE_ASSET_PACK_SUCCESS: {
       const { assetPack } = action.payload
+      const assets: AssetState['data'] = {}
+
+      for (const asset of assetPack.assets) {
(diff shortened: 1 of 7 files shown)


src/modules/asset/reducer.ts
```diff
@@ -52,6 +62,18 @@ export const assetReducer = (state = INITIAL_STATE, action: AssetReducerAction):
...
     }
+    case SAVE_ASSET_PACK_SUCCESS: {
+      const { assetPack } = action.payload
+
+      return {
+        loading: loadingReducer(state.loading, action),
+        error: null,
+        data: {
+          ...state.data,
+          ...assetPack.assets.reduce((accum, asset) => ({ ...accum, [asset.id]: asset }), {})
```
cazala: lets change this into:

```
...assetPack.assets.reduce((acc, asset) => { 
  acc[asset.id] = asset
  return acc
}, {})
```

it's `O(n)` vs `O(n!)`

PR #682, opened 2019-09-18
docs: add banner for hackathon

2fd:


src/components/HomePage/TopBanner/TopBanner.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import "./TopBanner.css"
+
+export default class TopBanner extends React.PureComponent {
+
+  render() {
+    return <Banner className="TopBanner" isClosable={true} name="dcl-top-banner">
+      <p>
+      <span>{t('banners.top_banner_left')}</span>
+      <a href="https://hack.decentraland.org/">{' '}{t('banners.top_banner_action')}{' '}</a>
```
cazala: is this url correct? shouldn't it be `https://gamejam.decentraland.org` ?

src/modules/translation/languages/en.json
```diff
@@ -72,7 +72,10 @@
     "promo_assetpack_banner": "The Fantasy Asset Pack allows you to conjure all manner of dazzling creations!",
     "promo_body": "Get the Fantasy Asset Pack and conjure all manner of dazzling creations!",
     "promo_subtitle": "Where the magic happens!",
-    "promo_title": "Decentraland"
+    "promo_title": "Decentraland",
+    "top_banner_left":"Sign up for our upcoming",
+    "top_banner_action":"Hackathon!",
```
cazala: just double checking, I thought we dropped the "hackathon" in favor of "game jam", is this copy ok?

src/components/HomePage/TopBanner/TopBanner.tsx
2fd: Yes, it is correct. Must be the same than docs.decentraland.org
Screen Shot 2019-09-18 at 23 11 43

src/modules/translation/languages/en.json
2fd: yes, it is ok

src/components/HomePage/TopBanner/TopBanner.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import Banner from 'components/Banner'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+
+import "./TopBanner.css"
+
+export default class TopBanner extends React.PureComponent {
+
+  render() {
+    return <Banner className="TopBanner" isClosable={true} name="dcl-top-banner">
```
belohlavek: I'd change `dcl-top-banner` for something like `dcl-hackathon-banner`
Once a banner is closed the state is saved in localstorage and we won't be able to use that name again, let's keep them specific to the event just to avoid overlapping

belohlavek:
Looking good ;)

PR #713, opened 2019-09-24
feat: Profile API migration

nachomazzara:
- Usage of a custom API with _`fetch`_ instead of _`axios`_ because of how plattform created its API. Maybe we should consider adding a _`BaseFetchAPI`_ along with the one we have at `decentraland-dapps`.


nicosantangelo:
The problem it's not axios is it? is that decentraland-dapps uses a convention the  platform team didn't decide to use?

belohlavek:
@cazala when should this be merged?

cazala:
we need to coordinate with @nachomazzara but i think we should merge and release today

PR #726, opened 2019-09-25
feat: added default asset pack name

### LINKED ISSUE #688: Default asset pack name

cazala:
Closes #688

### DIFF
--- a/src/components/Modals/CreateAssetPackModal/CreateAssetPackModal.tsx
+++ b/src/components/Modals/CreateAssetPackModal/CreateAssetPackModal.tsx
@@ -29,7 +29,7 @@ export default class CreateAssetPackModal extends React.PureComponent<Props, Sta
     let name = defaultName
     let suffix = 2
     while (assetPacks.some(assetPack => assetPack.title === name)) {
-      name = `${defaultName} (${suffix})`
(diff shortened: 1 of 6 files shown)


src/components/Modals/CreateAssetPackModal/CreateAssetPackModal.tsx
```diff
@@ -23,12 +23,24 @@ export default class CreateAssetPackModal extends React.PureComponent<Props, Sta
...
+  getDefaultAssetPackName() {
+    const { assetPacks } = this.props
+    const defaultName = t('asset_pack.default_name')
+    let name = defaultName
+    let suffix = 2
+    while (assetPacks.some(assetPack => assetPack.title === name)) {
+      name = `${defaultName} (${suffix})`
+      suffix++
+    }
+    return name
```
nicosantangelo: Suggestion: you could add the suffix as a variable in the translation, to avoid problems if we add languages that read from right to left. Like "default_name": "My Asset Pack ({suffix})"
and then
`return t('asset_pack.default_name', { suffix })`

Not really necessary now tho

cazala:
@nicosantangelo done ✅

PR #728, opened 2019-09-25
feat: ask for confirmation before exit only if dirty

### LINKED ISSUE #727: Ask for confirmation before closing only if asset pack is dirty

cazala:
Closes #727


nicosantangelo:
Because of how's implemented if I delete a character and then press `cmd+z (undo)` the "are you sure?" modal will popup anyways. Is this something we want to avoid?

cazala:
I think the most important thing is avoiding getting that confirmation if you just opened the modal. Knowing if the asset pack changed means diffing and I don't think that complexity is worth it.

PR #736, opened 2019-09-27
fix: snap scale

### LINKED ISSUE #673: Limit minimum scale
Should be higher than 0 to prevent the asset from flipping or disappearing

### LINKED ISSUE #725: Scaling an item to zero makes it unselectable

belohlavek:
Closes #725 
Closes #673

### DIFF
--- a/src/modules/editor/utils.ts
+++ b/src/modules/editor/utils.ts
@@ -8,6 +8,10 @@ const script = require('raw-loader!../../ecsScene/scene.js')
 
 export const THUMBNAIL_WIDTH = 984
 export const THUMBNAIL_HEIGHT = 728
+export const POSITION_GRID_RESOLUTION = 0.5
+export const ROTATION_GRID_RESOLUTION = Math.PI / 16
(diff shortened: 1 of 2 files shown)


src/modules/editor/utils.ts
```diff
@@ -81,3 +82,11 @@ export function resizeScreenshot(screenshot: string, maxWidth: number, maxHeight
     img.src = screenshot
   })
 }
+
+export function snapScale(scale: Vector3): Vector3 {
+  return {
+    x: scale.x === 0 ? 0.001 : scale.x,
+    y: scale.y === 0 ? 0.001 : scale.y,
+    z: scale.z === 0 ? 0.001 : scale.z
```
nicosantangelo: Question, maybe it'd be safer to do this?

```ts
  const threshold = 0.001
  return {
    x: scale.x < threshold ? threshold : scale.x,
    y: scale.y < threshold ? threshold : scale.y,
    z: scale.z < threshold ? threshold : scale.z
  }
```

PR #761, opened 2019-10-02
feat: sync sign in

belohlavek:
Adds sign in link to sync toast
Missing copy changes because link first looks ugly @abarmat


cazala:
suggestion: why not using the same copy as in the other sign in toast?

abarmat:
@cazala you mean like this one? `"subtitle": "You need to sign in before you publish to the Scene pool"`

cazala:
There’s one that says something like: “Scenes are saved in your browser’s storage. Sign in to save them in the Cloud”, it’s the toast inside the editor.

cazala:
https://github.com/decentraland/builder/blob/master/src/modules/translation/languages/en.json#L409

PR #772, opened 2019-10-16
feat: use new lightweight scene-writer

eordano:


cazala:
This would also use the `LightweightWriter` when exporting the scene (SDK export), we need to support both exports. Also the current `dcl-scene-writer` doesn't work with the latest ECS (i'm working on a fix for that on the `dcl-scene-writer` repo) until then, we can't use it here.

PR #846, opened 2019-11-19
fix: renameEntity and removeEntityReferences

### LINKED ISSUE #839: Cloudsaves indicator shouldn't have a cursor pointer since it's not clickable
Cloudsaves indicator shouldn't have a cursor pointer since it's not clickable

belohlavek:
Closes #839 

And makes it so that entity fields are also affected by renameEntity/removeEntityReferences

### DIFF
--- a/src/modules/scene/utils.ts
+++ b/src/modules/scene/utils.ts
@@ -218,7 +218,7 @@ export function getDefaultValues(entityName: string, parameters: AssetParameter[
  *
  * Mutates the object.
  *
- * @param paraneters The Asset parameters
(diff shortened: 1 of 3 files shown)


src/modules/scene/utils.ts
```diff
@@ -217,20 +217,26 @@ export function getDefaultValues(entityName: string, parameters: AssetParameter[
  * Swaps all references to `oldName` for `newName` inside a script's actions.
  *
  * Mutates the object.
+ *
+ * @param paraneters The Asset parameters
```
cazala: paraNeters

PR #887, opened 2019-12-03
fix: ground entity names

### LINKED ISSUE #882: Ground deployment failure on multipleparcel scenes

belohlavek:
Properly generate ground entity names (entity{N})
Get unique names from assets rather than GLTFShapes (fixes case where there's an scene.gltf model that conflicts with the scene entity wrapper on export)
Keep old naming system for legacy purposes/migration

Closes #882

### DIFF
--- a/src/modules/scene/utils.ts
+++ b/src/modules/scene/utils.ts
@@ -139,7 +139,7 @@ export function getUniqueName(components: AnyComponent[], takenNames: Readonly<S
       if (component.type === ComponentType.GLTFShape) {
         const asset = assets[(component as ComponentDefinition<ComponentType.GLTFShape>).data.assetId]
         if (asset) {
-          rawName = asset.name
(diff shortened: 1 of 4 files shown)


src/modules/scene/utils.ts
```diff
@@ -127,17 +127,20 @@ export function getEntityName(scene: Scene, entityComponents: EntityDefinition['
...
   let attempts = 1
   let rawName = 'entity'
 
   for (let component of components) {
     try {
       if (component.type === ComponentType.GLTFShape) {
-        rawName = getGLTFShapeName(component as ComponentDefinition<ComponentType.GLTFShape>)
+        const asset = assets[(component as ComponentDefinition<ComponentType.GLTFShape>).data.assetId]
+        if (asset) {
+          rawName = asset.name
```
cazala: this should be sanitized, and preferably camelcased too

Issue #986, opened 2020-05-12
alert error on builder when using assets search box: Cannot read property 'toLowerCase' of null


pabloes:
Browser: Chrome Version 81.0.4044.138 (Official Build) (64-bit)
SO: macOS Catalina 10.15.4 (19E287) 

Way to reproduce:
I open an scene 4x4 that only has floor
Browser URL is https://builder.decentraland.org/editor/4af3cf08-a9b0-42c8-b81c-7e086f699726?realm=localhost-stub
I search in assets "NFT" (it happens with any search)

**ERROR IN ALERT SHOWN:**
`Invoking error handler due to
Uncaught TypeError: Cannot read property 'toLowerCase' of null

**CONSOLE ERROR:**
```
react-dom.production.min.js:2835 Uncaught TypeError: Cannot read property 'toLowerCase' of null
    at selectors.ts:30
    at selectors.ts:80
    at Array.filter (<anonymous>)
    at selectors.ts:80
    at index.js:69
    at index.js:30
    at index.js:82
    at index.js:30
    at Function.mapToProps (SideBar.container.tsx:22)
    at r (wrapMapToProps.js:41)
    at Function.r.mapToProps (wrapMapToProps.js:50)
    at r (wrapMapToProps.js:41)
    at selectorFactory.js:22
    at connectAdvanced.js:208
    at Object.useMemo (react-dom.production.min.js:3910)
    at useMemo (react.production.min.js:341)
    at f (connectAdvanced.js:193)
    at pi (react-dom.production.min.js:3610)
    at Xi (react-dom.production.min.js:4201)
    at Yi (react-dom.production.min.js:4189)
    at Wi (react-dom.production.min.js:4172)
    at Ss (react-dom.production.min.js:6785)
    at Sc (react-dom.production.min.js:5648)
    at wc (react-dom.production.min.js:5637)
    at yc (react-dom.production.min.js:5393)
    at react-dom.production.min.js:2829
    at t.unstable_runWithPriority (scheduler.production.min.js:270)
    at Wa (react-dom.production.min.js:2794)
    at Qa (react-dom.production.min.js:2824)
    at Xa (react-dom.production.min.js:2814)
    at vc (react-dom.production.min.js:5434)
    at Object.notify (Subscription.js:23)
    at e.t.notifyNestedSubs (Subscription.js:65)
    at e.t.handleChangeWrapper (Subscription.js:70)
    at Object.x [as dispatch] (redux.js:220)
[...]

PR #989, opened 2020-05-17
ISSUE-986 assets with name null performs errors on search

pabloes:
When you have a NFT kittie asset without name, its value is null.

I just changed code assuming that name can be null in assets, but I don't have time right now to look deeper if there is necessity to create a different type and its use, the fact is that in this commit, I'm checking asset.name in places where maybe is supposed it cannot be null.

If you need any change from side just tell me.
https://github.com/decentraland/builder/issues/986#issue-616423703


cazala:
Hey @pabloes thanks for sending this PR and filling the issue! There's an easier way to handle this by updating the `DARAsset` type to have `name` as `string | null` instead of changing the `Asset` type. This way we only need to handle the `null` name when we convert the `DARAsset` into an `Asset`. I sent PR #990 in order to do that.

Thanks for contributing!

PR #1046, opened 2020-08-28
feat: create item

cazala:


nicosantangelo:
Oh! one more thing. We migth want to wire the two buttons we have on the empty page, and fix a tiny bug in the process.

If you go to AvatarPage.tsx, on line 61 you see:
```
<Card.Group>
          {count >= 0 ? (
```

that should be

```
<Card.Group>
          {count > 0 ? (
```

Fixing that, and removing the mocked items and collection you can see the empty page:

image

cazala:
> One question, I was able to create a new item with an extremely long name. Is that desirable?

We do trim the name, but the max length is set to 48 chars. It used to be 20 but we have some items that have longer names than that, like `"Horizontal fantasy platform"`

PR #1055, opened 2020-09-03
fix: like modal

nicosantangelo:


src/components/Modals/LoginModal/LoginModal.tsx
```diff
@@ -1,38 +1,40 @@
 import * as React from 'react'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
 import { Button, ModalNavigation } from 'decentraland-ui'
 import Modal from 'decentraland-dapps/dist/containers/Modal'
 
 import { Props } from './LoginModal.types'
 
 import './LoginModal.css'
```
cazala, line 8: It looks like this file `src/components/Modals/LoginModal/LoginModal.css` has been deleted. It's causing Vercel to fail to build.

cazala (changes requested):

src/components/Modals/LoginModal/LoginModal.tsx
nicosantangelo: My bad! fixed

cazala (approved):

PR #1059, opened 2020-09-04
feat: edit item

nicosantangelo:
**Continues** #1052


src/modules/translation/languages/en.json
```diff
@@ -761,5 +761,10 @@
     "rarity": "Rarity",
     "price": "Price",
     "beneficiary": "Beneficiary"
+  },
+  "edit_price_and_beneficiary_modal": {
+    "title": "Set price and address for {name}",
+    "price_label": "Price",
+    "beneficiary_label": "Beneficiary"
```
cazala, line 768: These seem to be only in the english translations

cazala (changes requested):

src/modules/translation/languages/en.json
nicosantangelo: great catch!

cazala (approved):

PR #1066, opened 2020-09-08
feat: add representation to item

### LINKED ISSUE #1021: Add representation to item
When a user creates an item and flags it as Shape A or Shape B, we should ask if it's part of an existing item. If so, instead of creating a new item, a representation will be generated and added to an existing item.

cazala:
Closes #1021

### DIFF
--- a/src/components/ItemDetailPage/ItemDetailPage.tsx
+++ b/src/components/ItemDetailPage/ItemDetailPage.tsx
@@ -12,6 +12,7 @@ import LoggedInDetailPage from 'components/LoggedInDetailPage'
 import NotFound from 'components/NotFound'
 import { Props } from './ItemDetailPage.types'
 import './ItemDetailPage.css'
+import { getMissingBodyShapeType } from 'modules/item/utils'
 
 const STORAGE_KEY = 'dcl-item-notice'
 
@@ -21,11 +22,18 @@ export default class ItemDetailPage extends React.PureComponent<Props> {
     onDelete(item!)
   }
 
+  handleAddRepresentationToItem = () => {
+    const { item, onOpenModal } = this.props
+    onOpenModal('CreateItemModal', { addRepresentationTo: item })
+  }
+
   renderPage() {
     const { onNavigate } = this.props
 
     const item = this.props.item!
     const data = item.data as WearableData
+
+    const missingBodyShape = getMissingBodyShapeType(item)
 
     return (
       <>
@@ -51,6 +59,12 @@ export default class ItemDetailPage extends React.PureComponent<Props> {
                       direction="left"
                     >
                       <Dropdown.Menu>
+                        {missingBodyShape !== null ? (
+                          <Dropdown.Item
+                            text={t('item_detail_page.add_representation', { bodyShape: t(`global.representation.${missingBodyShape}`) })}
+                            onClick={this.handleAddRepresentationToItem}
+                          />
+                        ) : null}
                         <ConfirmDelete
                           name={item.name}
                           onDelete={this.handleDeleteItem}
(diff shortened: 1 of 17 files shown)


src/components/Modals/CreateItemModal/CreateItemModal.tsx
```diff
@@ -2,70 +2,125 @@ import * as React from 'react'
...
+        // add female representation
+        if (bodyShape === BodyShapeType.FEMALE || bodyShape === BodyShapeType.UNISEX) {
+          item.data.representations.push({
+            bodyShape: [WearableBodyShape.FEMALE],
+            mainFile: model,
+            contents: Object.keys(contents),
+            overrideHides: [],
+            overrideReplaces: []
+          })
+        }
```
nicosantangelo, line 123: Question:
- Is it better to maybe have this on a separate function? maybe on an utils file? I'm not certain to be honest because this Component represents the creation of a new `item`, but it seems like a bit too much logic for a component? not sure

src/modules/item/utils.ts
```diff
@@ -8,3 +8,29 @@ export function isEditable(item: Item) {
   const data = item.data as WearableData
   return !item.rarity || !data.category
 }
+
+export function getBodyShapeType(item: Item) {
```
nicosantangelo, line 12: Maybe this should be used to refactor ItemBadge? it's using a similar but slightly different logic to accomplish the same thing.

src/components/ItemDetailPage/ItemDetailPage.tsx
```diff
@@ -51,6 +59,12 @@ export default class ItemDetailPage extends React.PureComponent<Props> {
                       direction="left"
                     >
                       <Dropdown.Menu>
+                        {missingBodyShape !== null ? (
+                          <Dropdown.Item
+                            text={t('item_detail_page.add_representation', { bodyShape: t(`global.representation.${missingBodyShape}`) })}
```
nicosantangelo, line 64: I know it wasn't introduced in this PR but maybe this should be called `
```
t(`bodyShapes.${missingBodyShape}`)
```
on the translations file?

src/components/Modals/CreateItemModal/CreateItemModal.tsx
```diff
@@ -255,21 +331,61 @@ export default class CreateItemModal extends React.PureComponent<Props, State> {
...
+                      <Field className="name" label={t('create_item_modal.name_label')} value={name} onChange={this.handleNameChange} />
+                    ) : (
+                      <>
+                        {isAddingRepresentation ? null : (
+                          <Section>
+                            <Header sub>{t('create_item_modal.existing_item')}</Header>
+                            <Row>
+                              <div
+                                className={`option ${isRepresentation === true ? 'active' : ''}`}
+                                onClick={() => this.setState({ isRepresentation: true })}
```
nicosantangelo, line 356: I asked this somewhere else but why's this not a bound instance method like `this.handleToggleIsRepresentation`?

nicosantangelo (commented):
A few comments. Non blocking

src/components/ItemDetailPage/ItemDetailPage.tsx
cazala: done 👍

src/modules/item/utils.ts
cazala: done 👍

src/components/Modals/CreateItemModal/CreateItemModal.tsx
cazala: done 👍

src/components/Modals/CreateItemModal/CreateItemModal.tsx
cazala: mm i didn't add it as a utils because i don't think we would reuse it anywhere for now, also it uses stuff from the component's state, and class methods, so if i wanted to abstract it out, i would still need to pass all that stuff as arguments, which doesn't seem ideal

PR #1079, opened 2020-09-23
feat: manage collaborators

nicosantangelo:


src/components/Profile/Profile.tsx
```diff
@@ -14,7 +14,7 @@ export default class Profile extends React.PureComponent<Props> {
   }
 
   render() {
-    const { address, avatar, textOnly, imageOnly, size } = this.props
+    const { address, avatar, textOnly, imageOnly, avatarOnly, blockieOnly, size } = this.props
```
cazala, line 17: What's the purpose of `avatarOnly` and `blockieOnly`? isn't the same as using the `imageOnly` prop?

src/modules/collection/sagas.ts
```diff
@@ -129,16 +134,54 @@ function* handleSetCollectionMintersRequest(action: SetCollectionMintersRequestA
...
+    }
+
+    const txHash = yield call(() =>
+      implementation.methods
+        .setManagers(addresses, values)
+        .send({ from })
+        .getTxHash()
+    )
+
+    console.log('NEW MANAGERS', newManagers)
```
cazala, line 179: oopsie

src/components/Profile/Profile.tsx
nicosantangelo: The use case is a bit contrived and I didn't like this solutions but what I need to do is show only the blockie if it exists but show image + name if it's an avatar so I end up rendering the Profile like so:

`<Profile address={manager} blockieOnly={true} />`

You think it's too confusing?

cazala (approved):

PR #1081, opened 2020-09-24
feat: ENS for Land

malaniz:
# ENS for Lands
A new modal in land detail to assign a registered ENS.


src/lib/api/marketplace.ts
```diff
@@ -0,0 +1,42 @@
...
+    subdomain: string[]
+  }
+}
+
+type SubdomainQueryResult = {
+  nfts: SubdomainTuple[]
+}
+
+export class MarketplaceAPI {
+  fetchDomainList = async (_address: string): Promise<string[]> => {
```
nicosantangelo, line 29: is the `_` necessary? maybe we can just use `address`

src/lib/api/marketplace.ts
```diff
@@ -0,0 +1,42 @@
...
+    const owner = _address.toLowerCase()
+    const { data } = await graphClient.query<SubdomainQueryResult>({
+      query: getSubdomainQuery(),
+      variables: { owner }
+    })
+    console.log({data})
```
nicosantangelo, line 35: Whops, `console.log`!

src/lib/api/marketplace.ts
```diff
@@ -0,0 +1,42 @@
...
+    const domains: string[] = data.nfts.map(x => `${x.ens.subdomain}.dcl.eth`)
```
nicosantangelo, line 36: two small things:
- Maybe we can rename `x` to `nft` to be a bit more explicit
- We could just return this map

What do you think?

src/components/LandEnsPage/LandEnsForm/LandEnsForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const { onSetENS, land } = this.props
+    const { selectedSubdomain } = this.state
+    onSetENS(selectedSubdomain, land)
+    this.setState({done: true})
+  }
+  
+  render() {
+    const { land, ens, isLoading, error } = this.props
+    const { subdomainList, selectedSubdomain, done } = this.state
+    const selectOptions = subdomainList.map(x => ({value: x.toLowerCase(), text: x.toLowerCase()}))
```
nicosantangelo, line 52: Nitpick but, maybe we can rename `x` to `subdomain` to be a bit more explicit
What do you think?

src/lib/api/marketplace.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type SubdomainTuple = {
+  ens: {
+    subdomain: string[]
+  }
+}
+
+type SubdomainQueryResult = {
+  nfts: SubdomainTuple[]
+}
```
nicosantangelo, line 26: question for @cazala! for these `api` files we're keeping the types alongside the implementation right? there's no need to move this to it's own file

src/modules/ens/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const ENS_EMPTY_RESOLVER = '0x0000000000000000000000000000000000000000'
```
nicosantangelo, line 1: I don't think this is inherently bad, but what do you think about having a constant that's not related to ENS and it's just "EMPTY_ADDRESS" ?
We could also replace `components/Modals/ContestModal/ContestModal.tsx`s placeholder on line 163.

src/modules/ens/reducer.ts
```diff
@@ -0,0 +1,89 @@
...
+      }
+    }
+    case SET_ENS_SUCCESS: {
+      const { ens, data } = action.payload
+      
+      return {
+        ...state,
+        loading: loadingReducer(state.loading, action),
+        data: {
+          [ens] : data
```
nicosantangelo, line 73: Careful! I think you're missing the rest of the `state.data` object here, like:
```ts
data: {
  ...state.data,
  [ens] : data
}
```

if you don't do that, each time a new `ENS` succeeds it'll overwrite all other `ENS`es

*Same thing for line 48 above!*

src/components/LandEnsPage/LandEnsPage.container.ts
```diff
@@ -0,0 +1,21 @@
...
+import { MapStateProps, MapDispatchProps, MapDispatch } from './LandEnsPage.types'
+import LandEditPage from './LandEnsPage'
+import { getENSRequest, GET_ENS_REQUEST, SET_ENS_REQUEST, setENSRequest } from 'modules/ens/actions'
+import { isLoadingType } from 'decentraland-dapps/dist/modules/loading/selectors';
+import { getState, getLoading, getError } from 'modules/ens/selectors';
+import { Land } from 'modules/land/types';
+
+const mapState = (state: RootState): MapStateProps => ({
+  error: getError(state),
+  ens: getState(state),
```
nicosantangelo, line 12: We typically use the `data` as the value we pass to components instead of having to supply the entiry state (which also contains things like loading props and errors).
Maybe we can change it here and the components that receive it down the chain

src/types.d.ts
```diff
@@ -0,0 +1 @@
+declare module "*";
```
nicosantangelo, line 1: Why's this necessary?

src/modules/translation/languages/en.json
```diff
@@ -491,6 +492,21 @@
...
+    "select_names": {
+      "placeholder": "Your registered names",
+      "message": {
+        "choose_name": "Select a name for this land",
+        "name_available": "This name will work for your land",
+        "name_assigned": "This name has been assigned to this land"
+      }
+    },
+    "submit_message": {
+      "success": "Name set properly to your land"
```
nicosantangelo, line 507: @nachomazzara @cazala I'm confused but should this `land` be `Land` or `LAND` or neither?
Also, should we rename it on the Spanish translations? (it says `tierra` in multiple places)

src/modules/ens/sagas.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  const fromAddress = yield select(getAddress)
+  if (!fromAddress) {
+    throw new Error(`Invalid from address: ${fromAddress}`)
+  }
+
+  const from = Address.fromString(fromAddress)
+
+  return [eth, from]
+}
```
nicosantangelo, line 128: This is fine! but can we add a comment with a TODO that says:
`// TODO: use wallet/utils getCurrentAddress instead` ?

I've added that helper to avoid repeating code here https://github.com/decentraland/builder/blob/feat/collaborators/src/modules/wallet/utils.ts but it's not merged yet so there's no need to change this file yet

src/modules/ens/sagas.ts
```diff
@@ -0,0 +1,129 @@
...
+          type: 'EmptyResolver'
+        })
+      )
+    }
+
+    const resolverContract = new ENSResolver(eth, Address.fromString(ENS_RESOLVER_ADDRESS))
+    const ipfsHash = yield call(() => ipfs.uploadRedirectionFile(land))
+    const hash = contentHash.fromIpfs(ipfsHash)
+    const currentContent = yield call(() => resolverContract.methods.contenthash(nodehash).call())
+    console.log({ens, resolver: resolverAddress.toString(), ipfsHash, hash, currentContent})
```
nicosantangelo, line 82: Whops! console.log!

src/modules/ens/sagas.ts
```diff
@@ -0,0 +1,129 @@
...
+      resolverContract.methods
+        .setContenthash(nodehash, `0x${hash}`)
+        .send({from})
+        .getTxHash()
+    )
+    yield call(() => 
+      ensContract.methods
+        .setResolver(nodehash, Address.fromString(ENS_RESOLVER_ADDRESS))
+        .send({from})
+        .getTxHash()
```
nicosantangelo, line 51: I think we might want to keep track of these transactions right? show them on the /activity page and/or if they're loading to disable controls or show a notice. To do that we can use `buildTransactionPayload` like we do on `land/actions.ts`

@cazala @nachomazzara am I missing the mark? I confess I'm looking at the code without trying it out yet

nicosantangelo (changes requested):
Great work!! 🔥 
I have to test it a bit more but I'll close this review in case you want to start working on some of these things.

**Some general comments**
- I've noticed that the code is not consistently indented. We're using prettier for this and I think it's a good idea to have it `on save` with your editor of choice. You can also run it before pushing of course
- Something we probably forgot to tell you! we use semantic-release's way of naming branches. You *don't* have to rename this one! but it's a good practice to keep the tree consistent. In other projects like https://github.com/decentraland/decentraland-dapps the branch name determines which part of the lib version changes

src/modules/ens/sagas.ts
```diff
@@ -0,0 +1,129 @@
...
+  const { land, ens } = action.payload
+  try {
+    const [eth, from] = yield getEth()
+    const nodehash = namehash(ens)
+    const ensContract = new ENS(eth, Address.fromString(ENS_ADDRESS))
+    const resolverContract = new ENSResolver(eth, Address.fromString(ENS_RESOLVER_ADDRESS))
+
+    const ipfsHash = yield call(() => ipfs.uploadRedirectionFile(land))
+    const hash = contentHash.fromIpfs(ipfsHash)
+    yield call(() =>
```
nachomazzara, line 41: Is there any reason why we do this in a different order than ENS does? I think that they set the resolver first.  I remember that we have talked about it offline but not remember why we end up doing it this way 😄

src/modules/ens/sagas.ts
nachomazzara: IMO, we should separate the saga in two different transactions. One for setting the resolver and after confirmation (mined) the second one to set the content. Adding both to the activity page using `buildTransactionPayload`.

src/modules/ens/sagas.ts
```diff
@@ -0,0 +1,129 @@
...
+function* handleSetENSRequest(action: SetENSRequestAction) {
+  const { land, ens } = action.payload
+  try {
+    const [eth, from] = yield getEth()
+    const nodehash = namehash(ens)
+    const ensContract = new ENS(eth, Address.fromString(ENS_ADDRESS))
+    const resolverContract = new ENSResolver(eth, Address.fromString(ENS_RESOLVER_ADDRESS))
+
+    const ipfsHash = yield call(() => ipfs.uploadRedirectionFile(land))
+    const hash = contentHash.fromIpfs(ipfsHash)
```
nachomazzara, line 40: What do you think of moving `contentHash.fromIpfs(ipfsHash)` to a utils/helper file? You will be able to return just `0x${hash}` from there which is the only thing that you will end up using. Also, it seems that this piece of code is repeated below.

```javascript
const ipfsHash = yield call(() => ipfs.uploadRedirectionFile(land))
const hash = contentHash.fromIpfs(ipfsHash)
```

src/modules/ens/sagas.ts
```diff
@@ -0,0 +1,129 @@
...
+    yield call(() => 
+      resolverContract.methods
+        .setContenthash(nodehash, `0x${hash}`)
+        .send({from})
+        .getTxHash()
+    )
+    yield call(() =>
```
cazala, line 47: this is always sending the `setResolver` transaction isn't it? Shouldn't we do that only if the resolver is not set?

src/modules/ens/sagas.ts
cazala: I agree, the `setResolver` transaction should be handled in a separate saga, and only dispatch that saga before this one if the resolver needs to be set. That way wee can return both `txHash` in each success action and keep track of each transaction individually.

.env.example
```diff
@@ -24,13 +24,16 @@ REACT_APP_HOTJAR_ID=
...
+REACT_APP_IPFS_URL=
 
 # Contracts
 
 REACT_APP_MANA_TOKEN_CONTRACT_ADDRESS=0x2a8fd99c19271f4f04b1b7b9c4f7cf264b626edb
 REACT_APP_LAND_REGISTRY_CONTRACT_ADDRESS=0xf87e31492faf9a91b02ee0deaad50d51d56d5d4d
 REACT_APP_ESTATE_REGISTRY_CONTRACT_ADDRESS=0x959e104e1a4db6317fa58f8295f586e1a978c297
-
+REACT_APP_ENS_CONTRACT_ADDRESS=0x112234455c3a32fd11230c42e7bccd4a84e02010
+REACT_APP_RESOLVER_CONTRACT_ADDRESS=0x12299799a50340FB860D276805E78550cBaD3De3
```
cazala, line 36: We should add these 4 new env vars to the `.ci/.env.dev`, `.ci/.env.stg` and `.ci/.env.prod` files

src/modules/translation/languages/en.json
```diff
@@ -479,6 +479,7 @@
     "transfer": "Transfer",
     "edit": "Edit",
     "set_operator": "Set Operator",
+    "set_name": "Set name",
```
cazala, line 482: I think this should be renamed to something else like `Set ENS` or `Set link`, because each land already has a name, that can be set by going to the `Edit` page.

src/modules/translation/languages/en.json
cazala: Since the introduction of the Land tab in the builder we started using "Land" or "land" when we are taking about either parcels or estates. We are not using LAND anymore when talking about individual parcels, we use "Parcel". We use LAND when we talk about the token. That's as far as I understand it.

src/modules/translation/languages/en.json
```diff
@@ -491,6 +492,21 @@
     "total_land": "Total Land",
     "empty_land": "Empty Land"
   },
+  "land_ens_page" : {
+    "title": "Set Name",
+    "subtitle": "Set the ethereum name service to your land",
```
cazala, line 497: I think the `ethereum name service` part is not very user friendly, maybe something more literal like: `Assign one of your Names to act as a link to your land.` ?

package.json
```diff
@@ -45,6 +46,7 @@
     "eslint-plugin-import": "2.14.0",
     "eslint-plugin-jsx-a11y": "6.1.2",
     "eslint-plugin-react": "7.11.1",
+    "ethers": "^5.0.12",
```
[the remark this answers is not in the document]
malaniz: Sorry! It is not necesary now.

src/components/LandEnsPage/LandEnsForm/LandEnsForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+  componentDidMount() {
+    (async () => {
+     const state = store.getState()
+     const address = getAddress(state)
+     
+     if (address) {
+       const subdomainList = await marketplace.fetchDomainList(address)
+       this.setState({subdomainList})
+     }
+    })()
```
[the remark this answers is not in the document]
malaniz: Done!

src/components/LandEnsPage/LandEnsForm/LandEnsForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+export default class LandEnsForm extends React.PureComponent<Props, State> {
+  state: State = {
+    selectedSubdomain: '',
+    subdomainList: [],
+    done: false
+  }
+
+  componentDidMount() {
+    (async () => {
+     const state = store.getState()
```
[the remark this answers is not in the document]
malaniz: Done! thanks!

src/components/LandEnsPage/LandEnsForm/LandEnsForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { Props, State } from './LandEnsForm.types'
+import './LandEnsForm.css'
+import { marketplace } from 'lib/api/marketplace'
+import { store } from 'modules/common/store'
+import { getAddress } from 'decentraland-dapps/dist/modules/wallet/selectors'
+import { env } from 'decentraland-commons'
+import { SelectNames } from 'components/SelectNames'
+
+export const ENS_CONTRACT_ADDRESS = env.get('REACT_APP_ENS_CONTRACT_ADDRESS', '')
+export const RESOLVER_CONTRACT_ADDRESS = env.get('REACT_APP_RESOLVER_CONTRACT_ADDRESS', '')
```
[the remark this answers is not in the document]
malaniz: typo from old code. Thanks!

src/lib/api/marketplace.ts
malaniz: oops! fixed!.

src/modules/ens/sagas.ts
malaniz: Done! thanks for the warning!

src/modules/ens/sagas.ts
malaniz: Fixed! thanks!

src/components/LandEnsPage/LandEnsForm/LandEnsForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+        case 'EqualContent': return {
+          message: t('land_ens_page.select_names.message.name_assigned'),
+          disable: true
+        }
+        default: return {
+          message: t('land_ens_page.select_names.message.choose_name'),
+          disable: true
+        }
+      }
+    })(messageType)
```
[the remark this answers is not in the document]
malaniz: Thanks for the detailed review! completely agree with you. I've changed the code with your feedback.

src/components/LandEnsPage/LandEnsForm/LandEnsForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const { subdomainList, selectedSubdomain, done } = this.state
+    const selectOptions = subdomainList.map(x => ({value: x.toLowerCase(), text: x.toLowerCase()}))
+ 
+ 
+    if ( isLoading) {
+      return <Row><Loader size="large" active /></Row>
+    }
+
+    const messageType = ens.data[selectedSubdomain] ? ens.data[selectedSubdomain].type : 'default'
+    const { message: selectMessage, disable: disableButton } = ((type) => {
```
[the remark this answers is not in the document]
malaniz: I've changed the `disableButton` with `isDisabledButton`. Thanks for the feedback!

src/components/LandEnsPage/LandEnsPage.container.ts
```diff
@@ -0,0 +1,21 @@
...
+  isLoading: isLoadingType(getLoading(state), SET_ENS_REQUEST) || isLoadingType(getLoading(state), GET_ENS_REQUEST)
+})
+
+const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
+  onSetENS: (ens:string, land:Land) => dispatch(setENSRequest(ens, land)),
+  onGetENS: (ens:string, land:Land) => dispatch(getENSRequest(ens, land))
```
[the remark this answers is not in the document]
malaniz: Done! thanks!

src/components/SelectNames/SelectNames.css
```diff
@@ -0,0 +1,28 @@
...
+@media (max-width: 768px) {
+  .SelectNames {
+    margin-bottom: 24px;
+    width: 100%;
+  }
+
+  .SelectNames .ui.sub.header.name {
+    font-weight: normal;
+  }
+}
```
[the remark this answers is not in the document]
malaniz: I'm not sure to be sincere. I just use the `Select` component from marketplace using the same styling.

src/lib/api/ipfs.ts
```diff
@@ -0,0 +1,42 @@
+import { env } from 'decentraland-commons'
+import { Land } from 'modules/land/types'
+import { getSelection, getCenter, coordsToId } from 'modules/land/utils'
+
+export const IPFS_URL = env.get('REACT_APP_IPFS_URL', '')
```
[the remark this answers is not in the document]
malaniz: Fixed! thanks!

src/lib/api/marketplace.ts
malaniz: Totally agree. Fixed

src/components/LandEnsPage/LandEnsForm/LandEnsForm.tsx
malaniz: Totally agree. Fixed

src/types.d.ts
malaniz: This is necessary for 
```import * as contentHash from 'content-hash'``

File: src/components/LandEnsPage/LandEnsForm/LandEnsForm.tsx (hunk 21 above)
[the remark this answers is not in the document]
Comment (malaniz) [reply src/components/LandEnsPage/LandEnsForm/LandEnsForm.tsx]: Thanks I've changed this completely using a saga instead of calling the marketplace here.

File: src/components/LandEnsPage/LandEnsForm/LandEnsForm.tsx [hunk 29]
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  componentDidMount() {
+    (async () => {
+     const state = store.getState()
+     const address = getAddress(state)
+     
+     if (address) {
+       const subdomainList = await marketplace.fetchDomainList(address)
+       this.setState({subdomainList})
```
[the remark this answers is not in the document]
malaniz: This changed to use a saga and this subdomainList is received from props

src/modules/ens/sagas.ts
malaniz: Thanks for remembered me. fixed!

src/types.d.ts
nachomazzara: I'm not an expert but try using `const contentHash = require('content-hash')`. @nicosantangelo may have a better workaround

src/modules/ens/sagas.ts
malaniz: If it have another resolver and the user want to change it to our resolver, then we need to change it. Or am I missing something?

src/types.d.ts
malaniz: I've changed this line to 
```
const contentHash = require('content-hash')
```
To not need this file.

src/modules/translation/languages/en.json
malaniz: Done! thanks!

src/modules/translation/languages/en.json
malaniz: Done! thanks!

src/modules/ens/sagas.ts
malaniz: Agree with using 2 sagas, in some point I made it without `buildTransactionPayload` on purpose. I didn't find a way to combine two txHash in one saga to use it as payload in the action.  
*Working on it!*

src/components/LandEnsPage/LandEnsForm/LandEnsForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+      <Form className="LandEnsForm">
+        { 
+          done ? <>
+            <Row>
+              <p style={{paddingTop: '20px'}}> {submitMessage} </p>
+            </Row>
+            <Row>
+              <Link className="cancel" to={locations.landDetail(land.id)}>
+                <Button>{t('global.done')}</Button>
+              </Link>
```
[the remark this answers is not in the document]
malaniz: I know. I'm waiting the designs to know what would be better for type of things.

src/lib/api/ipfs.ts
```diff
@@ -0,0 +1,42 @@
...
+    <head>
+      <meta 
+        http-equiv="refresh" 
+        content="0; URL=https://play.decentraland.org?position=${coordsToId(x, y)}" 
+      />
+    </head>
+    <body>
+      <p>
+        If you are not redirected 
+        <a href="https://play.decentraland.org?position=${coordsToId(x, y)}">click here</a>.
```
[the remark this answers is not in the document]
malaniz: @nicosantangelo should be an existing variable? or a new one?

src/modules/ens/reducer.ts
malaniz: done! thanks!

src/modules/ens/sagas.ts
nachomazzara: The resolver is the _public_ one. Is not _our_. We can check if the user is using the public resolver, if not. We can send the tx to change it.

nicosantangelo:
Feedback on the webapp

## Copies

1)
I think the text we're using is not very consistent. Using the feature I see
- Set **Link**
- Assign/claim **Name**
- **Owner** (once the name is assigned)
I think we should at least use the same concept everywhere, no? I first think I'm setting the link, next the name for the land and then *my* name as an owner.

2)
> You don't have any name claimed, so first you'll need to claim one.
I understand what you're trying to say but it's a bit weirdly worded. Maybe "You don't have any available names. To assign a name to you LAND, you first need need to claim it"

3)
> `This name will work for your land`
What's `work` in this context? is that hole description necessary? I feel maybe you can keep the first notice as an empty state for the select and remove it
image

4)
> Send TX
I think we're avoiding talking about transactions in the rest of the app. Moreover I don't think using the TX acronym is a good idea. Maybe we can put the actual action on the buton?
Like using **Content** as a title and `SET CONTENT` on the button?

## Features

1)
image

The `CLAIM A NEW NAME` button does nothing

2) 
If the first transaction fails the UI says `✔ Done` and allows me to continue

3)
> Assign a name to ${LAND_NAME}
Missing link on the subtitle, LAND_NAME should link to the land

## Styles

The UI looks both too clumped together and the boxes too large. I think we should try to get it as close to the design as possible:

**Branch**
image

**Designs**
image

.ci/.env.stg
```diff
@@ -20,3 +20,6 @@ REACT_APP_AUTH0_REDIRECT=https://builder.decentraland.net/callback
 REACT_APP_MANA_TOKEN_CONTRACT_ADDRESS=0x0f5d2fb29fb7d3cfee444a200298f468908cc942
 REACT_APP_LAND_REGISTRY_CONTRACT_ADDRESS=0xf87e31492faf9a91b02ee0deaad50d51d56d5d4d
 REACT_APP_ESTATE_REGISTRY_CONTRACT_ADDRESS=0x959e104e1a4db6317fa58f8295f586e1a978c297
+REACT_APP_IPFS_URL=https://ipfs.infura.io:5001/api/v0/add?pin=false
+REACT_APP_ENS_CONTRACT_ADDRESS=0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e
+REACT_APP_RESOLVER_CONTRACT_ADDRESS=0x4976fb03c32e5b8cfe2b6ccb31c09ba78ebaba41
```
cazala, line 25: This is missing the `REACT_APP_MARKETPLACE_GRAPH_URL` var

.ci/.env.prd
```diff
@@ -19,4 +19,8 @@ REACT_APP_AUTH0_DOMAIN=decentraland.auth0.com
 REACT_APP_AUTH0_REDIRECT=https://builder.decentraland.org/callback
 REACT_APP_MANA_TOKEN_CONTRACT_ADDRESS=0x0f5d2fb29fb7d3cfee444a200298f468908cc942
 REACT_APP_LAND_REGISTRY_CONTRACT_ADDRESS=0xf87e31492faf9a91b02ee0deaad50d51d56d5d4d
-REACT_APP_ESTATE_REGISTRY_CONTRACT_ADDRESS=0x959e104e1a4db6317fa58f8295f586e1a978c297
\ No newline at end of file
+REACT_APP_ESTATE_REGISTRY_CONTRACT_ADDRESS=0x959e104e1a4db6317fa58f8295f586e1a978c297
+REACT_APP_MARKETPLACE_GRAPH_URL=https://api.thegraph.com/subgraphs/name/decentraland/marketplace-ropsten
```
cazala, line 23: For production this var should be `https://api.thegraph.com/subgraphs/name/decentraland/marketplace` (without the `-ropsten` at the end)

cazala (changes requested):

src/components/LandEnsPage/LandSetNameForm/LandSetNameForm.tsx
```diff
@@ -0,0 +1,154 @@
...
+        buttonSetResolver = <> { t('global.retry_tx') } </>
+        isSetResolverButtonDisabled = false
+        isSetContentButtonDisabled = true
+        setResolverButtonClassName = ''
+      } else if (error.origin === 'SET_ENS_CONTENT') {
+        buttonSetContent =  <> { t('global.retry_tx') } </>
+        isSetContentButtonDisabled = false
+        setContentButtonClassName = ''
+      }
+    }
```
nicosantangelo, line 100: I think this code could be extracted a bit. Let me know if you think these suggestions make sense:
1. We could name a few of these conjoined conditions and then reuse them. For example: 
```ts
// Test code, do not copy :P
isLoading() {
  return isLoading || isWaitingTxSetResolver || isWaitingTxSetContent
}
(...)
const isSubmitDisabled = !(isSetContentDisabled && isSetResolverDisabled) || this.isLoading()
let isSetResolverButtonDisabled = isSetResolverDisabled || this.isLoading()
let isSetContentButtonDisabled = isSetContentDisabled  || this.isLoading()
```

2. Commented below as well but it's a good idea to just use the actual action here instead of the string that represents it, so if the action changes, this code will still work and wont fail silently (`'SET_ENS_RESOLVER' => SET_ENS_RESOLVER_FAILURE` for example)

3. I'm a bit confused by the use of the `state` here. Shouldn't it all come from redux? why are you setting it here also? Correct me if I'm wrong but it seems like you have all the information you need on the state and you just have to feed it to the container?

4. **This is just a suggestion i have no idea if it's a good idea**: If you're having two "boxes" (Resolver and Content) that have similar behaviours but different sources for data, maybe you're looking at a missing abstraction. You could have a `TransactionBox` Component (maybe a bad name) which is agnostic to the situation but does all the heavy lifting, and you can then feed the booleans from here.

nicosantangelo (changes requested):
🚀

I still see that the code lacks prettier! remember to run it whenever you can. You might want to have it as an on-save trigger on your editor of choice.

malaniz (commented):
Thanks for testing again all the code and the detailed feedback!

src/modules/ens/reducer.ts
```diff
@@ -0,0 +1,160 @@
...
+        ...state,
+        loading: loadingReducer(state.loading, action)
+      }
+    }
+    case GET_DOMAINLIST_SUCCESS: {
+      const { data } = action.payload
+      return {
+        ...state,
+        loading: loadingReducer(state.loading, action),
+        subdomainList: data
```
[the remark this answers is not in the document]
malaniz: The `data` is just the result: a list of domain names owned by the current account. There is no need to copy previous result. The same for the error, just an object with error message, a code and an origin for the error. But always we need to update all the object.

src/modules/ens/sagas.ts
```diff
@@ -0,0 +1,158 @@
...
+    const resolverContract = new ENSResolver(eth, Address.fromString(ENS_RESOLVER_ADDRESS))
+    const ipfsHash = yield call(() => ipfs.uploadRedirectionFile(land))
+    const hash = contentHash.fromIpfs(ipfsHash)
+    const currentContent = yield call(() => resolverContract.methods.contenthash(nodehash).call())
+    if (currentContent === ENS_EMPTY_CONTENT) {
+      return yield put(
+        getENSSuccess(ens, {
+          resolver: ENS_RESOLVER_ADDRESS,
+          content: currentContent,
+          type: 'EmptyContent'
```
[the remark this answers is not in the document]
malaniz: Done thanks

src/components/LandEnsPage/LandSetNameForm/LandSetNameForm.css
```diff
@@ -0,0 +1,43 @@
...
+  padding: 10px 20px 10px 20px;
+  margin: 50px
+}
+
+.grey-button {
+  background-color: #444 !important;
+}
+i.check.icon {
+  padding-left: 10px;
+}
```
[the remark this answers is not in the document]
malaniz: Done thanks

cazala (approved):

.env.example
```diff
@@ -24,12 +24,17 @@ REACT_APP_HOTJAR_ID=
 REACT_APP_INTERCOM_APP_ID=
 REACT_APP_ROLLBAR_ACCESS_TOKEN=
 REACT_APP_DELIGHTED_API_KEY=
+REACT_APP_MARKETPLACE_GRAPH_URL=https://api.thegraph.com/subgraphs/name/decentraland/marketplace
+REACT_APP_IPFS_URL=https://ipfs.infura.io:5001/api/v0/add?pin=false
+REACT_APP_CLAIM_NAME_URL=https://avatars.decentraland.org
```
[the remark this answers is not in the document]
malaniz: Thanks @cazala !

.ci/.env.dev
```diff
@@ -19,4 +19,10 @@ REACT_APP_AUTH0_DOMAIN=dcl-test.auth0.com
 REACT_APP_AUTH0_REDIRECT=https://builder.decentraland.io/callback
 REACT_APP_MANA_TOKEN_CONTRACT_ADDRESS=0x2a8fd99c19271f4f04b1b7b9c4f7cf264b626edb
 REACT_APP_LAND_REGISTRY_CONTRACT_ADDRESS=0x7a73483784ab79257bb11b96fd62a2c3ae4fb75b
-REACT_APP_ESTATE_REGISTRY_CONTRACT_ADDRESS=0x124bf28a423b2ca80b3846c3aa0eb944fe7ebb95
\ No newline at end of file
+REACT_APP_ESTATE_REGISTRY_CONTRACT_ADDRESS=0x124bf28a423b2ca80b3846c3aa0eb944fe7ebb95
+REACT_APP_MARKETPLACE_GRAPH_URL=https://api.thegraph.com/subgraphs/name/decentraland/marketplace-ropsten
+REACT_APP_IPFS_URL=https://ipfs.infura.io:5001/api/v0/add?pin=false
+REACT_APP_LAND_POSITION_URL=https://play.decentraland.org?position=
```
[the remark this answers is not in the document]
malaniz: Good eye! thanks!

nicosantangelo (approved):

PR #1083, opened 2020-09-25
feat: added right panel

### LINKED ISSUE #1031: Editor right panel:
show selected item details
allow user to edit details

cazala:
Closes #1031


src/components/ItemEditorPage/RightPanel/Input/Input.types.ts
```diff
@@ -0,0 +1,10 @@
+export type Props = {
+  itemId: string
+  label: string
+  value: string | null | undefined
```
nicosantangelo, line 4: Question: Is this type necessary? couldn't we just use `value?` here?

src/components/ItemEditorPage/RightPanel/Tags/Tags.tsx
```diff
@@ -0,0 +1,93 @@
+import * as React from 'react'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+import { Props, State } from './Tags.types'
+import './Tags.css'
+
+const WHITE_SPACE_KEY_CODE = 32
+const ENTER_KEY_CODE = 13
+const COMMA_KEY_CODE = 118
+const BACKSPACE_KEY_CODE = 8
+const DELETE_KEY_CODE = 46
```
nicosantangelo, line 10: It seems like these constants should live elsewhere, no? It seems out of place on the Tags Component

What do you think?

nicosantangelo (commented):
Left two silly comments, the code looks great overall and the PR works. However the I'm able to do actions on published items that I shouldn't be able to. 

I fixed few of these cases on https://github.com/decentraland/builder/pull/1086, like not being able to delete a published item.

So, is up to you, if you want we could leave this PR as-is and check those cases on a separate issue(s).

cazala:
@nicosantangelo thanks for the feedback! I pushed the changes in the comments, and as discussed offline, lets leave the handling of published items for another PR

nicosantangelo:
woot woot

PR #1108, opened 2020-10-06
feat: post items to the content server

### LINKED ISSUE #1092: Upload item contents to the content server (consider `dcl-catalyst-client`)

nicosantangelo:
Closes #1092

### DIFF
--- a/src/modules/collection/sagas.ts
+++ b/src/modules/collection/sagas.ts
@@ -42,15 +42,13 @@ import { getCurrentAddress } from 'modules/wallet/utils'
 import { ERC721_COLLECTION_FACTORY_ADDRESS, ERC721_COLLECTION_ADDRESS } from 'modules/common/contracts'
 import { ERC721CollectionFactoryV2 } from 'contracts/ERC721CollectionFactoryV2'
 import { ERC721CollectionV2 } from 'contracts/ERC721CollectionV2'
-import { setItemsTokenIdRequest, FETCH_ITEMS_SUCCESS } from 'modules/item/actions'
+import { setItemsTokenIdRequest, deployItemContentsRequest, FETCH_ITEMS_SUCCESS } from 'modules/item/actions'
 import { locations } from 'routing/locations'
 import { getCollectionId } from 'modules/location/selectors'
 import { builder } from 'lib/api/builder'
 import { closeModal } from 'modules/modal/actions'
 import { Item } from 'modules/item/types'
 import { getItems } from 'modules/item/selectors'
-import { getIdentity } from 'modules/identity/utils'
-import { deployItemContents } from './export'
 import { getCollection, getCollectionItems } from './selectors'
 import { initializeCollection } from './utils'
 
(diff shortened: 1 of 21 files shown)


cazala (changes requested):
I found a few issues while trying to publish a collection:

1) When a collection is created, it doesn't have a `salt` yet, so when I click on "publish" it fails with the following error: 

```
"invalid bytes32 value (arg="_salt", coderType="bytes32", value=undefined, version=1)"
```

In order to make it work I need to refresh, and once the collection comes from the server it does have a `salt` and the publish transaction can be sent.

2) When the transaction is mined, the app tries to initiate the deployment of the items' contents to the catalyst, but since the items in memory don't have yet a `tokenId` it fails with the following error:

```
Error: You need the collection and item to be published to get the catalyst entity id
    at getCatalystEntityId (utils.ts:24)
```

I need to refresh and then the app resumes the upload of the contents and this time it works since they have `tokenId`.

I have one suggestion regarding the current implementation: I would rather have specific actions to upload the contents of an item, ie: `UPLOAD_ITEM_CONTENTS_REQUEST/SUCCESS/FAILURE` with its corresponding saga, instead of doing it inside the transaction and collections sagas, so we can easily then add visual feedback to the items that are being uploaded (like showing the blinking cloud icon we use when a scene is being uploaded). I think it also improves the debugging experienc, cos currently these uploads don't show up in the redux feed as successful or failed actions, it is hidden within other actions.

nicosantangelo:
Agreed on everything. I fixed the issues and added the action. LMK if you like the implementation

src/modules/item/sagas.ts
```diff
@@ -154,6 +161,24 @@ function* handleSetItemsTokenIdRequest(action: SetItemsTokenIdRequestAction) {
...
+function* handleDeployItemContentsRequest(action: DeployItemContentsRequestAction) {
+  const { collection, item } = action.payload
+
+  try {
+    const identity = yield getIdentity()
+    if (!identity) {
+      throw new Error('Invalid identity')
+    }
+
+    console.log('DEPLOYING ITEM CONTENTS', item, 'in catalyst?', item.inCatalyst)
```
cazala, line 173: oops 2 :)

src/modules/collection/sagas.ts
```diff
@@ -226,3 +236,50 @@ function* handleMintColectionItems(action: MintCollectionItemsRequestAction) {
...
+    yield take(FETCH_ITEMS_SUCCESS)
+  }
+
+  try {
+    const { collections } = action.payload
+
+    for (const collection of collections) {
+      if (!collection.isPublished) continue
+      for (const item of allItems) {
+        if (item.collectionId === collection.id) {
```
cazala, line 277: here i'd add `&& !item.inCatalyst` so we prevent sending an unnecessary upload request

src/modules/collection/sagas.ts
```diff
@@ -226,3 +236,50 @@ function* handleMintColectionItems(action: MintCollectionItemsRequestAction) {
...
+
+  try {
+    switch (transaction.actionType) {
+      case PUBLISH_COLLECTION_SUCCESS: {
+        // We re-fetch the collection from the store to get the updated version
+        const collectionId = transaction.payload.collection.id
+        const collection = yield select(state => getCollection(state, collectionId))
+        const items = yield select(state => getCollectionItems(state, collectionId))
+
+        for (const item of items) {
```
cazala, line 251: I'd filter out the items that are already `inCatalyst` to prevent sending an unnecessary upload request

src/modules/collection/sagas.ts
nicosantangelo: Thought about it but wasn't sure, gotcha!

src/modules/collection/sagas.ts
nicosantangelo: Thought about it but wasn't sure, gotcha!

cazala (approved):

PR #1112, opened 2020-10-14
feat: share editor urls

nicosantangelo:
Depends on https://github.com/decentraland/builder-server/pull/127

### DIFF
--- a/src/components/CollectionProvider/CollectionProvider.container.ts
+++ b/src/components/CollectionProvider/CollectionProvider.container.ts
@@ -1,0 +1,29 @@
+import { connect } from 'react-redux'
+import { isLoadingType } from 'decentraland-dapps/dist/modules/loading/selectors'
+import { isConnecting } from 'decentraland-dapps/dist/modules/wallet/selectors'
+import { RootState } from 'modules/common/types'
+import { getCollectionId } from 'modules/location/selectors'
+import { isLoggingIn } from 'modules/identity/selectors'
+import { getLoading, getCollections, getCollectionItems } from 'modules/collection/selectors'
+import { FETCH_COLLECTION_REQUEST, fetchCollectionRequest } from 'modules/collection/actions'
+import { MapStateProps, MapDispatchProps, MapDispatch, OwnProps } from './CollectionProvider.types'
+import CollectionProvider from './CollectionProvider'
+
+const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
+  const id = ownProps.id || getCollectionId(state)
+  const collections = getCollections(state)
+  const collection = collections.find(collection => collection.id === id) || null
+  const items = collection ? getCollectionItems(state, collection.id) : []
+  return {
+    id,
+    collection,
+    items,
+    isLoading: isLoadingType(getLoading(state), FETCH_COLLECTION_REQUEST) || isLoggingIn(state) || isConnecting(state)
+  }
+}
+
+const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
+  onFetchCollection: id => dispatch(fetchCollectionRequest(id))
+})
+
+export default connect(mapState, mapDispatch)(CollectionProvider)
--- a/src/components/CollectionProvider/CollectionProvider.container.ts
+++ b/src/components/CollectionProvider/CollectionProvider.container.ts
@@ -4,15 +4,14 @@ import { isConnecting } from 'decentraland-dapps/dist/modules/wallet/selectors'
 import { RootState } from 'modules/common/types'
 import { getCollectionId } from 'modules/location/selectors'
 import { isLoggingIn } from 'modules/identity/selectors'
-import { getLoading, getCollections, getCollectionItems } from 'modules/collection/selectors'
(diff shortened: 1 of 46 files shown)


src/components/CollectionProvider/CollectionProvider.container.ts
```diff
@@ -0,0 +1,29 @@
...
+import { getCollectionId } from 'modules/location/selectors'
+import { isLoggingIn } from 'modules/identity/selectors'
+import { getLoading, getCollections, getCollectionItems } from 'modules/collection/selectors'
+import { FETCH_COLLECTION_REQUEST, fetchCollectionRequest } from 'modules/collection/actions'
+import { MapStateProps, MapDispatchProps, MapDispatch, OwnProps } from './CollectionProvider.types'
+import CollectionProvider from './CollectionProvider'
+
+const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
+  const id = ownProps.id || getCollectionId(state)
+  const collections = getCollections(state)
```
cazala, line 14: doesn't this only return collections owned by the logged in user? If i'm logged in as a different user as the owner, wouldn't this always fail to find the collection i'm looking for?

src/components/CollectionProvider/CollectionProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+  componentDidMount() {
+    const { id, collection, onFetchCollection } = this.props
+    if (id && !collection) {
+      onFetchCollection(id)
+    }
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    const { id, onFetchCollection } = this.props
+    if (id && id !== prevProps.id) {
```
cazala, line 14: should we check if the collection is in memory already before triggering the fetch here?

src/components/ItemProvider/ItemProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+    const { id, item, onFetchItem } = this.props
+    if (id && !item) {
+      onFetchItem(id)
+    }
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    const { id, onFetchItem } = this.props
+    if (id && id !== prevProps.id) {
+      onFetchItem(id)
```
cazala, line 15: same question as in `CollectionProvider`

src/modules/item/reducer.ts
```diff
@@ -127,7 +145,35 @@ export function itemReducer(state: ItemState = INITIAL_STATE, action: ItemReduce
...
-          [item.id]: { ...item }
+          ...toItemObject([item])
+        },
+        loading: loadingReducer(state.loading, action),
+        error: null
+      }
+    }
+    case FETCH_COLLECTIONS_SUCCESS: {
+      const { collections } = action.payload
+      const items = ([] as Item[]).concat.apply(
```
cazala, line 156: what do you think about 

```ts
collections.reduce<Item>((items, collection) => items.concat(collection.items), [])
``` 

?

src/components/CollectionProvider/CollectionProvider.container.ts
nicosantangelo: Yes! I was testing with an specific case and forgot

src/modules/item/reducer.ts
nicosantangelo: Much better, I let my inner js ninja come out, I'm sorry

cazala (changes requested):
When i navigate straight into an item editor url, I'm seeing the request to fetch the item and collection failing because they are sent before the wallet is connected, causing the request to be sent without an auth chain, and returning as 401s and 404s

no-auth-chain

Maybe this is not a frontend issue tho, because I don't see why the sever would need the auth chain to return an item or a collection, the only thing needed should be the id which is already in the path

cazala:
The requests failing above are causing the editor to be empty:

Screen Shot 2020-10-15 at 6 06 39 PM

src/components/CollectionProvider/CollectionProvider.tsx
nicosantangelo: I thought about it, but I'm not sure. I left it out because I maybe updating the local version could be useful? what do you think?

nicosantangelo:
Hm, I left `withAuthentication` on the server so at least a wallet is connected when asking for the collection/items, even if the address is not being used. If that's not a concern I can simply remove the middleware
If not I should check the wallet finished loading before sending the fetch

What do you think?

cazala:
mm i think we will eventually need those endpoints to work unathenticated to support share functionality don't we? Also i noticed that only the `/collection/:id` endpoint has the `withAuthorization` middleware, but the `/item/:id` does not, and it's returning a 404 instead... even tho the item does exist. Maybe the `withModelExist` depends on `withAuthorization` and if blowing up resulting in a 404?

src/components/CollectionProvider/CollectionProvider.tsx
cazala: good point, I guess it doesn't hurt to refresh it

cazala (approved):

PR #1120, opened 2020-10-21
feat: added image with rarity background to item deployment

### LINKED ISSUE #1109: Add the image to the Item
The props should now be:
- thumbnail: 256x256 transparent background
- image: 1024x1025 with the rarity color background

cazala:
Closes #1109 

Also fixed issue where collections where uploaded with `items` and was breaking the server's schema

### DIFF
--- a/src/lib/api/builder.ts
+++ b/src/lib/api/builder.ts
@@ -301,10 +301,10 @@ function fromRemoteItem(remoteItem: RemoteItem) {
   if (remoteItem.rarity) item.rarity = remoteItem.rarity
   if (remoteItem.total_supply !== null) item.totalSupply = remoteItem.total_supply // 0 is false
   if (remoteItem.collection) {
-    collection = fromRemoteCollection(utils.omit(remoteItem.collection, ['items']))[0]
+    collection = fromRemoteCollection(utils.omit(remoteItem.collection, ['items'])).collection
   }
 
-  return [item, collection] as const
+  return { item, collection }
 }
 
 function toRemoteCollection(collection: Collection): RemoteCollection {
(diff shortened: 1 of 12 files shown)


src/lib/api/builder.ts
```diff
@@ -314,35 +318,35 @@ function toRemoteCollection(collection: Collection): RemoteCollection {
...
     updatedAt: +new Date(remoteCollection.updated_at)
   }
 
+  let items: Item[] = []
+
   if (remoteCollection.salt) collection.salt = remoteCollection.salt
   if (remoteCollection.contract_address) collection.contractAddress = remoteCollection.contract_address
   if (remoteCollection.items) {
-    collection.items = remoteCollection.items.map(item => fromRemoteItem(utils.omit(item, ['collection'])))
+    items = remoteCollection.items.map(item => fromRemoteItem(utils.omit(item, ['collection']))[0])
```
nicosantangelo, line 346: I know the rest of the app is using arrays to return multiple values but I was wondering, why not just returning an object?
```diff
- [item, collection]
+ { item, collection }
```

that'd allow for doing this: `fromRemoteItem(utils.omit(item, ['collection'])).item` instead of having to figure out what `[0]` is in that context

I also ask because it's not really obvious that `fromRemoteItem` returns two values

src/modules/item/utils.ts
```diff
@@ -108,3 +120,39 @@ export function toItemObject(items: Item[]) {
...
+  const canvas = document.createElement('canvas')
+  canvas.width = width
+  canvas.height = height
+  const context = canvas.getContext('2d')
+
+  // fail
+  if (!context || !item.rarity) return thumbnail
+
+  // render gradient
+  const gradient = context.createRadialGradient(width / 2, height / 2, 0, width / 2, height / 2, width / 1.75)
```
nicosantangelo, line 139: 🚀

nicosantangelo (commented):
Only a tiny winy comment

cazala:
thanks for the review @nicosantangelo, I applied the feedback

PR #1124, opened 2020-10-23
feat: mask thumbnail

cazala:
Takes a frontal thumbnail for mask, eyes, and eyebrow categories. Also set category and rarity at creation time


src/lib/getModelData.ts
```diff
@@ -24,13 +24,13 @@ type Options = {
   width: number
   height: number
   mappings?: Record<string, string>
-  thumbnailType: '3d' | '2d'
+  thumbnailType: 'default' | 'top' | 'front'
```
nicosantangelo, line 27: You think it might be a good idea to add this to an enum and name it? It's easier to type everywhere and to edit (if necessary in the future)

nicosantangelo (commented):
I've noticed that creating a new item outside the editor won't bring up the category/rarity dropdowns. Is that intended?

Also, I think it's a good idea show all inputs at once and keep the name on top, so it's not obscured by the dropdown options and it's immediately visible (I didn't see it the first time)
Keeping all inputs visible would avoid having to resize the modal when you pick something.

What do you think?

src/lib/getModelData.ts
cazala: done 👍

nicosantangelo:
Apparently it was just me! merging

nicosantangelo (approved):

PR #1125, opened 2020-10-23
fix: left panel scroll and item image

cazala:
adds scroll to the left panel to prevent this:

Screen Shot 2020-10-23 at 5 30 34 PM

Also fixes the item image that looked stretched when there were 3 items:

Screen Shot 2020-10-23 at 5 29 53 PM

I had to set a max-width to prevent this:

Screen Shot 2020-10-23 at 5 34 20 PM


nicosantangelo:
Why would you ever prevent this 
image

nicosantangelo (approved):

PR #1145, opened 2020-10-27
feat: Feature Flag for ENS

malaniz:
Feature Flag for ENS

If REACT_APP_FF_ENS exist then will be turned on, if not defined will be turned OFF.


cazala (changes requested):
i think we should also add this env var to the `now.json` file. Also maybe there are other missing env vars from the ens branch that we should add there so the feature works on the vercel previews

src/components/LandDetailPage/LandDetailPage.tsx
```diff
@@ -11,6 +11,7 @@ import Profile from 'components/Profile'
 import Scene from './Scene'
 import { Props, State } from './LandDetailPage.types'
 import './LandDetailPage.css'
+import { env } from 'decentraland-commons'
```
cazala, line 14: nip tick: cam we move this import to the top? we usually have the modules on top and the relative imports below

cazala (approved):

src/components/LandDetailPage/LandDetailPage.tsx
```diff
@@ -146,7 +149,12 @@ export default class LandDetailPage extends React.PureComponent<Props, State> {
                             text={t('land_detail_page.set_operator')}
                             onClick={() => onNavigate(locations.landOperator(land.id))}
                           />
-                          <Dropdown.Item text={t('land_detail_page.set_link')} onClick={() => onNavigate(locations.landEns(land.id))} />
+                          {FF_ENS ? (
```
nicosantangelo, line 152: Sorry I saw this one second, but could we change it to look the same as the one juanca did? here https://github.com/decentraland/builder/pull/1151/files#diff-e6f76a45490df7734399812bad8e19dda9c880883a47f3101bf25025124d65afR20

it's super minor, but for consistency sake

src/routing/Routes.tsx
```diff
@@ -79,7 +81,7 @@ export default class Routes extends React.Component<Props, State> {
             <Route exact path={locations.land()} component={LandPage} />
             <Route exact path={locations.landDetail()} component={LandDetailPage} />
             <Route exact path={locations.landTransfer()} component={LandTransferPage} />
-            <Route exact path={locations.landEns()} component={LandEnsPage} />
+            { FF_ENS && <Route exact path={locations.landEns()} component={LandEnsPage} /> }
```
nicosantangelo, line 84: Same comment as above

nicosantangelo (approved):
LGTM super minor comment. I'll approve anyways

nicosantangelo (approved):
🔥

PR #1148, opened 2020-10-27
fix: avatar page grid

cazala:
Fixes the avatar page grid alignment, now it fits 4 elements per row:

Screen Shot 2020-10-27 at 6 39 00 PM

Also a small page in the Scenes page to keep it consistent with Land and Avatar pages


nicosantangelo (changes requested):
this was driving me crazy ❤️ 

I think I found a bug on a smaller resolution tho?
image

nicosantangelo (approved):

PR #1149, opened 2020-10-27
feat: open editor

cazala:
Added button to open editor from the dashboard:

Screen Shot 2020-10-27 at 7 47 28 PM

Also added an empty state for the editor and fixed a bunch of small bugs and warnings

### DIFF
--- a/src/components/ItemEditorPage/RightPanel/MultiSelect/MultiSelect.tsx
+++ b/src/components/ItemEditorPage/RightPanel/MultiSelect/MultiSelect.tsx
@@ -50,7 +50,7 @@ export default class MultiSelect<T extends string> extends React.PureComponent<P
         <div className="values">
           {value.length > 0 ? (
             value.map(value => (
-              <div className="value">
+              <div className="value" key={value}>
                 {labels[value]}
                 <div
                   className="remove"
--- a/src/components/ItemEditorPage/RightPanel/Tags/Tags.tsx
+++ b/src/components/ItemEditorPage/RightPanel/Tags/Tags.tsx
@@ -65,7 +65,7 @@ export default class Tags extends React.PureComponent<Props, State> {
       <div className={`Tags ${value.length > 0 ? '' : 'blank'} ${isDisabled ? 'is-disabled' : ''}`.trim()}>
         <div className="values">
           {value.map(value => (
-            <div className="value">
+            <div className="value" key={value}>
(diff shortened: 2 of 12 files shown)


nicosantangelo:
is this empty state for collections correct?

nicosantangelo:
Also the border on empty collections is...weird
image

src/components/ItemEditorPage/RightPanel/MultiSelect/MultiSelect.tsx
```diff
@@ -50,7 +50,7 @@ export default class MultiSelect<T extends string> extends React.PureComponent<P
         <div className="values">
           {value.length > 0 ? (
             value.map(value => (
-              <div className="value">
+              <div className="value" key={value}>
```
nicosantangelo, line 53: Maybe we should be using the `index` here? because values on a multiselect aren't necessarily unique right?

src/components/ItemEditorPage/RightPanel/Tags/Tags.tsx
```diff
@@ -65,7 +65,7 @@ export default class Tags extends React.PureComponent<Props, State> {
       <div className={`Tags ${value.length > 0 ? '' : 'blank'} ${isDisabled ? 'is-disabled' : ''}`.trim()}>
         <div className="values">
           {value.map(value => (
-            <div className="value">
+            <div className="value" key={value}>
```
nicosantangelo, line 68: Are tags unique? if not maybe we should use the `index` here

nicosantangelo (commented):
Two minor comments

src/components/ItemEditorPage/RightPanel/MultiSelect/MultiSelect.tsx
cazala: they are actually :) if you try to write the same tag twice you will notice the second one doesn't get added (like if it was a Set), that's why i used value as a unique id

src/components/ItemEditorPage/RightPanel/Tags/Tags.tsx
cazala: same as above, both tags and multiselect can't have the same value repeated

cazala:
style issues fixed

nicosantangelo (approved):

PR #1155, opened 2020-10-29
feat: ability to change item's description

cazala:
also fixed the metadata issue with bodyshapes + added description to metadata on chain

### DIFF
--- a/src/components/ItemEditorPage/RightPanel/RightPanel.tsx
+++ b/src/components/ItemEditorPage/RightPanel/RightPanel.tsx
@@ -6,7 +6,7 @@ import ItemProvider from 'components/ItemProvider'
 import ConfirmDelete from 'components/ConfirmDelete'
 import { isEqual } from 'lib/address'
 import { getMissingBodyShapeType } from 'modules/item/utils'
-import { Item, ItemRarity, WearableCategory } from 'modules/item/types'
+import { Item, ItemRarity, ITEM_DESCRIPTION_MAX_LENGTH, ITEM_NAME_MAX_LENGTH, WearableCategory } from 'modules/item/types'
 import Collapsable from './Collapsable'
 import Input from './Input'
 import Select from './Select'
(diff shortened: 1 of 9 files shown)


src/components/ItemEditorPage/RightPanel/RightPanel.tsx
```diff
@@ -116,6 +116,13 @@ export default class RightPanel extends React.PureComponent<Props> {
                         disabled={item.isPublished || !isOwner}
                         onChange={name => this.handleChange({ ...item, name })}
                       />
+                      <Input
+                        itemId={item.id}
+                        label={t('global.description')}
+                        value={item.description}
+                        disabled={item.isPublished || !isOwner}
```
nachomazzara, line 123: What do you think of moving `item.isPublished || !isOwner` to a variable like `canEdit`? Seems to be used also in other components.

src/modules/item/utils.ts
```diff
@@ -107,7 +107,9 @@ export function getMetadata(item: Item) {
     case ItemType.WEARABLE: {
       const data = item.data as WearableData
       const bodyShapes = getBodyShapes(item)
-      return `${version}:${type}:${slug}:${data.category}:${bodyShapes.join(',')}`
+      return `${version}:${type}:${item.name}:${item.description}:${data.category}:${bodyShapes
+        .map(bodyShape => bodyShape.split('/').pop())
```
nachomazzara, line 111: I'm wondering if we can add a comment here explaining why we are doing 
```javascript
bodyShape.split('/').pop()
```

Like: 
```bash
// bodyshapes are dcl://BaseFemale and dcl://BaseMale. We store only the name of the shape: BaseFemale or BaseMale
```

src/components/Modals/CreateItemModal/CreateItemModal.tsx
```diff
@@ -144,6 +144,7 @@ export default class CreateItemModal extends React.PureComponent<Props, State> {
         item = {
           id,
           name,
+          description: '',
```
nachomazzara, line 147: Do we need to add the category to this modal?

nachomazzara (changes requested):
Thanks @cazala. Just small comments.

Also, can we limit names and descriptions to [32 max](https://github.com/decentraland/builder/pull/1156#pullrequestreview-520126863)? If you want, it can be done in another PR or here.

src/components/Modals/CreateItemModal/CreateItemModal.tsx
cazala: The category is added in this modal already. It's inside `item.data.category`.

src/modules/item/utils.ts
cazala: done 👍

src/components/ItemEditorPage/RightPanel/RightPanel.tsx
cazala: mm I can't create a `canEdit` variable at the root of the `render` method because it depends on `item` that is taken from a lambda within the return value of `render`. I could create a variable within that scope but it's not going to be at the root level with the rest of the variables which i don't think it makes it much clearer tbh

cazala:
@nachomazzara thanks for the feedback, i added the 32 char limit and addressed your comments

src/components/Modals/CreateItemModal/CreateItemModal.tsx
nachomazzara: I meant `description`. Solved off-line

PR #1156, opened 2020-10-29
feat: edit collection name

cazala:
Closes #1155


nachomazzara (commented):
The PR looks great!
I have one request and one question 😇

Request: Can we limit the collection name to have a maximum length of 32? Would be great if can fit in one word of 32 bytes. What do you think?

Question: Do we need to handle error messages when the name save failed (server failure)? and if you agree with the question above, when the collection is longer than the maximum allowed?

cazala:
@nachomazzara I added a limit of 32 characters for the collection name. Regarding your question, currently we are not showing server errors requests associated with the wearables editor on the UI, but with this limit the user wouldn't be able to send an invalid request because the UI will limit it to a maximum of 32 chars.

nachomazzara:
> @nachomazzara I added a limit of 32 characters for the collection name. Regarding your question, currently we are not showing server errors requests associated with the wearables editor on the UI, but with this limit the user wouldn't be able to send an invalid request because the UI will limit it to a maximum of 32 chars.

Cool! Should we need to add a PR to check this also in the server?

nachomazzara (approved):

cazala:
@nachomazzara i added a PR for that here: https://github.com/decentraland/builder-server/pull/128

PR #1163, opened 2020-10-30
refactor: change the ENS state to support multiple wallets

nicosantangelo:
Needs a bit more testing. Ropsten is not cooperating today


malaniz (approved):
LGTM! thanks for all this refactor and improvements! as we talk, it works superb!

PR #1169, opened 2020-11-03
feat: land analytics

### LINKED ISSUE #979: Fetch analytics data for deployed scenes from frontend
Fetch analytics data for deployed scenes from frontend. Store it in the app state.

### LINKED ISSUE #981: Show analytics data on UI
Show analytics data on UI.

* Land Detail
* Scene Detail

The data to show will be:

- Weekly Active Users
- Weekly Sessions
- Median session time
- Max concurrent users.

Everything for the last 7 days

cazala:
Closes #979 
Closes #981

### DIFF
--- a/src/components/SceneStats/SceneStats.container.ts
+++ b/src/components/SceneStats/SceneStats.container.ts
@@ -1,0 +1,19 @@
+import { connect } from 'react-redux'
+import { RootState } from 'modules/common/types'
+import { fetchSceneStatsRequest, FetchSceneStatsRequestAction, FETCH_SCENE_STATS_REQUEST } from 'modules/stats/action'
+import { getStatsByDeploymentId, getLoading } from 'modules/stats/selectors'
+import { MapStateProps, MapDispatchProps, MapDispatch, OwnProps } from './SceneStats.types'
+import SceneStats from './SceneStats'
+
+const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => ({
+  stats: getStatsByDeploymentId(state)[ownProps.deployment.id],
+  isLoading: getLoading(state)
+    .filter(action => action.type === FETCH_SCENE_STATS_REQUEST)
+    .some(action => (action as FetchSceneStatsRequestAction).payload.base === ownProps.deployment.base)
+})
+
+const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
+  onFetchStats: base => dispatch(fetchSceneStatsRequest(base))
+})
+
+export default connect(mapState, mapDispatch)(SceneStats)
--- a/src/components/SceneStats/SceneStats.container.ts
+++ b/src/components/SceneStats/SceneStats.container.ts
@@ -1,20 +1,21 @@
 import { connect } from 'react-redux'
 import { RootState } from 'modules/common/types'
-import { fetchSceneStatsRequest, FetchSceneStatsRequestAction, FETCH_SCENE_STATS_REQUEST } from 'modules/stats/action'
-import { getStatsByDeploymentId, getLoading } from 'modules/stats/selectors'
+import { fetchWeeklySceneStatsRequest, FetchWeeklySceneStatsRequestAction, FETCH_WEEKLY_SCENE_STATS_REQUEST } from 'modules/stats/action'
+import { getWeeklyStatsByDeploymentId, getLoading } from 'modules/stats/selectors'
 import { MapStateProps, MapDispatchProps, MapDispatch, OwnProps } from './SceneStats.types'
 import SceneStats from './SceneStats'
 
 const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => ({
-  stats: getStatsByDeploymentId(state)[ownProps.deployment.id],
+  stats: getWeeklyStatsByDeploymentId(state)[ownProps.deployment.id],
   isLoading: getLoading(state).some(
     action =>
-      action.type === FETCH_SCENE_STATS_REQUEST && (action as FetchSceneStatsRequestAction).payload.base === ownProps.deployment.base
+      action.type === FETCH_WEEKLY_SCENE_STATS_REQUEST &&
+      (action as FetchWeeklySceneStatsRequestAction).payload.base === ownProps.deployment.base
   )
 })
 
 const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
-  onFetchStats: base => dispatch(fetchSceneStatsRequest(base))
+  onFetchStats: base => dispatch(fetchWeeklySceneStatsRequest(base))
 })
 
 export default connect(mapState, mapDispatch)(SceneStats)
--- a/src/components/SceneStats/SceneStats.container.ts
+++ b/src/components/SceneStats/SceneStats.container.ts
@@ -7,9 +7,10 @@ import SceneStats from './SceneStats'
 
 const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => ({
   stats: getStatsByDeploymentId(state)[ownProps.deployment.id],
-  isLoading: getLoading(state)
-    .filter(action => action.type === FETCH_SCENE_STATS_REQUEST)
-    .some(action => (action as FetchSceneStatsRequestAction).payload.base === ownProps.deployment.base)
+  isLoading: getLoading(state).some(
+    action =>
+      action.type === FETCH_SCENE_STATS_REQUEST && (action as FetchSceneStatsRequestAction).payload.base === ownProps.deployment.base
+  )
 })
 
 const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
--- a/src/components/SceneStats/SceneStats.tsx
+++ b/src/components/SceneStats/SceneStats.tsx
@@ -3,6 +3,7 @@ import { t } from 'decentraland-dapps/dist/modules/translation/utils'
 import { formatTime } from 'lib/date'
 import Stats from './Stats'
 import { Props } from './SceneStats.types'
+import { Loader } from 'decentraland-ui'
 
 export default class SceneStats extends React.PureComponent<Props> {
   componentWillMount() {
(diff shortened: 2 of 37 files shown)


src/modules/deployment/selectors.ts
```diff
@@ -115,3 +115,10 @@ export const getEmptyTiles = createSelector<RootState, Record<string, Deployment
     return result
   }
 )
+
+export const getDeploymentsByBase = createSelector<RootState, DeploymentState['data'], Record<string, Deployment>>(getData, deployments =>
```
nicosantangelo, line 119: I think this is not being used?

src/components/SceneStats/SceneStats.container.ts
```diff
@@ -0,0 +1,19 @@
...
+import { fetchSceneStatsRequest, FetchSceneStatsRequestAction, FETCH_SCENE_STATS_REQUEST } from 'modules/stats/action'
+import { getStatsByDeploymentId, getLoading } from 'modules/stats/selectors'
+import { MapStateProps, MapDispatchProps, MapDispatch, OwnProps } from './SceneStats.types'
+import SceneStats from './SceneStats'
+
+const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => ({
+  stats: getStatsByDeploymentId(state)[ownProps.deployment.id],
+  isLoading: getLoading(state)
+    .filter(action => action.type === FETCH_SCENE_STATS_REQUEST)
+    .some(action => (action as FetchSceneStatsRequestAction).payload.base === ownProps.deployment.base)
```
nicosantangelo, line 12: Maybe this can be a selector on its own? if not, you can just use some for the entire condition instead of looping twice right?

src/modules/stats/action.ts
```diff
@@ -0,0 +1,16 @@
+// Fetch Stats
```
nicosantangelo, line 1: I think this comment is meant to be below the imports right?

src/modules/stats/reducer.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    case FETCH_SCENE_STATS_SUCCESS: {
+      const { base, stats } = action.payload
+      return {
+        ...state,
+        loading: loadingReducer(state.loading, action),
+        error: null,
+        data: {
+          ...state.data,
+          [base]: stats
```
nicosantangelo, line 43: This is a new object, but shouldn't we copy it `{ ...stats }` anyways?

src/components/SceneStats/SceneStats.tsx
```diff
@@ -0,0 +1,44 @@
...
+          {stats => (stats ? stats.users.toLocaleString() : '0')}
+        </Stats>
+        <Stats label={t('analytics.sessions')} stats={stats} isLoading={isLoading}>
+          {stats => (stats ? stats.sessions.toLocaleString() : '0')}
+        </Stats>
+        <Stats label={t('analytics.median_session_time')} stats={stats} isLoading={isLoading}>
+          {stats => (stats ? formatTime(stats.medianSessionTime) : '0s')}
+        </Stats>
+        <Stats label={t('analytics.max_concurrent_users')} stats={stats} isLoading={isLoading}>
+          {stats => (stats ? stats.maxConcurrentUsers.toLocaleString() : '0')}
```
nicosantangelo, line 39: why are these functions? I looked at Stats.tsx and it looks like 
`isLoading && !stats ? <Loader active size="mini" /> : children(stats)`

so isn't it better to just move that condition to each of these? and have Stats just be a component that returns 
```tsx
      <div className="Stats">
        <div className="value">{value}</div>
        <div className="label">{label}</div>
      </div>
```
Also, these are not the same Stats as decentraland-uis right?

src/components/SceneStats/SceneStats.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const { stats, isLoading, onFetchStats } = this.props
+    if (!stats && !isLoading) {
+      onFetchStats(base)
+    }
+  }
+
+  render() {
+    const { stats, isLoading } = this.props
+    return (
+      <>
```
nicosantangelo, line 28: This is not wrong but I'm a bit hesitant to return a component that spits out unwrapped elements cause we can't scope them (to style it for example)

You think that might be a problem at some point?

src/modules/common/reducer.ts
```diff
@@ -57,6 +58,7 @@ export function createRootReducer(history: History) {
       item,
       collection,
       location,
+      stats,
```
nicosantangelo, line 61: Couldn't `stats` be a bit too generic? or are we planning to re-use it if we ever need to display stats elsewhere?

Also! maybe it's a bit weird that the types and part of the code call the Stats `WeeklyStats` but the store/variable names don't reflect that

nicosantangelo (commented):
it works great 🔥

src/components/SceneStats/SceneStats.tsx
cazala: I'm formatting `value` different on some of those stats, that's why I assed it as a function. For instance for the stat `users: 1024` we show `"1,024"` but for `medianSessionTime: 3600` we show `"1h"`. Also when the stat is `null` the default value for `users` is `"0"` while for `medianSessionTime` is `"0s"`.

src/components/SceneStats/SceneStats.tsx
cazala: mm since these are all `Stats`, if we wanted to style them from the inside we can just use `Stats.css`. If we wanted to style them from the outside, the parent can still wrap it and style it. But if I wrap it here, it's impossible to use it unwrapped from the outside, for instance to add these as siblings to other components and align them using flex, which is what I'm doing on the parent components. That's why I left it unwrapped.

src/modules/common/reducer.ts
cazala: Maybe, what name would you recommend? I didn't use analytics because that's an already existing module for our analytics.

Regarding thhe second point, there was a discussion to add daily/monthly analytics too, and being able to switch from a dropdown, but it required some work on the redshift side. That's why i named the endpoint and the data type "weekly", I can wrap these stats under `data.weekly` instead of just `data`, would that make it clearer?

src/modules/deployment/selectors.ts
cazala: good catch! i ll remove it

src/modules/stats/reducer.ts
cazala: mm i don't think we do that anywhere tbh, i just checked a few reducers. but i don't see why would you clone it, if we are not mutating?

src/components/SceneStats/SceneStats.container.ts
cazala: done 👍

src/components/SceneStats/SceneStats.tsx
cazala: ah you mean moving the Loader outside? But then instead of having stuff like 

```
{stats => (stats ? stats.users.toLocaleString() : '0')}
```

I ll have 

```
{isLoading && !stats ? <Loader active size="mini" /> : stats ? ? stats.users.toLocaleString() : '0')}
```

repeated 4 times

src/modules/common/reducer.ts
nicosantangelo: > Maybe, what name would you recommend? I didn't use analytics because that's an already existing module for our analytics.

I'm not sureeeee. Will ping you on Slack about it 🚀

> Regarding thhe second point, there was a discussion to add daily/monthly analytics too, and being able to switch from a dropdown, but it required some work on the redshift side. That's why i named the endpoint and the data type "weekly", I can wrap these stats under data.weekly instead of just data, would that make it clearer?

Oh! maybe it's not necessary then, we can do that when we get there. If you think it makes it a bit clearer in case someone else has to develop that feature go right ahead! we can also leave a comment

src/components/SceneStats/SceneStats.tsx
nicosantangelo: True, I guess you can just wrap them on the parent. What I was thinking is that if you do `.Parent .Stats` that could mean these or decentraland-ui's for example.
But not a problem really, just wanted to point that out just in case

src/modules/stats/reducer.ts
nicosantangelo: I just checked and I think we are doing it in some cases like asset packs (LOAD_ASSET_PACKS_SUCCESS or LOAD_COLLECTIBLES_SUCCESS) right?

My thought process was that you don't know where else that object is going to go via de action and if you don't clone it here, another part of the store can save it and share a reference

src/components/SceneStats/SceneStats.tsx
nicosantangelo: True, but thing is that when I read the code I thought each Stat was fetching the object (or at least trying to), because that's the pattern we usually use. It's weird to see a function for an object you yourself supplied.

That said, it's not really _a problem_ it's A-OK the way it is

src/modules/stats/reducer.ts
cazala: i checked and yes, there are a few reducers where we do this, but tbh i don't think we get anything out of breaking a reference if the data hasn't changed. We want to avoid at all costs mutating because that's what breaks pure components and selectors. But if the data is not changing, I don't see why not keep that reference.

For instance in assetPacks we add an `assets` prop to each assetPacks, thus we need to clone it. On collections we omit the `items` prop. But on most of the modules we just assign the instances to the reducer `data`.

src/components/SceneStats/SceneStats.tsx
cazala: yes, it's true that passing `stats` to the children lambda makes it seem like it's fetching/providing it, and you can still use it even if not provided in that function, is the same instance that's already present in the scope.. maybe I can make a helper class method of the parent instead of a behavior of Stats

cazala: done 👍

src/modules/common/reducer.ts
cazala: i added "weekly" to the actions names and within the reducer data

nicosantangelo (approved):
LGTM

Issue #1136, opened 2020-10-26
List names


nachomazzara:
It shouldn't add any functionality other than showing the action links.

* Name
  * Show person icon near to the already avatar's name in use.
* Being Assigned
* Assigned to
* Actions:
  - Assign/Edit
  - Use as alias

nicosantangelo:
# How to go about developing this issue

The general action points for this issue are:
1. Add a new page and tab for `Names`
2. Grab the ENSs of the connected wallet and show them in the new table. Remember that this module exists on the state right now, called `ENS` and lives on `moules/ens`. For example the selector for the ENSs wallet is `getENSByWallet`
3. Update `handleFetchDomainListRequest` found on `modules/ens/sagas` which is the method that handles the request for the user's ENSs. Right now this method works as intended BUT it cant determine from which land each ENS is. For each ENSs fetched, we need to update the `landId`
4. Add pagination. For this you'll need to use the `ui` module, which is already being used for things like the dashboard.

That's about it for this PR, the rest of the UI should be drawn but not implemented (like `use as alias`)

Issue #1134, opened 2020-10-26
Add "+" action to the name list


nachomazzara:
+ refers to the top right "add" action

PR #1174, opened 2020-11-09
feat: ENS List

malaniz:
Include changes for the following issues.

https://github.com/decentraland/builder/issues/1136
https://github.com/decentraland/builder/issues/1135
https://github.com/decentraland/builder/issues/1134
https://github.com/decentraland/builder/issues/1133


nicosantangelo (changes requested):
**General things**
- I think we're missing a spinner, I see 0 results for quite some time before seeing my ens list
- I see the assign action a tad to the left
image
- I think there's a race condition because sometimes, when I try to load my name I run into this error:
image

and the page is stuck there

nicosantangelo (approved):

PR #1180, opened 2020-11-12
feat: re assign ens

### LINKED ISSUE #1141: Re-assign Name
- Show Map to select the name parcel
- After selecting the new parcel and clicking "continue", the user should be redirected to the Edit Name page with the parcel and name selected ready to send the `Set content" transaction

### LINKED ISSUE #1170: Refactor: Consistent naming for components
We have some name inconsistencies around ENS, mostly regarding Components. Some examples are:

```
- LandSetNameForm.tsx
- LandSelectNameForm.tsx
- LandEnsPage.tsx
```

I think it's a good idea to keep `ENS` as the name for the code even if we then use `Name` for things like the tab on the UI.

Also for variable names I'd keep a capitalized `ENS` except for a single one. So:

```ts
const ens = { ... }
function getENS() {}
type ENS = { ... }
const ensList = [ ... ]
```

And
```
- LandSetNameForm.tsx    =>  SetENSDataForm.tsx  // Sets the resolver and content
- LandSelectNameForm.tsx =>  SelectENSForm.tsx   // Selects a ENS from the list to set on a Land
- LandEnsPage.tsx        =>  SetENSPage.tsx      // Contains the other two
```

What do you think @cazala @malaniz
I don't love the names I came up with for the files even tho I think they're better than what we have now no?
Also, is there any other name I'm missing? ( @malaniz )

nicosantangelo:
Closes #1141 
Closes #1170


cazala:
A few thingys:

1) The navigation in the `/set-name` page is showing the tabs and not showing the back button, and the map is missing the rounded corners:

Screen Shot 2020-11-16 at 7 39 50 PM

Designs:

Screen Shot 2020-11-16 at 8 14 49 PM

2) When I click on the Wearables tab, the app blows up, and the router complains with: 

```
Warning: <Route> elements should not change from uncontrolled to controlled (or vice versa). You initially used no "location" prop and then provided one on a subsequent render.
```

Seems like a rebase issue?

Other than these minor issues the functionality of the feature is working as expected 👍

malaniz (changes requested):
There is only one detail that I found. When you want to assign a land that is currently used for that domain it goes to the `SetResolver / SetContent` page with only available "Finish Button".
img-bUK9Bs_000

I think it is a good idea to check and show a message as we made in the /land/<coord>/set-name and disabling the `continue` button.

malaniz (approved):
LGTM!

cazala (approved):

Issue #1175, opened 2020-11-11
Use as Alias


nachomazzara:
Use as alias is an action where a user decides to use a name on his avatar. 

The action should send a profile to the content-server with the current metadata (avatar), but changing **ONLY** its name.

A confirm popup should be great explaining what is going to happen.

This repo has how to replace a profile's name: https://github.com/nchamo/dcl-name-swap/blob/main/src/main.ts#L72

Working app: https://dcl-name-swap.surge.sh/

PR #1183, opened 2020-11-17
feat: Use as Alias

malaniz:
WIP.

https://github.com/decentraland/builder/issues/1175


src/components/Modals/UseAsAliasModal/UseAsAliasModal.tsx
```diff
@@ -0,0 +1,45 @@
...
+  render() {
+    const { onClose, isLoading } = this.props
+    const { name, oldname } = this.props.metadata
+    const { done } = this.state
+    return (
+      <Modal name={name} onClose={onClose} size="tiny">
+        <ModalNavigation title={t('use_as_alias_modal.title')} subtitle={t('use_as_alias_modal.subtitle')} onClose={onClose} />
+        <Form onSubmit={this.handleSubmit}>
+          <ModalContent>
+            <p> {t('use_as_alias_modal.body', { name, oldname })} </p>
```
nicosantangelo, line 28: Makes little difference but we can update this to 
```suggestion
            <p>{t('use_as_alias_modal.body', { name, oldname })}</p>
```

src/components/Modals/UseAsAliasModal/UseAsAliasModal.container.ts
```diff
@@ -0,0 +1,19 @@
...
+import { isLoadingType } from 'decentraland-dapps/dist/modules/loading/selectors'
+import { RootState } from 'modules/common/types'
+import { getLoading } from 'modules/ens/selectors'
+import { MapStateProps, MapDispatchProps, MapDispatch } from './UseAsAliasModal.types'
+import UseAsAliasModal from './UseAsAliasModal'
+import { changeProfileRequest, CHANGE_PROFILE_REQUEST } from 'modules/ens/actions'
+
+const mapState = (state: RootState): MapStateProps => ({
+  address: getAddress(state),
+  isLoading: isLoadingType(getLoading(state), CHANGE_PROFILE_REQUEST)
```
nicosantangelo, line 12: Given that you're already connecting the Component to the store for isLoading (which is totally correct), what do you think of just getting the current name here? (`yield getName(state)`)
I think it's a bit cleaner to be able to call this modal like this:
`onOpenModal('UseAsAliasModal', { name })`
and also rids the Component that's calling it from that dependency

What do you think?

src/modules/ens/sagas.ts
```diff
@@ -207,3 +217,43 @@ function* handleFetchENSListRequest(_action: FetchENSListRequestAction) {
...
+      const authChain = Authenticator.createSimpleAuthChain(deployPreparationData.entityId, address, signature)
+      yield client.deployEntity({ ...deployPreparationData, authChain })
+
+      yield put(changeProfileSuccess())
+    }
+  } catch (error) {
+    const ensError: ENSError = { message: error.message }
+    yield put(changeProfileFailure(ensError))
+  }
+}
```
nicosantangelo, line 259: As a general comment we talked about in another PR. `yield`ing promises technically works, but the preferred method is to use `call`, the helper provided by redux-sagas. So code could be rewritten like this:
```diff
- yield client.deployEntity({ ...deployPreparationData, authChain })
+ yield call(() => client.deployEntity({ ...deployPreparationData, authChain }))
```

check other sites of the same occurence in this file like `yield call(() => ensContract.methods.resolver(nodehash).call())`

nicosantangelo (changes requested):
Sorry if the review is a bit all over the place. I struggled to tell where the best place for some of the comments was.

src/modules/ens/sagas.ts
```diff
@@ -207,3 +217,43 @@ function* handleFetchENSListRequest(_action: FetchENSListRequestAction) {
...
   }
 }
+
+function* handleChangeProfile(action: ChangeProfileRequestAction) {
+  try {
+    const { selectedName } = action.payload
+
+    const address = yield select(getAddress)
+
+    const client = new CatalystClient('https://peer.decentraland.org', 'builder')
```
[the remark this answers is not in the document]
cazala: yep, this is saving the entity on the production catalyst, probably the reason you didn't see the changes reflected on your profile after set the alias

src/modules/ens/actions.ts
```diff
@@ -65,3 +65,16 @@ export const fetchENSListFailure = (error: ENSError) => action(FETCH_ENS_LIST_FA
...
+export const CHANGE_PROFILE_SUCCESS = '[Success] Change Profile'
+export const CHANGE_PROFILE_FAILURE = '[Failure] Change Profile'
+
+export const changeProfileRequest = (selectedName: string) => action(CHANGE_PROFILE_REQUEST, { selectedName })
+export const changeProfileSuccess = () => action(CHANGE_PROFILE_SUCCESS, {})
+export const changeProfileFailure = (error: ENSError) => action(CHANGE_PROFILE_FAILURE, { error })
+
+export type ChangeProfileRequestAction = ReturnType<typeof changeProfileRequest>
+export type ChangeProfileSuccessAction = ReturnType<typeof changeProfileSuccess>
+export type ChangeProfileFailureAction = ReturnType<typeof changeProfileFailure>
```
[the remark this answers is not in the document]
nachomazzara: Names are associated with a profile/avatar, so yes! `Profile` is the right wording (and maybe place)

src/modules/ens/sagas.ts
```diff
@@ -207,3 +217,43 @@ function* handleFetchENSListRequest(_action: FetchENSListRequestAction) {
...
+    const eth = Eth.fromCurrentProvider()
+    if (eth) {
+      const personal = new Personal(eth.provider)
+      const signature = yield personal.sign(deployPreparationData.entityId, address, '')
+
+      // Deploy change
+      const authChain = Authenticator.createSimpleAuthChain(deployPreparationData.entityId, address, signature)
+      yield client.deployEntity({ ...deployPreparationData, authChain })
+
+      yield put(changeProfileSuccess())
```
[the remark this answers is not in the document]
cazala: I was going to comment something like this, I would recommend having this saga and action in the `ens` modules, maybe something like `SET_ALIAS_REQUEST/SUCCESS/FAILURE` and within the saga generate a new `Profile` and at the end of the saga dispatch an `UPDATE_PROFILE` action with the new `Profile`, and that action should live in the `profile` module. I'm saying this because the profile module is likely to be moved to `decentraland-dapps` to be used in the marketplace and other dapps, and we'll want the ability to update profiles (thus `UPDATE_PROFILE` should live there) but setting the alias is something related to ENS and problably shouldn't be mixed with the more general profile module.

cazala (changes requested):

src/modules/translation/languages/en.json
```diff
@@ -291,6 +291,12 @@
     },
     "title": "Publish your Scene"
   },
+  "use_as_alias_modal": {
```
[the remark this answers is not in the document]
nachomazzara: 🇨🇳 🇪🇸

src/modules/ens/actions.ts
```diff
@@ -65,3 +65,16 @@ export const fetchENSListFailure = (error: ENSError) => action(FETCH_ENS_LIST_FA
 export type FetchENSListRequestAction = ReturnType<typeof fetchENSListRequest>
 export type FetchENSListSuccessAction = ReturnType<typeof fetchENSListSuccess>
 export type FetchENSListFailureAction = ReturnType<typeof fetchENSListFailure>
+
+// Set Alias
+export const SET_ALIAS_REQUEST = '[Request] Change Profile'
```
cazala, line 70: Watch out these strings still say `Change Profile`

src/modules/ens/sagas.ts
```diff
@@ -27,13 +27,26 @@ import {
...
+  setAliasFailure
 } from './actions'
 import { ENS, ENSOrigin, ENSError } from './types'
 import { getLands } from 'modules/land/selectors'
 import { FETCH_LANDS_SUCCESS } from 'modules/land/actions'
+import { CatalystClient, DeploymentBuilder, DeploymentWithMetadataContentAndPointers } from 'dcl-catalyst-client'
+import { EntityType } from 'dcl-catalyst-commons'
+import { env } from 'decentraland-commons'
+import { Avatar } from 'decentraland-ui'
+import { Personal } from 'web3x-es/personal'
```
cazala, line 43: These non-relative imports should be on top :) 🔝

cazala:
😅

nicosantangelo:
I hope nico makes an appareance too

cazala:
Are they our doppelgangers...? or are WE the doppelgangers?? 😱 😱 😱

src/components/Modals/UseAsAliasModal/UseAsAliasModal.types.ts
```diff
@@ -0,0 +1,16 @@
+import { Dispatch } from 'redux'
+import { ModalProps } from 'decentraland-dapps/dist/providers/ModalProvider/ModalProvider.types'
+
+export type Props = ModalProps & {
+  address?: string
```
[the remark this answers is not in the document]
malaniz: Thanks @nicosantangelo! it is necessary now. It is used in 
```javascrtipt 
  handleSubmit = () => {
    const { onSubmit, address, metadata } = this.props
    if (address) {
      onSubmit(address, metadata.name)
    }
  }

```

src/components/Modals/UseAsAliasModal/UseAsAliasModal.tsx
```diff
@@ -0,0 +1,45 @@
...
+  handleSubmit = () => {
+    if (!this.state.done) {
+      this.setState({ done: true })
+      this.props.onSubmit(this.props.metadata.name)
+    }
+  }
+
+  render() {
+    const { onClose, isLoading } = this.props
+    const { name, oldname } = this.props.metadata
```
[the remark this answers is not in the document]
malaniz: Thanks @nicosantangelo. I've added types for metadata and changed `oldname` to `oldName`.

src/components/Modals/UseAsAliasModal/UseAsAliasModal.tsx
```diff
@@ -0,0 +1,45 @@
...
+    const { done } = this.state
+    return (
+      <Modal name={name} onClose={onClose} size="tiny">
+        <ModalNavigation title={t('use_as_alias_modal.title')} subtitle={t('use_as_alias_modal.subtitle')} onClose={onClose} />
+        <Form onSubmit={this.handleSubmit}>
```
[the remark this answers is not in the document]
malaniz: Thanks @nicosantangelo. Fixed! 100% agree with you.

src/modules/ens/actions.ts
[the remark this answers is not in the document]
malaniz: Done! and thanks!!!

src/components/ENSListPage/ENSListPage.tsx
```diff
@@ -167,7 +167,14 @@ export default class ENSListPage extends React.PureComponent<Props, State> {
...
                             {!this.isAlias(ens) ? (
                               <Column align="right">
-                                <Button className="ui basic button" onClick={() => alert('must be implemented')}>
+                                <Button
+                                  className="ui basic button"
+                                  onClick={() => {
+                                    const name = ens.subdomain.split('.')[0]
+                                    const oldname = alias
+                                    onOpenModal('UseAsAliasModal', { name, oldname })
+                                  }}
```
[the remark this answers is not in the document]
malaniz: Done and Thanks! I've coded 2 helpers in utils to make this. `getDomainFromName` and `getNameFromDomain`.

src/components/Modals/UseAsAliasModal/UseAsAliasModal.css
```diff
@@ -0,0 +1,3 @@
+.UseAsAliasModal .field .message {
+  display: none;
+}
```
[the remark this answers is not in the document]
malaniz: Done! thanks!

src/components/Modals/UseAsAliasModal/UseAsAliasModal.tsx
malaniz: Fixed!

src/modules/ens/actions.ts
malaniz: Fixed! thanks @cazala

src/modules/ens/sagas.ts
malaniz: Fixed! thanks @cazala

src/modules/ens/sagas.ts
malaniz: Done and Fixed!

src/components/Modals/UseAsAliasModal/UseAsAliasModal.container.ts
```diff
@@ -0,0 +1,20 @@
+import { connect } from 'react-redux'
+import { isLoadingType } from 'decentraland-dapps/dist/modules/loading/selectors'
+import { RootState } from 'modules/common/types'
+import { getENSUsedAsAlias, getLoading } from 'modules/ens/selectors'
+import { MapStateProps, MapDispatchProps, MapDispatch } from './UseAsAliasModal.types'
+import UseAsAliasModal from './UseAsAliasModal'
+import { setAliasRequest, SET_ALIAS_REQUEST } from 'modules/ens/actions'
+import { getAddress } from 'decentraland-dapps/dist/modules/wallet/selectors'
```
nicosantangelo, line 8: Absolute imports to the top! remember the [standards](https://github.com/decentraland/standards/blob/master/standards/react-redux.md#5-imports-order-on-files)

src/components/Modals/UseAsAliasModal/UseAsAliasModal.types.ts
```diff
@@ -0,0 +1,20 @@
...
+  isLoading: boolean
+  usedAsAlias: ENS[]
+  metadata: {
+    oldName: string
+    name: string
+  }
+  onSubmit: (address: string, name: string) => void
```
nicosantangelo, line 13: This is fine but we try to type the methods we use with the "actual" actions they're calling. Like we did on `ENSListPage` !

```ts
// ENSListPage/ENSListPage.types.ts
import { openModal } from 'modules/modal/actions'

export type Props = {
  // (...)
  onOpenModal: typeof openModal
}
```

src/components/Modals/UseAsAliasModal/UseAsAliasModal.types.ts
```diff
@@ -0,0 +1,20 @@
+import { Dispatch } from 'redux'
+import { ModalProps } from 'decentraland-dapps/dist/providers/ModalProvider/ModalProvider.types'
+import { ENS } from 'modules/ens/types'
+
+export type Props = ModalProps & {
+  address?: string
+  isLoading: boolean
+  usedAsAlias: ENS[]
```
nicosantangelo, line 8: This prop is here to represent the aliases a user has? if so, 
- is it posible for a user to have more than one alias?
- If not, what do you think about renaming this to `alias?: ENS`? if they can, maybe this could be re-named `aliases`? to be more succinct. What do you think?

src/modules/ens/reducer.ts
```diff
@@ -146,6 +158,22 @@ export function ensReducer(state: ENSState = INITIAL_STATE, action: ENSReducerAc
...
+      const domainName = getNameFromDomain(name)
+
+      return {
+        ...state,
+        loading: loadingReducer(state.loading, action),
+        data: {
+          ...state.data,
+          [domainName]: {
+            ...state.data[domainName],
+            isUsedAsAlias: true
```
nicosantangelo, line 172: Careful here! this prop is not necessary. Remember that each prop we put on the state is another property we have to maintain. Same way we'd treat a (relational) database, we try not to have data that can be inferred from somewhere else.
We've talked about this here: https://github.com/decentraland/builder/pull/1174/commits/da945cb10eb3814db3d8dc6a87b808ea68c47a42#r522255544 and we ended up with the `isAlias` method at the time, and I think that solution still applies. Furthermore, imagine if the user added an alias twice, this prop would be true in two entities but it should only be true for one of those

src/modules/ens/sagas.ts
```diff
@@ -44,6 +58,57 @@ function* handleConnectWallet() {
   yield put(fetchENSListRequest())
 }
 
+function* handleSetAlias(action: SetAliasRequestAction) {
+  const { address, name } = action.payload
+  try {
+    const client = new CatalystClient(env.get('REACT_APP_PEER_URL'), 'builder')
```
nicosantangelo, line 64: The PEER_URL can be imported from:
`import { PEER_URL } from 'lib/api/peer'`

src/modules/ens/selectors.ts
```diff
@@ -20,6 +20,12 @@ export const getENSByWallet = createSelector<RootState, ENS[], string | undefine
   ensList.filter(ens => isEqual(ens.address, address))
 )
 
+export const getENSUsedAsAlias = createSelector<RootState, ENS[], string | undefined, ENS[]>(
```
nicosantangelo, line 23: This naming is not necessarily bad but what do you think about being a bit more specific and calling it `getAlias(es)` ? Also echoing the comment above, is it posible for a user to have more than one alias? if not this could return a single `ENS`

By the way, I know the `isAlias` we've added on a previous PR lives on `ENSListPage`, but we can always move it to `ens/utils`

src/modules/profile/actions.ts
```diff
@@ -14,3 +15,13 @@ export const loadProfileFailure = (address: string, error: string) => action(LOA
 export type LoadProfileRequestAction = ReturnType<typeof loadProfileRequest>
 export type LoadProfileSuccessAction = ReturnType<typeof loadProfileSuccess>
 export type LoadProfileFailureAction = ReturnType<typeof loadProfileFailure>
+
+// Change Profile
+export const CHANGE_PROFILE = 'Change Profile'
+
+export const changeProfile = (
+  address: string,
+  profile: Partial<Profile> & { avatars: Partial<Avatar>[]; content?: any[]; metadata?: any[] }
```
nicosantangelo, line 24: Maybe we can name this type and add it to `/types.ts`.

@cazala  Does this exist elsewhere?

nicosantangelo (changes requested):

src/components/ENSListPage/ENSListPage.tsx
```diff
@@ -167,7 +167,13 @@ export default class ENSListPage extends React.PureComponent<Props, State> {
                             ) : null}
                             {!this.isAlias(ens) ? (
                               <Column align="right">
-                                <Button className="ui basic button" onClick={() => alert('must be implemented')}>
+                                <Button
+                                  className="ui basic button"
+                                  onClick={() => {
+                                    const name = getNameFromDomain(ens.subdomain)
+                                    onOpenModal('UseAsAliasModal', { name, oldName: alias })
+                                  }}
```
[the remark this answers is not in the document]
malaniz: I didn't change this to explicitly says that the modal will use it. I don't know if this is a good choice! Probably not! But I want to double check if there is a standard process to manage this type of parameters for modals.

src/components/Modals/UseAsAliasModal/UseAsAliasModal.container.ts
malaniz: btw, Did your linter fixes take care of this type of errors? not mine.

src/components/Modals/UseAsAliasModal/UseAsAliasModal.types.ts
malaniz: Great I'll make this from now on ...

src/components/Modals/UseAsAliasModal/UseAsAliasModal.types.ts
malaniz: Agree! I understand that the user in the interace can't have more than one avatar, but our code is prepared to have more than one. So, I'll prefer to follow with the Multiple Avatars option (then, multiple aliases) and preselet the [0] of the array as we made before. What do you think? In the mean team I'll re-name to `aliases`

src/modules/ens/selectors.ts
malaniz: Renamed to getAliases, based in my comment to the comment above! thanks!

src/modules/ens/reducer.ts
malaniz: Fixed! thanks!

src/modules/profile/actions.ts
cazala: this should be just `(address: string, profile: Profile)` no need to send the whole `entity` or a partial profile

malaniz: you're right! after a too many comings and goings with the Partial<Profile> is not needed anymore. Fixed!

nicosantangelo (approved):

PR #1184, opened 2020-11-18
refactor: remove nested collections and items

cazala:
**Depends on:** https://github.com/decentraland/builder-server/pull/135

This PR makes the necessary changes to use the version of the server that does't return nested items and collections.

Since collections don't come with their items inside, I added an extra endpoint to the server and a corresponding action on the frontend to fetch a collection's items

### DIFF
--- a/src/components/CollectionProvider/CollectionProvider.tsx
+++ b/src/components/CollectionProvider/CollectionProvider.tsx
@@ -3,23 +3,17 @@ import { Props } from './CollectionProvider.types'
 
 export default class CollectionProvider extends React.PureComponent<Props> {
   componentDidMount() {
-    const { id, collection } = this.props
+    const { id, collection, onFetchCollection } = this.props
     if (id && !collection) {
-      this.fetchCollection(id)
+      onFetchCollection(id)
     }
   }
 
   componentDidUpdate(prevProps: Props) {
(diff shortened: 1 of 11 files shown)


src/components/CollectionProvider/CollectionProvider.tsx
```diff
@@ -3,19 +3,25 @@ import { Props } from './CollectionProvider.types'
...
+    const { id, collection } = this.props
+    if (id && id !== prevProps.id && !collection) {
+      this.fetchCollection(id)
     }
   }
 
+  fetchCollection(id: string) {
+    const { onFetchCollection, onFetchCollectionItems } = this.props
+    onFetchCollection(id)
+    onFetchCollectionItems(id)
```
nicosantangelo, line 22: Would it make sense to just add this to the items saga? I think this is better because it's not coupled but it's worth mentioning, if we always want both together we can forego of the double action here

nicosantangelo (commented):
Looks great, testing but left a tiny winy comment

src/components/CollectionProvider/CollectionProvider.tsx
cazala: sure, i moved it to the sagas

nicosantangelo (approved):

PR #1185, opened 2020-11-23
feat: keep the cancel-submit order for forms

### LINKED ISSUE #1153: Action buttons order
The action buttons should always be in the following order:
`CANCEL   DONE`
There are some pages like LAND edit or transfer page that have it the other way around

nicosantangelo:
Closes #1153


nachomazzara (approved):
LGTM

PR #1186, opened 2020-11-27
feat: Claim approve functionality

malaniz:
WIP


src/components/ClaimENSPage/ClaimENSPage.tsx
```diff
@@ -1,16 +1,62 @@
...
+    const { tx } = this.state
+    if (tx && tx.status) {
+      this.setState({ isLoading: false, tx: undefined })
+    }
+  }
+
+  handleManaApprove = async () => {
+    const { address } = this.props
+    const eth: Eth | null = await createEth()
+    if (!eth) return
```
nicosantangelo, line 46: This will silently fail...when would this happen? should we handle it?
@cazala

src/components/ClaimENSPage/ClaimENSPage.tsx
```diff
@@ -1,16 +1,62 @@
...
+    const contractMANA = new MANAToken(eth, Address.fromString(MANA_ADDRESS))
```
nicosantangelo, line 47: I feel that we can extract this maybe? have a method to get the MANA contract and throw if eth is not available, so we can then `try catch` it

What do you think? might be overkill!

src/components/ClaimENSPage/ClaimENSPage.tsx
```diff
@@ -1,16 +1,62 @@
...
+
+    let tx
+    if (this.isManaApproved()) {
+      // > 100
+      tx = contractMANA.methods.approve(Address.fromString(CONTROLER_ADDRESS), 0)
+    } else {
+      tx = contractMANA.methods.approve(Address.fromString(CONTROLER_ADDRESS), 101)
```
nicosantangelo, line 54: Couldn't we just extract this to decide upon the amount to approve instead of the entire method? it's the only thing that changes right?
It seems like tx is not really necessary as an extra variable and can replace `txSent` below

src/components/ClaimENSPage/ClaimENSPage.tsx
```diff
@@ -21,12 +67,17 @@ export default class ClaimENSPage extends React.PureComponent<Props, State> {
     this.setState({ name: data.value })
   }
 
-  handleOnAction = () => {}
+  handleOnAction = () => ({})
```
nicosantangelo, line 70: You can send `undefined` to `onAction` and remove this :)
image

src/components/ClaimENSPage/ClaimENSPage.tsx
```diff
@@ -53,20 +104,25 @@ export default class ClaimENSPage extends React.PureComponent<Props, State> {
...
+                </Section>
+                <Section>
+                  <Radio toggle checked={this.isManaApproved()} onChange={this.handleManaApprove} label={'MANA approved'} />
+                </Section>
                 <Row className="actions">
                   <Button className="cancel" onClick={onBack}>
                     {t('global.cancel')}
                   </Button>
-                  <Button type="submit" primary disabled={!isValid}>
+                  <Button type="submit" primary disabled={!isValid || !this.isManaApproved()} loading={this.state.isLoading}>
```
nicosantangelo, line 125: You're already extracting the name from the state, you can add isLoading there right? (`L83`)

src/components/ClaimENSPage/ClaimENSPage.tsx
```diff
@@ -1,16 +1,62 @@
...
+    const eth: Eth | null = await createEth()
+    const { address } = this.props
+
+    if (eth && address) {
+      const contractMANA = new MANAToken(eth, Address.fromString(MANA_ADDRESS))
+      const allowance: string = await contractMANA.methods
+        .allowance(Address.fromString(address), Address.fromString(CONTROLER_ADDRESS))
+        .call()
+      const amountApproved: number = +allowance
+      this.setState({ amountApproved })
```
nicosantangelo, line 35: You can do this in one line if you want
```suggestion
      this.setState({ amountApproved: +allowance })
```
not trying to say this is strictly worse, but we're saving an unused variable

src/components/ClaimENSPage/ClaimENSPage.tsx
```diff
@@ -1,16 +1,62 @@
...
-    name: ''
+    name: '',
+    amountApproved: 0,
+    isLoading: false,
+    tx: undefined
+  }
+
+  async componentDidUpdate() {
+    const eth: Eth | null = await createEth()
+    const { address } = this.props
```
nicosantangelo, line 27: Sorry for the OCD but we usually keep `props` destructuring before actual methods (like L44 below). Do you think we can change it here?
Same thing for the `const { tx } = this.state` on this same file.

Again, this is not wrong in any way, is to preserve a bit of consistency

src/components/ClaimENSPage/ClaimENSPage.tsx
```diff
@@ -1,16 +1,62 @@
...
-import { Row, Column, Section, Narrow, InputOnChangeData, Header, Form, Field, Button, Mana } from 'decentraland-ui'
+import { Address } from 'web3x-es/address'
+import { Row, Column, Section, Narrow, InputOnChangeData, Header, Form, Field, Button, Mana, Radio } from 'decentraland-ui'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
 import { locations } from 'routing/locations'
 import Back from 'components/Back'
 import LoggedInDetailPage from 'components/LoggedInDetailPage'
 import { MAX_NAME_SIZE, isNameValid, PRICE } from 'modules/ens/utils'
 import { Props, State } from './ClaimENSPage.types'
+import { createEth } from 'decentraland-dapps/dist/lib/eth'
```
nicosantangelo, line 10: 🔝

src/components/ClaimENSPage/ClaimENSPage.tsx
```diff
@@ -1,16 +1,62 @@
...
+
 import './ClaimENSPage.css'
+import { ERC20 as MANAToken } from 'contracts/ERC20'
+import { CONTROLER_ADDRESS, MANA_ADDRESS } from 'modules/common/contracts'
```
nicosantangelo, line 14: Could we move this above `import { Props, State } from './ClaimENSPage.types'` and remove the newline there?  to keep consistent with (most :P) other Components

src/components/ClaimENSPage/ClaimENSPage.tsx
nicosantangelo, line 35: This piece of code is currently being executed quite a few times, which leads to multiple requests to the blockchain and `setStates`
image
I think we can take advantage of the state we already have so we do this only when necessary, or at least a few less times. You might need to have a way to tell when the amountApproved hasn't been fetched yet, because right now the default value is `0` which is also "not approved"

nicosantangelo (changes requested):
Some other things I found:
- I18n is missing in the entire PR!
- I can send new "approval" transactions using the toggle even if a transaction is loading, I think we should disable it. I think it'd be a good idea to either add a `Popup` to the `Claim Name` button explaining whats happening (see below) or change the text of the toggle to `loading`
- We need to add a Popup to the `Claim Name` button explaining why it's disabled! For example adding a "You need to approve MANA first"

src/components/ClaimENSPage/ClaimENSPage.tsx
cazala: mm currently this would only happen if `window.ethereum` doesn't exist, like when the user doesn't have metamask or in incognito mode, so i think i could never happen at this point since you have to login before you can get here, right?

nicosantangelo: Right, ok!

nicosantangelo (approved):

PR #1187, opened 2020-11-30
feat: claim name fat finger protection

nicosantangelo:


src/components/Modals/ClaimNameFatFingerModal/ClaimNameFatFingerModal.types.ts
```diff
@@ -0,0 +1,16 @@
+import { Dispatch } from 'redux'
+import { ModalProps } from 'decentraland-dapps/dist/providers/ModalProvider/ModalProvider.types'
+import { claimNameRequest, ClaimNameRequestAction } from 'modules/ens/actions'
+
+export type Props = ModalProps & {
```
malaniz, line 5: I don't know if make sense here, but `metadata` type? maybe?

src/modules/ens/sagas.ts
```diff
@@ -183,10 +183,11 @@ function* handleFetchDomainListRequest(_action: FetchDomainListRequestAction) {
 }
 
 function* handleClaimNameRequest(action: ClaimNameRequestAction) {
-  const { ens, name } = action.payload
+  const { name } = action.payload
   try {
     const [from, eth]: [Address, Eth] = yield getCurrentAddress()
     const txHash = ''
+    const ens = {} as ENS
     console.log('handleClaimName', ens, name, eth)
```
malaniz, line 191: Make sense to continue with these `console.log`s at this point? Maybe we need to clean them from the our code.

malaniz (approved):
The code works properly, I just add 2 comments (more suggestions than requirements)

src/components/Modals/ClaimNameFatFingerModal/ClaimNameFatFingerModal.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
+import { ModalNavigation, Field, Button } from 'decentraland-ui'
+import Modal from 'decentraland-dapps/dist/containers/Modal'
+import { t, T } from 'decentraland-dapps/dist/modules/translation/utils'
+import { Props, State } from './ClaimNameFatFingerModal.types'
+import './ClaimNameFatFingerModal.css'
+
+export default class ClaimNameFatFingerModal extends React.PureComponent<Props, State> {
+  state = {
```
cazala, line 9: this needs to be `state: State = {` in order to enforce the typing, otherwise you can put whatever you want in that `state` and typescript won't complain

cazala (approved):

src/components/Modals/ClaimNameFatFingerModal/ClaimNameFatFingerModal.types.ts
nicosantangelo: I deleted it by accident! my bad

nicosantangelo:
Updated, thanks!

PR #1188, opened 2020-12-02
fix: custom smart items

cazala:
This PR fixes an issue on prod where users can't import custom smart items. This is due to a change in the way the SDK bundles the JS, we were looking for a piece of code to extract the code of the smart item, but that piece of code doesn't exist anymore, so I updated it to a new regex that matches the way the code is bundled now.

Also I improved the error reporting of duplicated assets because the current one was not clear and caused confusion, now if an item has the id of an already existing item we show the following:

Screen Shot 2020-12-02 at 5 33 42 PM


nachomazzara (approved):
LGTM.

Maybe we can add [here](https://github.com/decentraland/builder/pull/1188/files#diff-73cd3fcc8a545c17eb751a071fee042b0fbb8846cf651d79723d29235843aad8R8) what the regex is intended to match for faster readability/change in the future.

PR #1191, opened 2020-12-04
feat: save thumbnail in catalyst

cazala:
Upload thumbnails to the catalyst so the explorer doesn't need to rely on external sources (like the builder API in this case), this should improve the experience over China.

I also fixed a few tiny issues:
- The highlight color of the SceneDetailPage was wrong, it was white `#ffffff` where it should be a light red or blue depending on your role
- Some translations where broken (expecting an argument that was never passed)
- Removed outdated code that uploaded the frame of a picture frame to the content server


nicosantangelo (approved):
LGTM

Issue #1200, opened 2020-12-10
I get a full loading page when setting a resolver or ENS content


nicosantangelo:
See video attached

[set-content.zip](https://github.com/decentraland/builder/files/5675883/set-content.zip)

PR #1210, opened 2020-12-15
fix: without massive loading in fullscreen

malaniz:
Fix
https://github.com/decentraland/builder/issues/1200


nachomazzara (changes requested):
Testing it, if I refresh the page at `/assign`, everything seems to be broken


You can try by going to https://builder-git-bug-use-as-alias-oldname-changes.decentraland1.vercel.app/land/66,157/name/A_NAME_YOU_HAVE/assign

Issue #1211, opened 2020-12-15
Retry Tx on AssignEns is not updating the buttons disabling / enabling flags as expected.


malaniz:
Retry Tx on AssignEns is not updating the buttons disabling / enabling flags as expected.

To Replicate:
- go to assign ENS
- select a land
- click on `set resolver` button
- Metamask pop up, click on reject
- click on `retry tx`
- Transaction will be approved but the button will continue with `retry tx` without enabling `Set Content` button.

The same condition happens when you click on Set content and reject on metamask for retry TX.

PR #1212, opened 2020-12-16
Fix: retry tx not working

malaniz:
Fix:
Cleaning previous error state when request happens.

https://github.com/decentraland/builder/issues/1211


nachomazzara (approved):
LGTM

Issue #1204, opened 2020-12-10
Check if the name is already in use when the user presses confirm


nicosantangelo:
This was a part of the original specification here https://github.com/decentraland/builder/issues/1138

We need to show the user an error message letting them know the user is not available to avoid having a failed transaction 

Names are unique, so choosing a name already claimed will just result in an error

To me, we should add it on the first CLAIM press, to avoid multiple requests on key press

PR #1213, opened 2020-12-17
Feat: claim name check repeated

malaniz:
Fix
https://github.com/decentraland/builder/issues/1204


src/components/ClaimENSPage/ClaimENSPage.tsx
```diff
@@ -76,14 +77,24 @@ export default class ClaimENSPage extends React.PureComponent<Props, State> {
     this.setState({ isLoading: false })
   }
 
-  handleClaim = () => {
+  handleClaim = async () => {
     const { onOpenModal } = this.props
     const { name } = this.state
-    onOpenModal('ClaimNameFatFingerModal', { originalName: name })
+    const isRepeated = !(await isNameAvailable(name))
```
nachomazzara, line 83: This way it is working but Imo I feel easier to read it as follows:
I would use `const isNameAvailable = await isNameAvailable(name)`

Thoughts @nicosantangelo ?

nachomazzara:
The PR works, What do you think of adding a `loading` while we check if the name is available or not?

src/components/ClaimENSPage/ClaimENSPage.tsx
nicosantangelo: you can't really repeat the name (it'll override the import), but maybe `const isAvailable = await isNameAvailable()` ?

nicosantangelo:
Isn't a loading missing here? After looking at the code a bit I tried out and after I clicked and had to wait for it to tell me the name is not available without any indication

src/components/ClaimENSPage/ClaimENSPage.tsx
nachomazzara: Yes, I meant to avoid `!(await) promise` pattern

src/components/ClaimENSPage/ClaimENSPage.tsx
```diff
@@ -76,14 +77,26 @@ export default class ClaimENSPage extends React.PureComponent<Props, State> {
     this.setState({ isLoading: false })
   }
 
-  handleClaim = () => {
+  handleClaim = async () => {
     const { onOpenModal } = this.props
     const { name } = this.state
-    onOpenModal('ClaimNameFatFingerModal', { originalName: name })
+    this.setState({ isLoading: true })
+    const isAvailable = await isNameAvailable(name)
```
nachomazzara, line 84: What do you think of adding a `try/catch` clause here? If for some reason `isNameAvailable` throws, the UI will keep loading forever and this _check_ is not a 100% blocker for the flow since the blockchain has the final word

malaniz: good eye! I'll make it.

src/components/ClaimENSPage/ClaimENSPage.tsx
```diff
@@ -81,12 +81,17 @@ export default class ClaimENSPage extends React.PureComponent<Props, State> {
...
+        onOpenModal('ClaimNameFatFingerModal', { originalName: name })
+        this.setState({ isLoading: false })
+      } else {
+        this.setState({ isAvailable: false, isLoading: false })
+      }
+    } catch (error) {
       this.setState({ isLoading: false })
-    } else {
-      this.setState({ isAvailable: false, isLoading: false })
+      throw error
```
nachomazzara, line 94: Are we throwing errors in components? what is the purpose of the try/catch then?

src/modules/ens/utils.ts
```diff
@@ -8,8 +8,8 @@ import { createEth } from 'decentraland-dapps/dist/lib/eth'
 
 export const PRICE_IN_WEI = 100000000000000000000 // 100 MANA
 export const PRICE = fromWei(PRICE_IN_WEI.toString(), 'ether')
-
 export const MAX_NAME_SIZE = 15
+export const NAME_MIN_CHARACTERS = 1
```
nachomazzara, line 12: I think that we can use `MIN_NAME_SIZE = 2` just to follow the name of the variable above.

Issue #1196, opened 2020-12-10
Update names table copies and/or content


nicosantangelo:
The Names table currently looks like this:

![image](https://user-images.githubusercontent.com/692077/101832452-39349a00-3b16-11eb-8b3b-ac166ced86ac.png)

I see a few problems with this:
1. `Being assigned` is a continuous construction but nothing is happening. Looks like it should be called `Land` or `Assigned to` ?
2. What's assigned to supposed to be showing? Right now it's just the name repeated again once you assign your ENS to a Land. Maybe we should move `Is Alias?` here?
3. What should we show when the ENS is assigned to an Estate instead of a Parcel? We can see the ID now but I think it's not really indicative of anything (image below). Do we have a way to show this in the builder already @cazala? Maybe we can add a link/name to these assets?
![image](https://user-images.githubusercontent.com/692077/101833932-6bdf9200-3b18-11eb-83bc-430c861bd593.png)

What do you think?
@nachomazzara @malaniz

nachomazzara:
Thanks Nico, great explanation! 

I think that Mat can also help here but `Assigned to` and `Being assigned` seems redundant. I guess that maybe we can show:
`Name | Type | Assigned to | ` or  `Name | Assigned To`

Where the content should be:
`nacho.dcl.eth | Profile | avatar`
`nachoLand.dcl.eth | LAND | 10,10`
`nachoEstate.dcl.eth | Estate | 1

or 

`nacho.dcl.eth | Avatar`
`nachoLand.dcl.eth | 10,10`
`nachoEstate.dcl.eth | 1`

malaniz:
Thanks @nicosantangelo, as We talk, I have the same doubts about point 1. Mat made a difference between names and domain, but after all the solution was redundant. We can use only the name (without dcl.eth) and assinedTo the complete domain (as is showed), but maybe will continue being redundant. 
Personally, the confusion about `being assigned` and `assigned to`, is more related about the "assign" word used for 2 different things. 
One is related with the interface `assign name to a land` that is used in `being assigned` (for me, this is correct use of assign word, the is not correct the `being`, I prefer here the `assigned to`). 

And the `assigned to` that is used only to reflect that the name has a domain that you bought and you can use. In this case maybe we can use other words to describe the column, maybe: `DCL Domain` 


Btw, point 2 can be a good way to fill with data the table, but is only one row with data, only one of all rows will be the alias.

nachomazzara:
[...]

PR #1221, opened 2020-12-18
Feat: ENS List Table names and content data format Update

malaniz:
Feat/Fix:
https://github.com/decentraland/builder/issues/1196


nachomazzara (changes requested):
Great!!! 🚀 

Just one small comment: The correct link should be: `https://name.dcl.eth.link`.
In short, is missing `.link`

Issue #1214, opened 2020-12-17
Handle no profile claim name


nicosantangelo:
if a user tries to use a name as an alias or interact with the name in any way without having a Profile created, the Catalyst will break.

Check the cases [here](https://github.com/decentraland/builder/issues/1138#issuecomment-730040520)

PR #1225, opened 2020-12-21
Fix: Using new default1 profile when there is not profile.

malaniz:
We need to use a default profile when there is not profile on `use as alias` operation.
https://github.com/decentraland/builder/issues/1214


src/modules/ens/sagas.ts
```diff
@@ -76,11 +76,9 @@ function* handleSetAlias(action: SetAliasRequestAction) {
   try {
     const client = new CatalystClient(PEER_URL, 'builder')
     const entities: Entity[] = yield client.fetchEntitiesByPointers(EntityType.PROFILE, [address.toLowerCase()])
-    const entity = entities.length > 0 ? entities[0] : null
+    const defaultEntity: Entity = yield getDefaultProfileEntity()
```
nachomazzara, line 79: Even this works, I think that we are using for consistency `yield call(....)`. Am I right? same for the line above.

src/modules/ens/sagas.ts
```diff
@@ -76,11 +76,9 @@ function* handleSetAlias(action: SetAliasRequestAction) {
...
+    const entity = entities.length > 0 ? entities[0] : defaultEntity
```
nachomazzara, line 80: I think that we can avoid the default entity's request when the user already has a profile. 

```typescript
let entity: Entity
if (entities.length > 0) {
    entity = entities[0]
} else {
   entity = yield call(() => getDefaultProfileEntity())
}
```

nachomazzara (changes requested):
🚀

src/modules/ens/sagas.ts
malaniz: It looks better in that way. Fixing! Thanks

src/modules/ens/sagas.ts
malaniz: fixing!

PR #1226, opened 2020-12-21
feat: upgrade dependencies and colors

cazala:
This PR updates the decentraland libraries, removes the profile modules in favor of the ones from the libs, and also updates the operator color to be purple instead of blue


nachomazzara (approved):
LGTM

PR #1231, opened 2020-12-30
Fix: replacing text

malaniz:
Replacing texts


cazala (approved):

cazala:
No related to this PR but maybe we can fix it, the first `Send Tx` / `Approved` button is misaligned to the right:

Screen Shot 2020-12-30 at 4 50 50 PM

src/modules/translation/languages/en.json
```diff
@@ -611,16 +611,16 @@
   "land_assign_ens_page": {
     "title": "Assign {name}",
     "subtitle": "Assign a name to {land}",
-    "set_name_message": "In order to set this name to your parcel you first need to send 2 transactions.",
+    "set_name_message": "Before you can assign this name to your parcel, you need to submit 2 transactions: one that translates the human readable, alphanumeric name into an Ethereum address and one that creates a hash code linking that generated address to your LAND.",
```
nachomazzara, line 614: This is not correct. Let's use:

```diff
- [...] one that translates the human readable, alphanumeric name into an Ethereum address and one that creates a hash code linking that generated address to your LAND.

+ one that allows the ENS public resolver to translate your name, and one that resolves your name with the desired link.
```

src/modules/translation/languages/en.json
```diff
@@ -611,16 +611,16 @@
...
     "set_resolver": "Set Resolver",
-    "set_resolver_explanation": "The Resolver is a Smart Contract responsible for the process of translating names into addresses.",
+    "set_resolver_explanation": "This translates your name into an Ethereum address.",
```
nachomazzara, line 616:
```diff
- This translates your name into an Ethereum address.

+ This translates your name into an Ethereum address, IPFS hash, or other decentralized standard formats.
```

src/modules/translation/languages/en.json
```diff
@@ -611,16 +611,16 @@
...
     "set_content": "Set Content",
-    "set_content_explanation": "The Content is a hash refering a redirection to your land page that will be saved on your Resolver."
+    "set_content_explanation": "This creates a redirect to your LAND page and stores it in the Resolver."
```
nachomazzara, line 618:
```diff
- This creates a redirect to your LAND page and stores it in the Resolver

+ This resolves your name with your LAND page
```

src/modules/translation/languages/en.json
cazala: isn’t this too technical? should a user even care about understanding what does ENS stand for?

src/modules/translation/languages/en.json
```diff
@@ -611,16 +611,16 @@
...
   },
   "ens_select_land_page": {
     "title": "Assign {name}",
-    "subtitle": "You can assign this name to any of your Lands",
-    "land_selected": "Land selected:",
+    "subtitle": "You can assign this name to any of your parcels or estates.",
```
nachomazzara, line 622: The name can only be assigned to parcels @nicosantangelo @malaniz

src/modules/translation/languages/en.json
cazala: same as above, what the resolver does in the end is mapping the string that represents your name into an ethereum address, aren’t we getting too technical on this?

cazala:
Again not related to this PR, but this still reads `Send Tx`, shouldn't we changing it for something like `Submit`? 

That's like a button saying `Send Request` instead of `Submit` because it performs an http request under the hood, I don't think we need to expose the user to such technical details.

Screen Shot 2020-12-30 at 7 16 08 PM

src/modules/translation/languages/en.json
cazala: mm you are not seeing Estates because the legacy tiles api is not working in `.zone` and is necessary for generating the estates in memory, but if you switch to a working api you should be able to assign to estates to, no?

src/modules/translation/languages/en.json
cazala: instead of describing what the tx does I would rather explain _why_ are we doing it. Sure we set the Resolver, because the Resolver allows us to map an ENS to an Ehereum address or an IPFS hash, but we do it _because we want to be able to link your name to a url_ (by mapping an IPFS hash to it and then resolving a URL using https://eth.link, but the users don't care about that).

I think it's fine saying something like "The first transaction allows you to use your name as a link" and "the second transaction links your name to your land" or something like that

nachomazzara: I totally agreed with what you are saying. I was thinking it for advanced users, only focused on using the right words for the exact action.

Both examples you put above are ok for me.

src/modules/translation/languages/en.json
nachomazzara: Hmmm `[...]is mapping the string that represents your name into an Ethereum address` this is only correct when your record is a Ethereum address type. You can have an IPFS link, a Swarn link, etc.

Thinking it for a basic user, and also, based on the comment above, maybe we can go with 

```
This allows your name to be linked to a parcel.
```

src/modules/translation/languages/en.json
nachomazzara: Hmmm, It depends because you can't go to the explorer with an estate Id:
`play.decentraland.org?{estate_id}`

You need to link your name to a parcel inside the estate to have something like

`play.decentraland.org?{x},{y}`

nachomazzara (dismissed):
Thanks @malaniz !!
Remember to change the `es` and `zh` along with the `en`

src/modules/translation/languages/en.json
malaniz: Thanks guys! done!

src/modules/translation/languages/en.json
malaniz: Done! LGTM!

src/modules/translation/languages/en.json
malaniz: LGTM!

src/modules/translation/languages/en.json
malaniz: I've changed all the text to refer only Parcels.

cazala (approved):

PR #1235, opened 2021-01-15
fix: names limit

cazala:
Currently the builder loads a maximum of 100 names, this PR makes it so it loads as many as the user has. Also I changed a bit the saga so the names are loaded in parallel, which is way faster for users with many names (1000 names took me 5 minutes to load before, and now 20 seconds).


nachomazzara (approved):
LGTM

PR #1245, opened 2021-02-04
feat: update dapps. make use of login modal

nicosantangelo:
Depends on https://github.com/decentraland/decentraland-dapps/pull/99


src/components/DeployContestButton/DeployContestButton.tsx
```diff
@@ -1,10 +1,6 @@
 import * as React from 'react'
-
 import { Button } from 'decentraland-ui'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
-
-// import { ShareModalMetadata, ShareModalType } from 'components/Modals/ShareModal/ShareModal.types'
```
cazala, line 6: should we remove this?

PR #1255, opened 2021-02-25
feat: update decentraland dapps

nicosantangelo:
Depends fully on the breaking changes off https://github.com/decentraland/decentraland-dapps/pull/111


src/components/ActivityPage/Transaction/TransactionDetail/TransactionDetail.tsx
```diff
@@ -14,9 +14,9 @@ import './TransactionDetail.css'
 
 const getHref = (tx: Transaction) => {
   if (tx.status === null) {
-    return
+    return ''
   }
-  return getEtherscanHref({ txHash: tx.replacedBy || tx.hash })
+  return getTransactionHref({ txHash: tx.replacedBy || tx.hash })
```
nachomazzara, line 19: What do you think of creating an issue to move this to `dcl-dapps` directly?

src/modules/wallet/utils.ts
```diff
@@ -1,19 +1,27 @@
...
     throw new Error('Wallet not found')
   }
 
-  const currentAddress: string = yield select(getAddress)
-  if (!currentAddress) {
-    throw new Error(`Invalid from address: ${currentAddress}`)
+  return new Eth(new LegacyProviderAdapter(provider as any))
+}
+
+export function* getWallet() {
```
nachomazzara, line 18: It seems weird to call it `getWallet` and return `eth` also. Maybe we can rename it to `getEthereum` or something more generic or maybe use two different methods. But it was as it before so I'm not sure if we want to change it.

nachomazzara (approved):
LGTM

src/components/ActivityPage/Transaction/TransactionDetail/TransactionDetail.tsx
nicosantangelo: You mean adding a method that takes a Transaction and returns an `href`? I think it's a good idea but I'm not sure how specific to the builder this method is. I'll add the issue there and ask juanca, linking this comment

src/modules/wallet/utils.ts
nicosantangelo: I think it's useful to return both because we're doing 
```
const [wallet, eth]: [Wallet, Eth] = yield getWallet()
```
and using both in a ton of places. It used to be
```
const [address, eth]: [string, Eth] = yield getCurrentAddress()
```

I think splitting it into two is not a bad idea but we'd have to call two methods everywhere

nachomazzara: ok!

PR #1256, opened 2021-02-26
Feat: allow item and collection managers and minters to edit items

nachomazzara:
This PR allows managers and minters to see collections/items

### DIFF
--- a/src/modules/collection/utils.ts
+++ b/src/modules/collection/utils.ts
@@ -49,3 +49,27 @@ export function toCollectionObject(collections: Collection[]) {
     return obj
   }, {} as Record<string, Collection>)
 }
+
+export function canSeeCollection(collection: Collection, address: string) {
+  return [collection.owner, ...collection.managers, ...collection.minters].some(addr => addr.toLowerCase() === address.toLowerCase())
+}
+
+export function isOwner(collection: Collection, address?: string) {
+  return address && collection.owner.toLowerCase() === address.toLowerCase()
+}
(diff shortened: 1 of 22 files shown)


src/modules/collection/utils.ts
```diff
@@ -49,3 +49,27 @@ export function toCollectionObject(collections: Collection[]) {
     return obj
   }, {} as Record<string, Collection>)
 }
+
+export function canSeeCollection(collection: Collection, address: string) {
```
nicosantangelo, line 53: There's a lib file named address that has a few utils for these address comparisons. You can use isEqual from there!

src/modules/item/utils.ts
```diff
@@ -158,3 +145,29 @@ export async function generateImage(item: Item, width = 1024, height = 1024) {
...
+export function isComplete(item: Item) {
+  return !isEditable(item) && !!item.beneficiary && !!item.price
+}
+
+export function isEditable(item: Item) {
+  const data = item.data as WearableData
+  return !item.rarity || !data.category
+}
+
+export function isOwner(item: Item, address?: string) {
```
nicosantangelo, line 158: There's a lib file named address that has a few utils for these address comparisons. You can use isEqual from there!

nicosantangelo (commented):
LGTM, just a refactor comment and it's ready to merge !

nicosantangelo (approved):
LGTM, to valhalla!

PR #1257, opened 2021-02-27
feat: prevent repeated collection names

### LINKED ISSUE #1094: We shouldn't allow for repeated collection names

nachomazzara:
Close #1094

### DIFF
--- a/src/modules/collection/sagas.ts
+++ b/src/modules/collection/sagas.ts
@@ -2,6 +2,7 @@ import { Eth } from 'web3x-es/eth'
 import { Address } from 'web3x-es/address'
 import { replace } from 'connected-react-router'
 import { select, take, takeEvery, call, put, takeLatest } from 'redux-saga/effects'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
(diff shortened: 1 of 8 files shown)


src/modules/collection/sagas.ts
```diff
@@ -100,7 +101,11 @@ function* handleSaveCollectionRequest(action: SaveCollectionRequestAction) {
     yield put(closeModal('CreateCollectionModal'))
     yield put(closeModal('EditCollectionNameModal'))
   } catch (error) {
-    yield put(saveCollectionFailure(collection, error.message))
+    const errorMessage = error.response.data.error || error.message
+    if (errorMessage === 'Name already in use') {
+      return yield put(saveCollectionFailure(collection, t('create_collection_modal.error_name_already_in_use')))
```
nicosantangelo, line 106: This works, but it shouldn't be the sagas's responsibility to know how the message looks I think. I'd move this logic to the component that shows the error, having two calls to the same method with a translated version seems odd

nicosantangelo (commented):
comment

nicosantangelo (approved):
💋

PR #1260, opened 2021-03-01
fix: prevent fetching empty coords

cazala:
Sending an empty `pointers` array causes the catalyst to return _all_ the entities, degrading the peer's performance and returning unexpected results. It will be fixed on the catalyst, but also we'll release this patch to prevent builder users from stressing the peers until then.


nachomazzara:
Close #1260

PR #1279, opened 2021-03-16
feat: committee

nicosantangelo:
Depends on https://github.com/decentraland/builder-server/pull/182

### DIFF
--- a/src/components/ENSListPage/ENSListPage.tsx
+++ b/src/components/ENSListPage/ENSListPage.tsx
@@ -39,8 +39,8 @@ export default class ENSListPage extends React.PureComponent<Props, State> {
         direction="left"
         value={sortBy}
         options={[
-          { value: SortBy.ASC, text: t('ens_list_page.order.name_asc') },
-          { value: SortBy.DESC, text: t('ens_list_page.order.name_desc') }
+          { value: SortBy.ASC, text: t('global.order.name_asc') },
+          { value: SortBy.DESC, text: t('global.order.name_desc') }
         ]}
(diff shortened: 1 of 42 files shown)


src/components/ENSListPage/ENSListPage.tsx
```diff
@@ -51,7 +51,7 @@ export default class ENSListPage extends React.PureComponent<Props, State> {
     const { ensList } = this.props
     const { page, sortBy } = this.state
 
-    const sortedEnsList = ensList
+    return ensList
```
nachomazzara, line 54: Depaso cañaso

src/modules/collection/utils.ts
```diff
@@ -64,3 +64,7 @@ export function canMintCollectionItems(collection: Collection, address?: string)
 export function canManageCollectionItems(collection: Collection, address?: string) {
   return isOwner(collection, address) || isManager(collection, address)
 }
+
+export function hasReviews(collection: Collection) {
```
nachomazzara, line 68: I would add a message here explaining that this solution is not 100% accurate since for example metadata, or price, or beneficiary update can update this value too.

src/modules/translation/languages/en.json
```diff
@@ -869,6 +870,27 @@
     "new_collection": "New Collection",
     "empty_description": "You have no Items or Collections yet, start creating them and dress the metaverse with style"
   },
+  "curation_page": {
```
nachomazzara, line 873: What do you think of moving the repeated key/values as `Name ↑` to global?

Issue #1242, opened 2021-01-19
Create a post in the forum once the collection is deployed


nachomazzara:
- Once the collection is deployed, we should create a post in the forum with all the information needed for the reviewer to review it:
  - We should create a template for the post with images, title, etc. TBD.

- We can handle the whole topic by embedding the comments inside the collection page at the builder as https://blog.codinghorror.com/welcome-to-the-internet-of-compromised-things/#discourse-comments.

- The creator and the committee members will communicate the changes needed by using this specific topic per collection

- UI Needed

PR #1281, opened 2021-03-29
feat: curation editor page

nicosantangelo:
Closes #1241
Closes #1248
Closes #1249
Closes #1250 
Add a top panel to approve/reject collections to the editor page

PR #1283, opened 2021-03-30
feat: forum

nicosantangelo:
Continues: #1281 
Depends on: https://github.com/decentraland/builder-server/pull/185
Closes: #1242

### DIFF
--- a/src/modules/forum/actions.ts
+++ b/src/modules/forum/actions.ts
@@ -1,0 +1,23 @@
+import { action } from 'typesafe-actions'
+import { Collection } from 'modules/collection/types'
+import { ForumPost } from './types'
+
+// Create collection forum post
+
+export const CREATE_COLLECTION_FORUM_POST_REQUEST = '[Request] Create collection forum post'
+export const CREATE_COLLECTION_FORUM_POST_SUCCESS = '[Success] Create collection forum post'
+export const CREATE_COLLECTION_FORUM_POST_FAILURE = '[Failure] Create collection forum post'
+
+export const createCollectionForumPostRequest = (collection: Collection, forumPost: ForumPost) =>
+  action(CREATE_COLLECTION_FORUM_POST_REQUEST, { collection, forumPost })
+export const createCollectionForumPostSuccess = (collection: Collection, forumLink: string) =>
+  action(CREATE_COLLECTION_FORUM_POST_SUCCESS, {
+    collection,
+    forumLink
+  })
+export const createCollectionForumPostFailure = (collection: Collection, error: string) =>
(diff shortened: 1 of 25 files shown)


nachomazzara:
I test the PR and I found this:

Once you reject a collection, this modal appears:

Screen Shot 2021-04-06 at 16 43 06

The link there is wrong: https://forum.decentraland.org/t/review-collection-nachocollection-2/1342. It should be https://forum.decentraland.org/t/review-collection-nachocollection-2/599

src/modules/forum/actions.ts
```diff
@@ -0,0 +1,23 @@
...
+export const CREATE_COLLECTION_FORUM_POST_FAILURE = '[Failure] Create collection forum post'
+
+export const createCollectionForumPostRequest = (collection: Collection, forumPost: ForumPost) =>
+  action(CREATE_COLLECTION_FORUM_POST_REQUEST, { collection, forumPost })
+export const createCollectionForumPostSuccess = (collection: Collection, forumLink: string) =>
+  action(CREATE_COLLECTION_FORUM_POST_SUCCESS, {
+    collection,
+    forumLink
+  })
+export const createCollectionForumPostFailure = (collection: Collection, error: string) =>
```
nachomazzara, line 18: Do you think that we need to use this action in the reducer to display an error or something? If the post to the forum fails, there is no way to retry.  Maybe we can add an issue to cover retries of this kind later to solve weird states.

src/modules/translation/languages/es.json
```diff
@@ -979,6 +979,14 @@
   "item_editor": {
     "open": "Abrir editor",
     "top_panel": {
+      "veredict_received": "Veredicto recibido",
+      "veredict_received_successfully": "Recibimos tu veredicto correctamente!",
+      "veredict_explanation": "Puedes contarnos por qué?",
+      "go_to_forum": "Por favor visita la discusión asocida y explicale al creador el por qué de tu decisión.",
```
nachomazzara, line 985:
```suggestion
      "go_to_forum": "Por favor visita la discusión asociada y explícale al creador el por qué de tu decisión.",
```

nachomazzara (changes requested):
LGTM

src/modules/forum/actions.ts
nicosantangelo: I think creating an issue is a good idea here because this is an action hidden to the user, so it's not like showing and error on the plubished collection makes sense.

I've created this issue to track it: https://github.com/decentraland/builder/issues/1290

nachomazzara (approved):

PR #1291, opened 2021-04-07
Feat/rarities breakdown

nachomazzara:
Publish the collection is a two step modal now:

## Step 1
Screen Shot 2021-04-06 at 21 15 05

## Step:2
Screen Shot 2021-04-06 at 21 17 13

### DIFF
--- a/src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
+++ b/src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
@@ -35,21 +35,24 @@ export default class PublishCollectionModal extends React.PureComponent<Props, S
     const { items } = this.props
     const { rarities, isFetchingRarities } = this.state
 
-    const itemsByRarity: Record<string, { name: ItemRarity; count: number; price: number }> = {}
+    const itemsByRarity: Record<string, { id: ItemRarity; name: ItemRarity; count: number; price: number }> = {}
     let totalPrice = 0
 
-    if (rarities.length > 0) {
-      items.forEach(item => {
-        const rarity = rarities.find(r => r.name === item.rarity)!
-        if (!itemsByRarity[rarity.name]) {
-          itemsByRarity[rarity.name] = { name: rarity.name, count: 0, price: 0 }
-        }
+    for (const item of items) {
+      const rarity = rarities.find(rarity => rarity.name === item.rarity)
 
-        const rarityPrice = fromWei(rarity.price, 'ether')
-        itemsByRarity[rarity.name].count++
-        itemsByRarity[rarity.name].price += parseInt(rarityPrice, 10)
-        totalPrice += parseInt(rarityPrice, 10)
-      })
+      if (!rarity) {
+        continue
+      }
+
+      if (!itemsByRarity[rarity.id]) {
+        itemsByRarity[rarity.id] = { id: rarity.id, name: rarity.name, count: 0, price: 0 }
+      }
+
+      const rarityPrice = parseInt(fromWei(rarity.price, 'ether'), 10)
+      itemsByRarity[rarity.name].count++
+      itemsByRarity[rarity.name].price += rarityPrice
+      totalPrice += rarityPrice
     }
 
     return (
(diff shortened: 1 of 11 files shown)


src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
```diff
@@ -1,39 +1,120 @@
...
+  state: State = { step: 1, rarities: [], isFetchingRarities: true }
+
+  async componentDidMount() {
     const { collection, onClose } = this.props
     if (!collection) {
       onClose()
     }
+
+    const rarities = await builder.fetchRarities()
+    this.setState({ rarities, isFetchingRarities: false })
```
nicosantangelo, line 22: Should we add this to the "hande error states" issue? (as it'd load forever, right? there's no `trycatch`). Wouldn't fix it here.

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
```diff
@@ -1,39 +1,120 @@
...
 
+  handleProceed = () => {
+    this.setState({ step: 2 })
+  }
+
+  renderFirstStep = () => {
+    const { items } = this.props
+    const { rarities, isFetchingRarities } = this.state
+
+    const itemsByRarity: Record<string, { name: ItemRarity; count: number; price: number }> = {}
```
nicosantangelo, line 38: I think this type should be indexed by ItemRarity, right? And, should we use rarity.id below?

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
```diff
@@ -1,39 +1,120 @@
...
+    let totalPrice = 0
+
+    if (rarities.length > 0) {
+      items.forEach(item => {
+        const rarity = rarities.find(r => r.name === item.rarity)!
```
nicosantangelo, line 43: Two things, non important:
- Most of the codebase uses `for..of` for these kinds of things, leaving forEach for renders (where for...of does not work) or really particular cases. **It's the same thing** here, as you don't need to break early, **but** if instead of using `!` we'd like to skip the rarity, on a for...of we could do `if(!rarity) continue`. No need to change it
- Can we call `r` -> `rarity`? it's not that many more caracters but's a bit more clear

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
```diff
@@ -1,39 +1,120 @@
...
+        if (!itemsByRarity[rarity.name]) {
+          itemsByRarity[rarity.name] = { name: rarity.name, count: 0, price: 0 }
+        }
+
+        const rarityPrice = fromWei(rarity.price, 'ether')
```
nicosantangelo, line 48: You could parseInt here to avoid doing it twice if you want

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
```diff
@@ -1,39 +1,120 @@
...
+        {isFetchingRarities ? (
+          <div className="loader-wrapper">
+            <Loader size="big" active={isFetchingRarities} />
+          </div>
+        ) : (
+          <>
+            {t('publish_collection_modal.items_breakdown_title', { items: items.length })}
+            <div className="items-breakdown">
+              {Object.values(itemsByRarity).map(itemByRarity => {
+                return (
```
nicosantangelo, line 66: We can change this to
```suggestion
              {Object.values(itemsByRarity).map(itemByRarity => 
                 (
```

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
nicosantangelo: Same goes for this section: https://github.com/decentraland/builder/pull/1291/files#diff-9088b58a90f358c685624ad30bde921a26304e094bd9fe43ca4cef8007a82a33R69, reading the code the `name` seems less of a slug that what `id` does

src/lib/api/builder.ts
```diff
@@ -618,9 +618,13 @@ export class BuilderAPI extends BaseAPI {
     return fromRemoteWeeklyStats(remoteStats)
   }
 
-  async getCommittee(): Promise<string[]> {
+  async fetchCommittee(): Promise<string[]> {
```
nicosantangelo, line 621: 💯 didn't see that there...wanna change `getWeeklyStats` above? 👀

nicosantangelo (commented):
Great job!

nicosantangelo (approved):

PR #1293, opened 2021-04-07
fix: fixes on deploy. add edition to published items

nicosantangelo:


src/components/Modals/MintItemsModal/MintItemsModal.tsx
```diff
@@ -103,8 +107,8 @@ export default class MintItemsModal extends React.PureComponent<Props, State> {
                   {t('global.cancel')}
                 </Button>
               ) : (
-                <Button primary onClick={this.handleMintItems} loading={isLoading}>
-                  {t('global.done')}
+                <Button primary onClick={this.handleMintItems} loading={isLoading} disabled={this.isDisabled()}>
```
nachomazzara, line 110: Maybe we can move `this.isDisabled()` before the render's _return_ to just use variables for components

src/modules/item/export.ts
```diff
@@ -41,9 +42,9 @@ async function getFiles(contents: Record<string, string>) {
   }, {})
 }
 
-function toCatalystItem(collection: Collection, item: Item): CatalystItem {
+function toCatalystItem(collection: Collection, item: Item, chainId: ChainId): CatalystItem {
   return {
-    id: item.id,
+    id: getCatalystURN(collection, item, chainId),
```
nachomazzara, line 47: I know that we have only collections right now, but maybe we can call it `getCatalystCollectionItemURN` or `getCatalystItemURN`. What do you think?

src/modules/translation/languages/en.json
```diff
@@ -951,7 +951,8 @@
   "mint_items_modal": {
     "title": "Mint Items",
     "add_item": "Add Item",
-    "no_items": "You have no items available to mint for \"{name}\""
+    "no_items": "You have no items available to mint for \"{name}\"",
+    "mint": "Mint"
```
nachomazzara, line 955: What do you think of moving `mint` to global?

nachomazzara (commented):
🚀  🧹

src/modules/item/export.ts
nicosantangelo: Oh good point, didn't realize it was scoped by collection

src/modules/translation/languages/en.json
nicosantangelo: will do

PR #1295, opened 2021-04-12
feat: use store address from dcl-transactions

nicosantangelo:


src/components/Modals/SellCollectionModal/SellCollectionModal.tsx
```diff
@@ -9,14 +9,14 @@ import './SellCollectionModal.css'
 
 export default class SellCollectionModal extends React.PureComponent<Props> {
   handleSell = () => {
-    const { collection, metadata, onSetMinters } = this.props
-    onSetMinters(collection, setOnSale(collection, metadata.isOnSale))
+    const { collection, wallet, metadata, onSetMinters } = this.props
+    onSetMinters(collection, setOnSale(collection, wallet.chainId, metadata.isOnSale))
```
nachomazzara, line 13: I think that we will need to use the mapped `chainId` instead of the connected one

nachomazzara (changes requested):
LGTM!

PR #1302, opened 2021-04-16
feat update thumbnails

nicosantangelo:
Closes https://github.com/decentraland/account/issues/107

### DIFF
--- a/src/components/AssetPackEditor/AssetPackEditor.tsx
+++ b/src/components/AssetPackEditor/AssetPackEditor.tsx
@@ -12,7 +12,7 @@ import Icon from 'components/Icon'
 import { Props, State } from './AssetPackEditor.types'
 import './AssetPackEditor.css'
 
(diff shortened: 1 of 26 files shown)


nachomazzara:
Thanks @nicosantangelo 

I get some feedback while testing it.

### Any user can change see and upload a thumbnail for any wearable

Maybe we can prevent displaying the thumbnail edition if the user does not have permission to edit the item.

### Camera icon is missing margin

The icon is close/over the thumbnail square.

Screen Shot 2021-04-20 at 09 23 02

### Unity Error once I update a non-published item thumbnail

Screen Shot 2021-04-20 at 13 40 02


### Infinite loading when updating a published item thumbnail

The code is looking for the files in the `builder-server` but the images are not there cause they have been uploaded only in the UI.

Maybe `onSavePublishedItem(item)` at `RightPanel.tsx` should use the `contents`.

Screen Shot 2021-04-20 at 10 15 53


I know that this is not a change of this PR but at `/modules/item/reducer.ts` `SAVE_PUBLISHED_ITEM_SUCCESS` is used twice for different case clauses.

Didn't see the code yet.

PR #1321, opened 2021-04-24
feat: add babylon support to thumbnail renderer

cazala:
Added an  `engine` option to the `getModelData` helper to select to whether use ThreeJS or BabylonJS as the rendering engine. This is to allow wearables to be rendered on Babylon, which doesn't have the issues importing GLTFs from Blender >-v2.8. This is a first step towards removing ThreeJS completely, but in order to do that we need to first add support for `mappings` to the Babylon renderer and also make it compute the models `metrics`. The implementation in this PR still uses ThreeJS to compute the metrics and is used for the custom asset thumbnails which require mappings.


src/components/Modals/CreateItemModal/CreateItemModal.tsx
```diff
@@ -403,7 +403,12 @@ export default class CreateItemModal extends React.PureComponent<Props, State> {
       }
     } else {
       const url = URL.createObjectURL(contents[model])
-      const { image, info } = await getModelData(url, { width: 1024, height: 1024 })
+      const { image, info } = await getModelData(url, {
+        width: 1024,
+        height: 1024,
+        extension: model.split('.').pop(),
```
nicosantangelo, line 409: I think we can use `getExtension` from file.ts both here and in line 435 below if you think is useful

We can probably replace a few more places with that method, but did a quick grep on the codebase and could not find anything

nicosantangelo (commented):
🔥

PR #1345, opened 2021-05-05
fix: zip files with mask

nachomazzara:
Testing files:

[Archive.zip](https://github.com/decentraland/builder/files/6424551/Archive.zip)
[hwn_2020_ddlm_01_eyes.zip](https://github.com/decentraland/builder/files/6424552/hwn_2020_ddlm_01_eyes.zip)

Current:
Screen Shot 2021-05-04 at 21 32 43

Expected:
Screen Shot 2021-05-04 at 21 35 30

### DIFF
--- a/src/modules/item/utils.ts
+++ b/src/modules/item/utils.ts
@@ -240,11 +240,16 @@ export function getCategories(contents: Record<string, any> | undefined = {}) {
     : SIMPLE_WEARABLE_CATEGORIES
 }
 
+export function isImageFile(fileName: string) {
+  return fileName.endsWith('.png')
+}
+
 export function isComplexFile(fileName: string) {
   return fileName.endsWith('.gltf') || fileName.endsWith('.glb')
 }
 
-export function isItemPNGFile(fileName: string) {
-  const name = fileName.toLowerCase()
-  return name.indexOf(THUMBNAIL_PATH) === -1 && !name.includes('_mask') && fileName.endsWith('.png')
+export function isModelPath(fileName: string) {
+  fileName = fileName.toLowerCase()
+  const isMask = fileName.includes('_mask')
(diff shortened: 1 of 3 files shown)


src/modules/item/utils.ts
```diff
@@ -242,3 +243,8 @@ export function getCategories(contents: Record<string, any> | undefined = {}) {
 export function isComplexFile(fileName: string) {
   return fileName.endsWith('.gltf') || fileName.endsWith('.glb')
 }
+
+export function isItemPNGFile(fileName: string) {
+  const name = fileName.toLowerCase()
+  return name.indexOf(THUMBNAIL_PATH) === -1 && !name.includes('_mask') && fileName.endsWith('.png')
+}
```
nicosantangelo, line 250: I don't know about this. The method is called `isItemPNGFile` but it receives a filename and it's checking for a `_mask`. There's also already a `isPNGModel` in `CreateItemModal` but that one is not checking for _mask (just png's like the name implies).

It seems to me that maybe we could just extract the original method on `CreateItemModal` and use some of the methods there?

```diff
const modelPath = fileNames.find(
-   fileName => isComplexFile(fileName) || (fileName.indexOf(THUMBNAIL_PATH) === -1 && fileName.endsWith('png'))
+   fileName => this.isModelPath(filename)
)
```

```ts
  isModelPath(fileName: string) {
    fileName = fileName.toLowerCase()
    const isMask = fileName.includes('_mask')
    return isComplexFile(fileName) || (fileName.indexOf(THUMBNAIL_PATH) === -1 && !isMask && this.isPNGModel(filename))
  }
```

what do you think?
(coded here, not ment to copypasta)

nicosantangelo (commented):
🚀

nicosantangelo (approved):

PR #1373, opened 2021-05-19
feat: pre fill item data

### LINKED ISSUE #1369: Set item as free and pre-fill with the connected address as beneficiary

nicosantangelo:
Closes #1369 

Also fixes
- entering to the settings page without being logged in which explodes in production
- adds the corresponding Polygon icons where appropiate
- fixes conditions on the edit price modal that allowed for incorrect submits
- adds Mana icon to that modal too


nachomazzara:
Excellent!!

Some things to change:

- If I set as free an item, the beneficiary should be the `0x000...0` address and the price should be 0.
- The `set as free` and `i'm the beneficiary` can't be both selected. Or you are the beneficiary or the item is free.

PR #1375, opened 2021-05-20
feat: matic intro

cazala:


src/components/IntroPopup/IntroPopup.tsx
```diff
@@ -0,0 +1,62 @@
...
+        {open ? <div className="IntroPopupOverlay" /> : null}
+        <Popup
+          position="top center"
+          open={open}
+          className="IntroPopup"
+          content={
+            <Column>
+              <Header>{t('intro_popup.title')}</Header>
+              <p className="content">
+                {
```
nicosantangelo, line 32: no need for these {} right?

nicosantangelo (approved):
LGTM

src/components/IntroPopup/IntroPopup.tsx
```diff
@@ -0,0 +1,62 @@
...
+                    id="intro_popup.content"
+                    values={{
+                      br: <br />,
+                      blog_link: (
+                        <a href="https://decentraland.org/blog/announcements/polygon-mana/" target="_blank" rel="no:opener no:referrer">
+                          {t('intro_popup.blog_link')}
+                        </a>
+                      ),
+                      account_link: (
+                        <a href="https://account.decentraland.org" target="_blank" rel="no:opener no:referrer">
```
nicosantangelo, line 43: I think the rel should be `noopener noreferrer` right?

nicosantangelo, line 43: I think this is fine, as this popup (supposedly) will not be here forever, but account.decentralad.org already exists as an env variable in REACT_APP_ACCOUNT_URL

nicosantangelo (commented):
Two more silly comments

PR #1440, opened 2021-06-10
fix: undefined component while export

pbosio:
fixes builder in world issue: https://github.com/decentraland/unity-renderer/issues/578
builder in world adds several unsupported components to entities.
`component` record is filtered by types but unsupported components in entities are still trying to be added while exporting


src/modules/project/export.ts
```diff
@@ -180,7 +180,7 @@ export async function createGameFile(args: { project: Project; scene: Scene; rot
       for (const componentId of entity.components) {
         const component = components[componentId]
```
pbosio, line 181: `componentId` might not exist in `components` Record

nicosantangelo (approved):

PR #1442, opened 2021-06-12
feat: global minters

### LINKED ISSUE #1388: Remove collaborator should send a transaction
https://images.zenhubusercontent.com/162741822/4494a7b6-5e43-4f26-9a21-33e32a318789/screen_recording_2021_05_25_at_19_43_27.mov

### LINKED ISSUE #1421: Set global minters
Should work as set a collaborator

nicosantangelo:
Closes #1421 
Closes #1388


nachomazzara:
Feedback from usage:

- As a minter I should only see `mint`. I can't edit the item files in the editor or the price.

Screen Shot 2021-06-14 at 16 58 47

- When the collection is on sale and I want to remove a winter, it removes also the store as a winter.

- Weird behavior when I tried to remove and add another collaborator within the same txs

https://user-images.githubusercontent.com/7549152/121952626-e17c2300-cd32-11eb-9c05-4ecbfca8fb01.mov

src/components/Modals/ManageCollectionRoleModal/ManageCollectionRoleModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+  handleCancelNew = (index: number) => {
+    this.setState({
+      roles: this.removeRoleAtIndex(index)
+    })
+  }
+
+  handleSubmit = () => {
+    const { collection, metadata } = this.props
+    const { roles } = this.state
+    const originalRoles = metadata.roles
```
nachomazzara, line 57: If the collection is on sale we are "removing" the store address because it is not at `this.state.roles` but at `originalRoles`, right?

src/components/Modals/ManageCollectionRoleModal/ManageCollectionRoleModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+    const { wallet } = this.props
+    return getSaleAddress(wallet.networks.MATIC.chainId)
+  }
+
+  setRoles(accessList: Access[]) {
+    const { wallet, metadata, collection, onSetManagers, onSetMinters } = this.props
+    const { type } = metadata
+
+    // Restore the store access to the collection, removed on the first `state`
+    accessList.push(...setOnSale(collection, wallet, isOnSale(collection, wallet)))
```
nachomazzara, line 92: Following the comment above. We are "adding" again the store here if it should be turned on, and therefore the transaction has a _turn off and on_ of the store. `[..., store, store], [...,false, true]`

src/components/Modals/ManageCollectionRoleModal/ManageCollectionRoleModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+        return onSetManagers(collection, accessList)
+      case RoleType.MINTER:
+        return onSetMinters(collection, accessList)
+      default:
+        throw new Error(`Invalid role type ${type}`)
+    }
+  }
+
+  hasRoleChanged() {
+    return this.getStartRoles().length !== this.state.roles.filter(role => !!role).length
```
nachomazzara, line 105: May are we missing the case when a user removes and adds another address? I guess that comparing against the length is not 100% correct.

src/components/Modals/ManageCollectionRoleModal/ManageCollectionRoleModal.tsx
nachomazzara: I guess that the same is happening when trying to add collaborators

src/components/Modals/ManageCollectionRoleModal/ManageCollectionRoleModal.tsx
nachomazzara: Can we omit the store address if we are editing collaborators?

src/modules/translation/languages/en.json
```diff
@@ -794,6 +794,8 @@
     "collection_items_minted": "Minted {count} items for the collection {name}",
     "added_collection_managers": "Added {count} {count, plural, one {collaborator} other {collaborators}} on the collection {name}",
     "removed_collection_managers": "Removed {count} {count, plural, one {collaborator} other {collaborators}} for the collection {name}",
+    "added_collection_minters": "Added {count} {count, plural, one {minter} other {minterss}} on the collection {name}",
```
nachomazzara, line 797: missed for `es`  and `zh`

src/modules/translation/languages/en.json
```diff
@@ -925,6 +927,7 @@
     "see_in_world": "See in world",
     "unpublished": "This collection not published yet",
     "managers": "Collaborators",
+    "minters": "Minters",
```
nachomazzara, line 930: Missed for `es` and `zh`

src/modules/translation/languages/es.json
```diff
@@ -996,12 +996,20 @@
...
+  "manage_collection_role_modal": {
+    "manager": {
+      "title": "Colaboradores",
+      "subtitle": "Si has colaborado con otros creadores, agrega sus direcciones de ETH aquí",
+      "add_new": "Agregar Colaborador",
+      "empty": "No hay colaboradores aún, empieza"
+    },
+    "minter": {
+      "title": "Creadores",
+      "subtitle": "Enter the ETH address you want to allow for minting your items",
```
nachomazzara, line 1008: too much spanglish

nachomazzara (changes requested):
Thanks nico! just a few comments.

Also, I'm still seeing the collection with the option as `Open in editor` when I'm only a minter.

nicosantangelo:
> Also, I'm still seeing the collection with the option as Open in editor when I'm only a minter.

oh, I did that on purpose because the collection will be published and thus, can be seen by everyone, right? @nachomazzara

nicosantangelo:
# Tests

## Managers

- Try to visit a collection you don't have access to and check it says not found

- Add 1> collaborators
	- Check activity to see if the text is correct
	- Use one of those addresses to check if you can enter and manage the collection

- Remove 1> collaborators
	- Check activity to see if the text is correct
	- Use one of those addresses to check if you can no longer enter the collection

## Minters

- Try to visit a collection you don't have access to and check it says not found

- Using a collection on sale and not on sale
	◦ Add 1> minters
		- Check activity to see if the text is correct
		- Use one of those addresses to check if you can enter and ONLY mint the collection items

	◦ Remove 1> minters
		- Check activity to see if the text is correct
		- Use one of those addresses to check if you can no longer enter the collection

## Both

- For both collaborators and minters, play around with the modal and check that the CONFIRM button sets to disabled when the inital state is reached (adding and removing addresses).
	- Play special attention to the Minters section, check with the collection on sale and NOT on sale

nachomazzara:
More feedback:

- If I remove a collaborator and add another one, I get:
Screen Shot 2021-06-15 at 21 15 34

- If I remove a minter, I get:  

Screen Shot 2021-06-15 at 21 13 26

- If I add a minter, I get: 

Screen Shot 2021-06-15 at 21 16 31

nachomazzara (approved):

PR #1463, opened 2021-07-01
feat: disable curation if items are being uploaded

### LINKED ISSUE #1459: Prevent committee members to approve a collection if the item is not in the catalysts

nicosantangelo:
Closes #1459

### DIFF
--- a/src/modules/translation/languages/en.json
+++ b/src/modules/translation/languages/en.json
@@ -1020,6 +1020,7 @@
       "discussion": "Discussion",
       "keep_reviewing": "Keep reviewing",
       "back_to_curation": "Back to curation",
+      "waiting_for_upload": "Some items in the collection are still being uploaded. Please check back soon to proceed with the curation.",
       "reject": {
         "title": "Reject Collection",
         "subtitle": "I guess it's just not good enough, huh?",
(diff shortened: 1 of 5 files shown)


src/modules/translation/languages/en.json
```diff
@@ -1020,6 +1020,7 @@
       "discussion": "Discussion",
       "keep_reviewing": "Keep reviewing",
       "back_to_curation": "Back to curation",
+      "waiting_for_upload": "Some items in the collection are still being uploaded. Please check back soon to proceed with the curation.",
```
nachomazzara, line 1023: What do you think of mention where the item needs to be uploaded?

`[...] still being uploaded to the catalysts`

nicosantangelo: I wasn't sure the user knew what that meant (there's no mention of catalysts on the entire site). Maybe is a good idea anyways? I'm not sure if members of the committee know what a Catalyst is (besides...shibu)

nachomazzara: I guess that will help them to know about the catalyst and also will be easy to report

nicosantangelo: agree, ok! added

PR #1467, opened 2021-07-06
fix: avoid Date.now()

nchamo:
If we use `Date.now()`, we depend on the local machine's clock. If the clock is off, the deployment will fail. To avoid this issue, we will use the server's time.


nachomazzara (approved):
LGTM

PR #1478, opened 2021-07-16
fix: Change environment variable to reflect the correct environment

LautaroPetaccio:
This PR changes the environment variable used for rollbar to another one reflecting the correct environment.


src/modules/analytics/rollbar.ts
```diff
@@ -1,28 +1,22 @@
 import { env } from 'decentraland-commons'
+import { isDevelopment } from 'lib/environment'
 
 let _rollbarConfig = {
   accessToken: env.get('REACT_APP_ROLLBAR_ACCESS_TOKEN', ''),
   captureUncaught: true,
   payload: {
-    environment: env.get('NODE_ENV'),
+    environment: env.get('REACT_APP_ENVIRONMENT'),
```
nachomazzara, line 8: What do you think of using the exported variable `development` from `lib/environment` here.

nachomazzara (changes requested):
Just one minor comment

PR #1487, opened 2021-07-22
feat: get thumbnails from content server

### LINKED ISSUE #1486: Use content server instead of builder-server if the item is in the catalyst
If the item is in the catalyst, we should use  `https://peer.decentraland.org/content/contents/` instead of `https://builder-api.decentraland.org/v1/storage/contents/`. 

Example: `https://peer.decentraland.org/content/contents/QmWWqMKjfKHnAouoMd8NjL9NZ8xGnuo1sTG8NdAdFazbDi` instead of 
`https://builder-api.decentraland.org/v1/storage/contents/QmWWqMKjfKHnAouoMd8NjL9NZ8xGnuo1sTG8NdAdFazbDi`

nachomazzara:
Close #1486

### DIFF
--- a/src/lib/api/peer.ts
+++ b/src/lib/api/peer.ts
@@ -4,6 +4,8 @@ import { ContentServiceScene } from 'modules/deployment/types'
 
 export const PEER_URL = env.get('REACT_APP_PEER_URL', '')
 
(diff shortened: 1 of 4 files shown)


nicosantangelo (approved):
LGTM

PR #1492, opened 2021-07-23
fix: namespace all files with asset id to prevent overrides

cazala:
This PR fixes an issue where, if two assets had some file with the exact same path + filename, the mappings from one asset would override the other. This only happened with static assets, since they were not namespaced with the asset id, and were saved all together under a `/models` folder. The smart items were namespaced, but stored at the root level. To keep things more consistent, I stored all the assets (either smart or not) under an `/assets` folder, and they are all namespaced using the asset id.

Some assets were reported to have this behavior, like most of the fountain assets.

A scene like this:
Screen Shot 2021-07-23 at 5 00 35 PM

Would end up like this once deployed (notice the repeated models):
Screen Shot 2021-07-23 at 4 32 23 PM

After this fix, it looks like this:
Screen Shot 2021-07-23 at 4 33 32 PM

This issue also affected exported scenes (.zip files to be used with the SDK), which now also works as expected:
Screen Shot 2021-07-23 at 4 53 23 PM

### DIFF
--- a/src/modules/project/export.ts
+++ b/src/modules/project/export.ts
@@ -14,7 +14,7 @@ import { Project, Manifest } from 'modules/project/types'
 import { Scene, ComponentType, ComponentDefinition } from 'modules/scene/types'
 import { getContentsStorageUrl } from 'lib/api/builder'
 import { getParcelOrientation } from './utils'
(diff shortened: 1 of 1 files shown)


nachomazzara (approved):
Lgtm

PR #1499, opened 2021-08-04
feat: Retry forum post

LautaroPetaccio:
This PR adds the retrying process to create a forum post of a recently published collection.


src/modules/collection/sagas.ts
```diff
@@ -379,12 +379,10 @@ function* handleRequestCollectionSuccess() {
   try {
     const collections: Collection[] = yield select(getWalletCollections)
 
+    // TODO: This is done synchronously, is this something we can / want to change?
```
LautaroPetaccio, line 382: Changing this might not be easy, but do we want it to work synchronously for some reason?

src/modules/item/sagas.ts
```diff
@@ -87,6 +87,7 @@ export function* itemSaga() {
   yield takeEvery(FETCH_COLLECTION_REQUEST, handleFetchCollectionRequest)
   yield takeLatest(FETCH_TRANSACTION_SUCCESS, handleTransactionSuccess)
 
+  // TODO: Why do we have take latest here? What if there are multiple failures waiting to be handled?
```
LautaroPetaccio, line 90: Why do we take latest here?

src/modules/forum/sagas.ts
```diff
@@ -2,21 +2,51 @@ import { takeEvery, call, put } from 'redux-saga/effects'
...
   try {
-    const forumLink: string = yield call(() => builder.createCollectionForumPost(collection, forumPost))
+    console.log('Creating forum post request')
+    const forumLink: string = yield call([builder, 'createCollectionForumPost'], collection, forumPost)
     yield put(createCollectionForumPostSuccess(collection, forumLink))
   } catch (error) {
-    yield put(createCollectionForumPostFailure(collection, error.message))
+    console.log('Failed forum post request', error)
+    const forumPostAlreadyExists = error?.response?.status === 409
+    const forumLink = error?.response?.data?.data?.forum_link
```
nicosantangelo, line 30: I think we can maybe take advantage of the fact that we're using Axios behind the scenes and type this a bit better. We already have a private method on builder.ts which does:

```ts
private isAxiosError(error: any): error is AxiosError {
```

so if we make it public and use it here we could

```ts
if (builder.isAxiosError(error)) {
  // error correctly typed here
}
```

to maybe avoid a few `.?`

src/modules/forum/sagas.ts
```diff
@@ -2,21 +2,51 @@ import { takeEvery, call, put } from 'redux-saga/effects'
...
+    if (forumPostAlreadyExists && forumLink) {
+      console.log('It was a 409')
+      yield put(createCollectionForumPostSuccess(collection, error?.response?.data?.data?.forum_link))
+    }
+    console.log('It wasnt a 409')
+    yield put(createCollectionForumPostFailure(collection, forumPost, error.message))
```
nicosantangelo, line 36: I don't know if it was on purpose but maybe we can remove the `logs`?

src/modules/item/sagas.ts
nicosantangelo: shouldn't it be the same? I mean, we are trying to fix it just once so handling multiple failure states should be the same as just handling the last one

src/modules/forum/sagas.ts
```diff
@@ -2,21 +2,51 @@ import { takeEvery, call, put } from 'redux-saga/effects'
...
+
+/**
+ * Handles the post to forum request failures by re-launching the action to retry
+ * the post undefinetly until it works.
+ *
+ * @param action - The failing action.
+ */
+function* handleCreateForumPostFailure(action: CreateCollectionForumPostFailureAction) {
+  const { collection, forumPost } = action.payload
+  console.log('Doing a retry')
```
nicosantangelo, line 48: Again, just a question, but maybe we can remove the log right?

src/modules/forum/sagas.ts
```diff
@@ -2,21 +2,51 @@ import { takeEvery, call, put } from 'redux-saga/effects'
...
+      yield put(createCollectionForumPostSuccess(collection, error?.response?.data?.data?.forum_link))
+    }
+    console.log('It wasnt a 409')
+    yield put(createCollectionForumPostFailure(collection, forumPost, error.message))
   }
 }
+
+/**
+ * Handles the post to forum request failures by re-launching the action to retry
+ * the post undefinetly until it works.
```
nicosantangelo, line 42:
```suggestion
 * the post indefinitely until it works.
```

🤓

src/modules/collection/sagas.ts
```diff
@@ -420,6 +410,31 @@ function* handleTransactionSuccess(action: FetchTransactionSuccessAction) {
...
 }
 
+/**
+ * Post proccesses a collection that was published to the blockchain by singaling the
+ * builder server that the collecton has been published, setting the item ids,
+ * deploys the item entities to the Catalyst sever and creates the forum post.
+ *
+ * @param collection - The collection to post process.
+ */
+function* postProcessCollectionTransaction(collection: Collection) {
```
nicosantangelo, line 420: Do you think we can figure out a concept for this method's name so we can avoid calling it `postProcess`?  I was thinking `finishCollectionPublish` or `complete(Collection)Publish` or something similar?

What do you think?

nicosantangelo (changes requested):
Mostly wording

src/modules/collection/sagas.ts
```diff
@@ -420,6 +410,31 @@ function* handleTransactionSuccess(action: FetchTransactionSuccessAction) {
...
+  const avatarName: string | null = yield select(getName)
```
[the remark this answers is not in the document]
LautaroPetaccio: I really would prefer that we didn't do any selects after dispatching actions in a sagas as it can lead to really complex side effects that are hard to debug. I moved it upwards to keep it before any other generator / async function that could change it. TBH, I would move these selectors to the functions that use `postProcessCollectionTransaction`, but `deployItems` and `publishCollection` already use selectors.

src/modules/item/sagas.ts
LautaroPetaccio: If we use latest, how do you handle two deploy or set items that need to be retried?
What if you load two collections (by doing a refresh of the site) that have been created in the blockchain but their items don't have their ids or their content wasn't deployed to the Catalyst. As far as I know, the post-processing of those elements will be done at the same time. If for example, the set items id request fails for both of them it could happen that only one of them is retried.
Also, there's no case were a failure action would appear more than once because the retrying mechanism happens only when one of those actions is handled.

src/modules/forum/sagas.ts
```diff
@@ -2,21 +2,51 @@ import { takeEvery, call, put } from 'redux-saga/effects'
...
+    const forumLink: string = yield call([builder, 'createCollectionForumPost'], collection, forumPost)
     yield put(createCollectionForumPostSuccess(collection, forumLink))
   } catch (error) {
-    yield put(createCollectionForumPostFailure(collection, error.message))
+    console.log('Failed forum post request', error)
+    const forumPostAlreadyExists = error?.response?.status === 409
+    const forumLink = error?.response?.data?.data?.forum_link
+    if (forumPostAlreadyExists && forumLink) {
+      console.log('It was a 409')
+      yield put(createCollectionForumPostSuccess(collection, error?.response?.data?.data?.forum_link))
```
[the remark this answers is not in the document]
LautaroPetaccio: Yes! we can, changed it

src/modules/collection/sagas.ts
LautaroPetaccio: Maybe we can create a ticket and refactor this with some testing involved.

src/modules/collection/sagas.ts
LautaroPetaccio: Changed it to `finishCollectionPublishing`

src/modules/forum/sagas.ts
LautaroPetaccio: The thing is that the only thing that I can remove is the optional chaining operator of error, that is the one that checks that error isn't null. That's because the first property, response, can be undefined in cases where the exception comes from Axios and in case where it doesn't. [Here's an example](https://github.com/axios/axios#handling-errors) on handling error that come from Axios.
I've removed that chaining operator to be a bit more clear on what we can receive.

src/modules/collection/sagas.ts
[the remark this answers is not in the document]
nicosantangelo: 100%, my bad. Did not realize that we where doing dispatches above the if I referenced! this is 💯

src/modules/item/sagas.ts
nicosantangelo: You're correct! In my mind this FAILURE case would retry _all collections/items_ and not the one that errored out. I think this should be a `takeEvery`

nicosantangelo (approved):

PR #1502, opened 2021-08-09
feat: update peer

nachomazzara:


.ci/.env.prd
```diff
@@ -4,7 +4,7 @@ PUBLIC_URL=/
 REACT_APP_CHAIN_ID=1
 REACT_APP_LOCAL_STORAGE_KEY=builder
 REACT_APP_EMAIL_SERVER_URL=https://decentraland.org/subscribe
-REACT_APP_PEER_URL=https://peer.decentraland.org
+REACT_APP_PEER_URL=https://peer-lb.decentraland.org
```
nachomazzara, line 7: [`getCollectionBaseURI`](https://github.com/decentraland/builder/blob/9dddc44ef63a372d84ec55520702f35bd06f893b/src/modules/collection/utils.ts#L55) needs to use `peer.decentraland.org`

.ci/.env.stg
```diff
@@ -4,7 +4,7 @@ PUBLIC_URL=/
 REACT_APP_CHAIN_ID=1
 REACT_APP_LOCAL_STORAGE_KEY=builder
 REACT_APP_EMAIL_SERVER_URL=https://decentraland.org/subscribe
-REACT_APP_PEER_URL=https://peer.decentraland.org
+REACT_APP_PEER_URL=https://peer-lb.decentraland.org
```
nachomazzara, line 7: [`getCollectionBaseURI`](https://github.com/decentraland/builder/blob/9dddc44ef63a372d84ec55520702f35bd06f893b/src/modules/collection/utils.ts#L55) needs to use `peer.decentraland.org`

.ci/.env.prd
nicosantangelo: you mean that the method should prefix without the new `-lb` dns?

nachomazzara: because that info will go on-chain and will be impossible to be changed

PR #1503, opened 2021-08-09
feat: Add TOS approval modal

LautaroPetaccio:
This PR adds a TOS step to the `PublishCollectionModal`.

TOS


src/components/CollectionDetailPage/CollectionDetailPage.tsx
```diff
@@ -190,7 +196,7 @@ export default class CollectionDetailPage extends React.PureComponent<Props, Sta
                         />
                       )
                     ) : (
-                      <Button primary compact disabled={!this.canPublish()} onClick={this.handlePublish}>
+                      <Button primary compact onClick={this.handlePublish}>
```
nicosantangelo, line 199: is there any reason to remove this check? (`canPublish`) it's letting me publish empty/incomplete collections

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
```diff
@@ -62,93 +79,165 @@ export default class PublishCollectionModal extends React.PureComponent<Props, S
...
+          </a>
+          <Button primary fluid onClick={this.handleNextStep}>
+            {t('global.next')}
+          </Button>
+        </Modal.Content>
+      </>
+    )
+  }
+
+  renderStep = () => {
```
nicosantangelo, line 168: Just for my own selfish OCD, could we move this method after `renderThirdStep` ?

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
```diff
@@ -62,93 +79,165 @@ export default class PublishCollectionModal extends React.PureComponent<Props, S
...
+      case 2:
+        return this.renderSecondStep()
+      case 3:
+        return this.renderThridStep()
+      default:
+        throw new Error('Step not found')
+    }
+  }
+
+  renderThridStep = () => {
```
nicosantangelo, line 182:
```suggestion
  renderThirdStep = () => {
```

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
```diff
@@ -62,93 +79,165 @@ export default class PublishCollectionModal extends React.PureComponent<Props, S
...
+                      {t('publish_collection_modal.content_policy')}
+                    </a>
+                  )
+                }}
+              />
+            </p>
+            <p>{t('publish_collection_modal.tos_second_condition')}</p>
+            <p>{t('publish_collection_modal.tos_third_condition')}</p>
+          </div>
+          <Field
```
nicosantangelo, line 214: I think we could do two things to improve the UX of this field:
- We can delete the email validation and just use a Field with type email
- We could add a Form (from dcl-ui) so that pressing Enter (return key) on the field submits the form

if any of these two things become a pain to do, the current implementation is fine!

nicosantangelo (changes requested):
Mini commentuss, just one thing that appears to be a bug

src/components/CollectionDetailPage/CollectionDetailPage.tsx
LautaroPetaccio: This is a bug, great catch!

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
LautaroPetaccio: Done!

nicosantangelo (approved):

PR #1509, opened 2021-08-13
feat: Add 'Time since published' to the curators page

LautaroPetaccio:
This PR adds a new column to the collections table in the curators page to show the time difference between the actual time and when the collection was published.
Closes #1509


nicosantangelo (commented):
Works great!

Just one comment, to avoid having two lines (at least in the English language) what do you think of changing the text from

image

to

image

cazala (approved):

PR #1515, opened 2021-08-17
feat: Add search bar in curators page

### LINKED ISSUE #1496: Add search collections by name at the curation list
Like the one we have in the marketplace

LautaroPetaccio:
This PR adds the following:
- A search bar to filter collections based on their name or their owner's address.
- A new filter option to filter collection that have not been rejected or approved.

curators-search

Closes #1496


src/modules/translation/languages/en.json
```diff
@@ -894,9 +894,11 @@
   "curation_page": {
     "filter": {
       "all_status": "All status",
+      "not_reviewed": "Not reviewed",
       "approved": "Approved",
       "rejected": "Rejected"
     },
+    "search_placeholder": "Search between {count} results by name or owner address...",
```
nicosantangelo, line 901: we can (marginally) improve this translation by adding plural like the key below!
`{count, plural, one {result} other {results}}`

we can skip it for zh

nicosantangelo (commented):
Mini comment, but I  have a question. Without looking at the code, is there an easy way to search thru names?

For example if I look at my curation list I see this:
image

and there's no way for me to see addresses, I (correctly) see profile names

LautaroPetaccio:
The problem is that the name that is displayed under `Owner` is rendered using the `Profile` component which sends a request to get the profile of the address and renders the name. This is done exclusively when the component is rendered and, as we have (logical) pagination, we won't be able to query those name beforehand.
We can choose to do one of the following:
1. Get the profile for each of the owners of the collections (this might not scale properly as we should make the user wait for all the requests to be completed. We can have a single request to get all the profiles though, but the user will still need to wait).
2. Remove the capability to search by owner name.
3. Keep it as it is right now.
What do you think that will be better?

nicosantangelo:
Oh bummer, understood, I think the current implementation is good tho! It was just an idea for an enhancement if the data was already in the state

If you wanna take a look at https://github.com/decentraland/builder/pull/1515#discussion_r692274274 we can approve -> merge

nicosantangelo (approved):

PR #1516, opened 2021-08-18
feat: Add blockchain id to the items page

### LINKED ISSUE #1508: Show item_blockchain_id at the item detail once the collection and the item is in the catalyst
We should show it as `Id`

LautaroPetaccio:
This PR adds the blockchain id to the items page.

Screen Shot 2021-08-18 at 13 37 29

Closes #1508

### DIFF
--- a/src/modules/translation/languages/es.json
+++ b/src/modules/translation/languages/es.json
@@ -963,6 +963,7 @@
     "open_in_editor": "Abrir en el editor"
   },
   "item": {
+    "blockchain_id": "ID en la Blockchain",
     "category": "Categoría",
     "rarity": "Rareza",
     "representation": "Representación",
--- a/src/modules/translation/languages/es.json
+++ b/src/modules/translation/languages/es.json
@@ -963,7 +963,7 @@
     "open_in_editor": "Abrir en el editor"
   },
   "item": {
-    "blockchain_id": "ID en la Blockchain",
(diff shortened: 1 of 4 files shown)


src/modules/translation/languages/es.json
```diff
@@ -963,6 +963,7 @@
     "open_in_editor": "Abrir en el editor"
   },
   "item": {
+    "blockchain_id": "ID en la Blockchain",
```
nachomazzara, line 966: Im not 100% sure. Maybe we can use blockchain id for this too. Wdyt?

nachomazzara (approved):
Just an open question 🚀🚀

src/modules/translation/languages/es.json
LautaroPetaccio: Both work for me, but I used an "almost" real representation of the phrase in Spanish thinking that it would match the language.

nachomazzara: a real reprensentation may be then `Id en la cadena de bloques`

LautaroPetaccio: That's the correct translation, but I used the words `ID` and `Blockchain` that are fairly known and identify concepts in both languages mixed with Spanish so the phrase wasn't entirely in english, that's why the "almost real" representation.
I changed it to `Blockchain ID` though.

PR #1518, opened 2021-08-20
fix: Import model with the same name

### LINKED ISSUE #1488: Allow using the same .glb name for the female and male representation
If you add a representation with the same `.glb` name. The first representation added gets overrided

LautaroPetaccio:
This PR adds a mechanism to prefix the contents so they don't get overwritten by having the same name. This was common if one of the representations (male/female) was added after the other one and had the same name.
The prefix uses the selected body-shape.

Closes #1488

### DIFF
--- a/src/components/Modals/CreateItemModal/CreateItemModal.tsx
+++ b/src/components/Modals/CreateItemModal/CreateItemModal.tsx
@@ -97,6 +97,16 @@ export default class CreateItemModal extends React.PureComponent<Props, State> {
   }
 
   /**
+   * Prefixes the content name by adding the adding the body shape name to it.
+   *
+   * @param bodyShape - The body shaped used to prefix the content name.
+   * @param contentKey - The content key or name to be prefixed.
+   */
+  prefixContentName(bodyShape: BodyShapeType, contentKey: string): string {
+    return `${bodyShape}/${contentKey}`
+  }
+
+  /**
    * Creates a new contents record with the names of the contents blobs record prefixed.
    * The names need to be prefixed so they won't collide with other
    * pre-uploaded models. The name of the content is the name of the uploaded file.
@@ -106,7 +116,7 @@ export default class CreateItemModal extends React.PureComponent<Props, State> {
    */
   prefixContents(bodyShape: BodyShapeType, contents: Record<string, Blob>): Record<string, Blob> {
     return Object.keys(contents).reduce((newContents: Record<string, Blob>, key: string) => {
-      newContents[`${bodyShape}/${key}`] = contents[key]
+      newContents[this.prefixContentName(bodyShape, key)] = contents[key]
       return newContents
     }, {})
   }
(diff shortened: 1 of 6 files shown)


src/components/Modals/CreateItemModal/CreateItemModal.tsx
```diff
@@ -96,14 +96,46 @@ export default class CreateItemModal extends React.PureComponent<Props, State> {
...
+   * Creates a new contents record with the names of the contents blobs record prefixed.
+   * The names need to be prefixed so they won't collide with other
+   * pre-uploaded models. The name of the content is the name of the uploaded file.
+   *
+   * @param bodyShape - The body shaped used to prefix the content names.
+   * @param contents - The contents which keys are going to be prefixed.
+   */
+  prefixContents(bodyShape: BodyShapeType, contents: Record<string, Blob>): Record<string, Blob> {
+    return Object.keys(contents).reduce((newContents: Record<string, Blob>, key: string) => {
+      newContents[`${bodyShape}:${key}`] = contents[key]
```
cazala, line 109: let's use a `/` instead of a `:` to namespace these, so it's a valid path (would be handy if we needed to convert the contents into zip files, as we do with scenes or assetpacks.

cazala (changes requested):

cazala:
I found a few issues testing this PR:

1) I created a new Item by using a `m_pride_hat.glb` file.
- The Blobs look OK (`payload.contents`) ✅ 
- The mappings look OK (`payload.item.contents`) ✅ 
- The representation's `contents` look OK ✅ 
- The representation's `mainFile` is missing the namespace 🚫 

Screen Shot 2021-08-23 at 5 27 19 PM

2) I went to the Item detail, clicked on `Add female representation`, dropped the same `m_pride_hat.glb` and confirmed.
- The Blobs look OK (only the new female representation blobs are included to be uploaded) ✅ 
- The mappings seem to be missing the newly added female representation's mappings 🚫 
- The previous representation still has the right `contents` and the wrong `mainFile`, but the new one seems to have no` contents` at all 🚫 

Screen Shot 2021-08-23 at 5 39 16 PM

src/components/Modals/CreateItemModal/CreateItemModal.tsx
```diff
@@ -472,6 +477,7 @@ export default class CreateItemModal extends React.PureComponent<Props, State> {
     const { model, contents } = this.state
 
     const isCustom = !!contents && THUMBNAIL_PATH in contents
+    // TODO: Only executes if custom? why?
```
cazala, line 480: Only executes if NOT custom. That means that if the users provides a custom thumbnail (either a `thumbnail.png` file in the zip file or uploading a custom one via de edit icon in the UI) we use that one instead of generatin a new one from the model.

cazala (approved):

PR #1519, opened 2021-08-23
fix: Copy button in Share Modal

### LINKED ISSUE #1483: Copy a scene doesnt work
The copy button does nothing, marking, right-click, copy with mouse does nothing... And ctrl+c doesnt copy the URL either..

LautaroPetaccio:
This PR fixes the copy button by using the `CopyToClipboard` component. It also adds a simple border to the button so it doesn't merge with the input next to it.
copy-share

Closes #1483

### DIFF
--- a/src/components/Modals/ShareModal/ShareModal.tsx
+++ b/src/components/Modals/ShareModal/ShareModal.tsx
@@ -38,6 +38,13 @@ export default class ShareModal extends React.PureComponent<Props, State> {
     }
   }
 
+  componentWillUnmount() {
+    const { copiedTimer } = this.state
+    if (copiedTimer !== undefined) {
+      clearTimeout(copiedTimer)
+    }
+  }
+
   handleClose = () => {
     return this.props.onClose()
   }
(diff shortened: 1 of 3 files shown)


src/components/Modals/ShareModal/ShareModal.tsx
```diff
@@ -46,19 +48,11 @@ export default class ShareModal extends React.PureComponent<Props, State> {
...
-      if (selection) {
-        selection.removeAllRanges()
-        input.blur()
-        this.setState({ copied: true })
-      }
-      this.props.onShare(ShareTarget.LINK)
+  handleCopyLink = () => {
+    const { copiedTimer } = this.state
+    if (copiedTimer === undefined) {
+      const newCopiedTimer = setTimeout(() => this.setState({ copied: false, copiedTimer: undefined }), 2000)
```
cazala, line 54: We should check the component is still mounted by the time the `setTimeout` is executed, otherwise the `this.setState` will complain. We can do that by setting a flag at the instance level when the component mounts (and unsetting it when it unmounts) and checking against it within the `setTimeout` callback.

cazala (changes requested):

cazala (approved):

PR #1527, opened 2021-08-30
Add max supply to rarity options in CreateItemModel

### LINKED ISSUE #1456: Add item rarity amount (x) to dropdown on creation
Show each rarity limit when picking on the CreateItemModal.

i.e: Mythic (10).

May remove this [utils method](https://github.com/decentraland/builder/blob/d19c2cf4e910e1dca38c1c7e947db284fab6d9aa/src/modules/item/utils.ts#L30) in favor of the one at [dcl-schema](https://github.com/decentraland/common-schemas/blob/main/src/dapps/rarity.ts#L51)

P.S: Maybe we can do the same for rarity colors

fzavalia:
Closes #1456 

image

### DIFF
--- a/src/modules/item/utils.ts
+++ b/src/modules/item/utils.ts
@@ -28,7 +28,7 @@ export const MAX_FILE_SIZE = 2097152 // 2MB
 export const MAX_NFTS_PER_MINT = 50
 
 export function getMaxSupply(item: Item) {
-  return RARITY_MAX_SUPPLY[item.rarity!]
+  return getMaxSupplyForRarity(item.rarity!)
 }
 
 export function getMaxSupplyForRarity(rarity: ItemRarity) {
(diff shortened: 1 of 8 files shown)


src/modules/item/utils.ts
```diff
@@ -31,6 +31,10 @@ export function getMaxSupply(item: Item) {
   return RARITY_MAX_SUPPLY[item.rarity!]
 }
 
+export function getMaxSupplyForRarity(rarity: ItemRarity) {
```
nicosantangelo, line 34: This method is great! but I think we can refactor [one above](https://github.com/decentraland/builder/pull/1527/files#diff-6e44a8ab5281e24e1251f5bee8340d324289f9b25dfbf8f77aba93077b4c71b5R30) to avoid repetition right?

nicosantangelo (commented):
Mini comment but great work!

src/modules/item/utils.ts
fzavalia: done!

nicosantangelo (approved):
LGTM

Issue #1517, opened 2021-08-18
I shouldn't be able to delete published collections from my list


nicosantangelo:
We should hide the option when the collection is published ![image.png](https://images.zenhubusercontent.com/601dcbd9715e60c791100933/35839e86-8e0a-4e7d-bcf5-ca789c114fe7)

PR #1528, opened 2021-08-30
Don't show delete option on published collections

fzavalia:
Closing #1517 

As it is the only option for collections, I decided to entirely remove the dropdown in this case so there is not an empty list of options.

image


nicosantangelo (approved):
LGTM, we could re-enable the dropdown if we ever add new actions

PR #1529, opened 2021-08-31
Fix: When asset pack creation failed, the modal could not be closed

### LINKED ISSUE #1511: Once error occurs in CreateAssetModal, it can't be closed
Once error occurs in CreateAssetModal, it can't be closed

fzavalia:
Closes #1511 

This was because on fail, the close button looped between different states of the modal view, going back to the same one every time.

### DIFF
--- a/src/components/Modals/CreateAssetPackModal/CreateAssetPackModal.tsx
+++ b/src/components/Modals/CreateAssetPackModal/CreateAssetPackModal.tsx
@@ -63,7 +63,7 @@ export default class CreateAssetPackModal extends React.PureComponent<Props, Sta
       view = CreateAssetPackView.SUCCESS
     } else if (progress.stage !== ProgressStage.NONE && !error) {
       view = CreateAssetPackView.PROGRESS
-    } else if (error) {
+    } else if (view !== CreateAssetPackView.EXIT && error) {
       view = CreateAssetPackView.EDIT_ASSET_PACK
     }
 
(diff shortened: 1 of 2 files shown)


src/components/Modals/CreateAssetPackModal/CreateAssetPackModal.tsx
```diff
@@ -63,7 +63,7 @@ export default class CreateAssetPackModal extends React.PureComponent<Props, Sta
       view = CreateAssetPackView.SUCCESS
     } else if (progress.stage !== ProgressStage.NONE && !error) {
       view = CreateAssetPackView.PROGRESS
-    } else if (error) {
+    } else if (view !== CreateAssetPackView.EXIT && error) {
```
fzavalia, line 66: When clicking the close button the modal enters the exit state, showing the message to confirm you want to discard the asset package.
In the case of the error, it tried to go to the exit state but this function made it so that it went back to the edit asset pack state, making it seem like nothing happened

src/modules/assetPack/reducer.ts
```diff
@@ -104,7 +104,8 @@ export const assetPackReducer = (state = INITIAL_STATE, action: AssetPackReducer
     case SAVE_ASSET_PACK_FAILURE: {
       return {
         ...state,
-        error: action.payload.error
+        error: action.payload.error,
+        loading: loadingReducer(state.loading, action)
```
fzavalia, line 108: Not having this caused the loading state to never finish

nicosantangelo: Good catch here, don't know how it wasn't there before 😅

nicosantangelo (approved):
it worked! 🚀

PR #1530, opened 2021-08-31
Add message about using Ethereum MANA for name claiming

### LINKED ISSUE #1523: Explain that the MANA needed to claim a name is only the Ethereum MANA
When claiming a name, the text isn't clear about which type of currency is the user going to spend to claim a name. We currently have Ethereum MANA and Matic MANA and that distinction should be clearer.
While fixing this, add logic to disable the claim button when there's an on-going transaction to claim a name.

Screenshot about the claim button:

fzavalia:
Closes #1523 

image


nachomazzara (approved):
LGTM

PR #1534, opened 2021-09-01
feat: Show modal when unlinking a name

### LINKED ISSUE #1216: Add a confirm modal when de-assigning a name.
Show a confirmation modal after prompt for the transaction explaining what is going to happen.

Also, the animation on hover shows the `-` circle first.

Screen Shot 2020-12-17 at 12 33 27

fzavalia:
Closes #1216 

image

Also redirects to `/activity` when done.

Added quick fix to `set ens resolver/content` in tracking because it was breaking and disrupting the flow.


src/modules/ens/sagas.ts
```diff
@@ -258,6 +258,10 @@ function* handleSetENSContentRequest(action: SetENSContentRequestAction) {
     )
 
     yield put(setENSContentSuccess(ens, content, land, from.toString(), wallet.chainId, txHash))
+
+    if (!land) {
+      yield put(replace(locations.activity()))
```
fzavalia, line 263: Redirect to activity only when unlinking the name from the land. Only on this case because when linking a name, the flow continues on the same page and would be affected by the redirect

src/modules/analytics/track.ts
```diff
@@ -300,8 +299,10 @@ addPayload(CREATE_COLLECTION_FORUM_POST_FAILURE, 'Create forum post error')
 
 // ENS analytics
 add(SET_ENS_RESOLVER_SUCCESS, 'Set ENS Resolver', action => {
-  const { payload } = action as FetchTransactionSuccessAction
-  const { ens, resolver, address } = payload.transaction.payload
+  // TODO: Use correct Type
+  const { payload } = action as any
```
fzavalia, line 303: There seems to have been changes in the object schema and transaction does not exist inside payload anymore. This caused a blocking error on the flow. This is a simple patch but should be investigated more

src/modules/analytics/track.ts
```diff
@@ -77,6 +76,7 @@ import {
   ClaimNameSuccessAction
 } from 'modules/ens/actions'
 import { CREATE_COLLECTION_FORUM_POST_FAILURE, CREATE_COLLECTION_FORUM_POST_SUCCESS } from 'modules/forum/actions'
+import { getTransactionFromAction } from 'decentraland-dapps/dist/modules/transaction/utils'
```
nicosantangelo, line 79: Could we move this to the top of the file! as per the [standards](https://github.com/decentraland/standards/blob/master/standards/react-redux.md#5-imports-order-on-files)

src/components/LandDetailPage/UnsetENSContentModal.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Link } from 'react-router-dom'
+import { locations } from 'routing/locations'
+
+type UnsetENSContentModalProps = {
+  land: Land
+  ens: ENS
+  open: boolean
+  onCancel: () => void
+  onConfirm: () => void
+}
```
nicosantangelo, line 15: I think we can extract this to it's own [`.types.ts`](https://github.com/decentraland/standards/blob/master/standards/react-redux.md#4-component-types) file right?

src/components/LandDetailPage/UnsetENSContentModal.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <Modal.Actions>
+      <Button onClick={onCancel}>{t('global.cancel')}</Button>
+      <Button primary onClick={onConfirm}>
+        {t('global.proceed')}
+      </Button>
+    </Modal.Actions>
+  </Modal>
+)
+
+export default UnsetENSContentModal
```
nicosantangelo, line 35: I think we can use either `React.PureComponent` or `React.memo` here to memoize the component!

src/components/LandDetailPage/UnsetENSContentModal.tsx
```diff
@@ -0,0 +1,35 @@
...
+  land: Land
+  ens: ENS
+  open: boolean
+  onCancel: () => void
+  onConfirm: () => void
+}
+
+const UnsetENSContentModal = ({ land, ens, open, onCancel, onConfirm }: UnsetENSContentModalProps) => (
+  <Modal size="tiny" open={open}>
+    <Modal.Header>{t('land_detail_page.clear_ens_content.title')}</Modal.Header>
```
nicosantangelo, line 19: I'm nitpicking here but the component is named *Unset*ENSContent and the i18n key is  *clear*_ens_content
maybe we can use the same word?

src/components/LandDetailPage/ENSChip/ENSChip.tsx
```diff
@@ -2,40 +2,63 @@ import * as React from 'react'
...
+          }
+          isActive={isLoading}
+          icon={isLoading ? '' : 'minus'}
+          onClick={isLoading ? this.handleOnClick : undefined}
+          onIconClick={this.handleOnIconClick}
+        />
+        <UnsetENSContentModal
+          land={land}
+          ens={ens}
+          open={this.state.showConfirmationModal}
```
nicosantangelo, line 57: Not very important but tend to destruct our props and states at the beginning of the file like we do with [props here](https://github.com/decentraland/builder/pull/1534/files#diff-46484747eefd33ff47ffc0413c52abb5fb52ffb19a704c972c7b5515467c7423R33), maybe we can move this to `const { showConfirmationModal } = this.state` ?

src/components/LandDetailPage/UnsetENSContentModal.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import { T, t } from 'decentraland-dapps/dist/modules/translation/utils'
+import { Button, Modal } from 'decentraland-ui'
+import { ENS } from 'modules/ens/types'
+import { Land } from 'modules/land/types'
+import { Link } from 'react-router-dom'
+import { locations } from 'routing/locations'
+
+type UnsetENSContentModalProps = {
+  land: Land
```
nicosantangelo, line 10: I'll leave this to @cazala but maybe it's a good idea to connect this component to te state instead of passing down the [prop here](https://github.com/decentraland/builder/pull/1534/files#diff-bd96a13dbca6fcb2f285a65246189d82dd9cff7d5ff348090283b64f7dcaa64aR8)

I say this because ENSChip needing a Land prop to work is a bit weird

nicosantangelo (dismissed):
A few comments but great work!!

src/modules/analytics/track.ts
fzavalia: Used an util function from dapps to solve this

src/modules/ens/sagas.ts
cazala: Could we move this to the `modules/location/saga.ts` saga? So we don't create a dependency on the location module within this one. I noticed that there are already some imports from the ens module to the modal and location modules, but those can be fixed on another PR. What I mean by moving it to the location saga is taking the `SET_ENS_CONTENT_SUCCESS` from there and doing the redirect on that saga, like this:

```ts
export function* locationSaga() {
  yield all([
    /* other sagas... */
    takeLatest(SET_ENS_CONTENT_SUCCESS, handleSetENSContentSuccess),
  ])
}

function* handleSetENSContentSuccess(action: SetENSContnetSuccessAction) {
  const { land } = action.payload
  if (!land) {
    yield put(replace(locations.activity()))
  }
}
```

src/components/LandDetailPage/ENSChip/ENSChip.tsx
fzavalia: Removed the state in the last change as the new modal has been connected to the store and is opened via openModal

src/components/LandDetailPage/UnsetENSContentModal.tsx
fzavalia: Right, It is now connected to the store and decoupled from the chip, land is no longer provided

src/components/LandDetailPage/UnsetENSContentModal.tsx
fzavalia: Updated

src/components/LandDetailPage/UnsetENSContentModal.tsx
fzavalia: Went for PureComponent as all other modals were using it

src/components/LandDetailPage/UnsetENSContentModal.tsx
fzavalia: Done

src/modules/analytics/track.ts
fzavalia: Done

src/components/LandDetailPage/ENSChip/ENSChip.container.ts
```diff
@@ -15,7 +14,6 @@ const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
 
 const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
   onNavigate: path => dispatch(push(path)),
-  onUnsetENSContent: ens => dispatch(setENSContentRequest(ens))
```
fzavalia, line 18: Moved to the UnsetENSContentModal

src/modules/ens/sagas.ts
fzavalia: Good idea.
In order to completely remove the locations module I had to do the same thing for the claim name sagas which did the same.

```ts
function* handleClaimNameRequest(action: ClaimNameRequestAction) {
  const { name } = action.payload
  try {
    const [wallet, eth]: [Wallet, Eth] = yield getWallet()
    const from = Address.fromString(wallet.address)

    const controllerContract = new DCLController(eth, Address.fromString(CONTROLLER_ADDRESS))
    const tx: SendTx<TransactionReceipt> = yield call(() => controllerContract.methods.register(name, from).send({ from }))
    const txHash: string = yield call(() => tx.getTxHash())

    const ens: ENS = {
      address: wallet.address,
      name: name,
      subdomain: getDomainFromName(name),
      resolver: Address.ZERO.toString(),
      content: Address.ZERO.toString()
    }
    yield put(claimNameSuccess(ens, name, wallet.address, wallet.chainId, txHash))
    yield put(closeModal('ClaimNameFatFingerModal'))
    yield put(push(locations.activity())) -> HERE
  } catch (error) {
    const ensError: ENSError = { message: error.message }
    yield put(claimNameFailure(ensError))
  }
}
```

fzavalia: Maybe it would also be nice to do the same for the closeModal dispatch. I've seen that there is a `modal/sagas.ts` that can be used

fzavalia:
Addressed the comments on the last commits

src/modules/location/sagas.ts
```diff
@@ -2,9 +2,14 @@ import { all, takeLatest, put, select } from 'redux-saga/effects'
...
 import { locations } from 'routing/locations'
 import { LOGIN_SUCCESS, LoginSuccessAction } from 'modules/identity/actions'
+import { CLAIM_NAME_SUCCESS, SetENSContentSuccessAction, SET_ENS_CONTENT_SUCCESS } from 'modules/ens/actions'
 
 export function* locationSaga() {
-  yield all([takeLatest(LOGIN_SUCCESS, handleLoginSuccess)])
+  yield all([
+    takeLatest(LOGIN_SUCCESS, handleLoginSuccess),
+    takeLatest(SET_ENS_CONTENT_SUCCESS, handleSetENSContentSuccess),
+    takeLatest(CLAIM_NAME_SUCCESS, goToActivity)
```
fzavalia, line 11: Didn't put both SET_ENS_CONTENT_SUCCESS and CLAIM_NAME_SUCCESS because the former does a validation first.
on the future we could have 
```ts
takeLatest([CLAIM_NAME_SUCCESS,...otherActions], goToActivity)
```
To redirect to activity

cazala (approved):

PR #1536, opened 2021-09-02
fix: Disable claim name confirm button when loading

### LINKED ISSUE #1532: Disable the claim confirm button when claiming a name and it is loading

fzavalia:
Closes #1532 

Sep-02-2021 12-11-55

Also fixes text in spanish on the ClaimNameFatFingerModal that was not being correctly replaced.

Screen Shot 2021-09-02 at 12 08 42

### DIFF
--- a/src/components/Modals/ClaimNameFatFingerModal/ClaimNameFatFingerModal.tsx
+++ b/src/components/Modals/ClaimNameFatFingerModal/ClaimNameFatFingerModal.tsx
@@ -53,7 +53,7 @@ export default class ClaimNameFatFingerModal extends React.PureComponent<Props, 
             <Button secondary onClick={this.handleClose} type="button">
               {t('global.cancel')}
             </Button>
-            <Button primary type="submit" disabled={areNamesDifferent} loading={isLoading}>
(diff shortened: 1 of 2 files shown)


src/components/Modals/ClaimNameFatFingerModal/ClaimNameFatFingerModal.tsx
```diff
@@ -53,7 +53,7 @@ export default class ClaimNameFatFingerModal extends React.PureComponent<Props,
             <Button secondary onClick={this.handleClose} type="button">
               {t('global.cancel')}
             </Button>
-            <Button primary type="submit" disabled={areNamesDifferent} loading={isLoading}>
+            <Button primary type="submit" disabled={areNamesDifferent || isLoading} loading={isLoading}>
```
fzavalia, line 56: Maybe the button should be implicitly disabled when the loading prop is true

LautaroPetaccio: That is not a bad idea at all, but we're using this button (that comes from semantic-react-ui) everywhere and we are going to be introducing a breaking change.

LautaroPetaccio (approved):

PR #1539, opened 2021-09-06
fix: Don't use cache when fetching contents

LautaroPetaccio:
Chrome seems to cache responses that come from non-fetched requests as the same responses that come from fetched requests. This is a big issue because non-fetched requests don't require CORS, but fetched requests might. This cache strategy caused an issue because, when trying to add a new representation of an existing item, the content got cached without CORS headers when we loaded it a preview image before doing the fetch, resulting in a fetch request without headers and a CORS request failing.

This PR adds headers to avoid using cache when requesting the older models when adding new representations to an item.

### DIFF
--- a/src/modules/item/export.ts
+++ b/src/modules/item/export.ts
@@ -65,7 +65,7 @@ export async function getFiles(contents: Record<string, string>) {
   const promises = Object.keys(contents).map(path => {
     const url = getContentsStorageUrl(contents[path])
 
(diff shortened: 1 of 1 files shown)


cazala:
Leaving this as a reference here: https://stackoverflow.com/questions/12648809/cors-policy-on-cached-image

cazala (approved):

PR #1541, opened 2021-09-08
fix: Fetch scene content without cache

LautaroPetaccio:
This PR adds headers to avoid cached requests when fetching content to deploy a scene.

### DIFF
--- a/src/modules/media/utils.ts
+++ b/src/modules/media/utils.ts
@@ -22,7 +22,7 @@ export function dataURLToBlob(dataUrl: string): Blob | null {
 }
 
 export async function objectURLToBlob(objectURL: string): Promise<Blob> {
-  return fetch(objectURL).then(res => res.blob())
+  return fetch(objectURL, { headers: { pragma: 'no-cache', 'cache-control': 'no-cache' } }).then(res => res.blob())
 }
 
 export async function blobToCID(blob: Blob, path: string) {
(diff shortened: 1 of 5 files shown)


src/modules/media/utils.ts
```diff
@@ -22,7 +22,7 @@ export function dataURLToBlob(dataUrl: string): Blob | null {
 }
 
 export async function objectURLToBlob(objectURL: string): Promise<Blob> {
-  return fetch(objectURL).then(res => res.blob())
+  return fetch(objectURL, { headers: { pragma: 'no-cache', 'cache-control': 'no-cache' } }).then(res => res.blob())
```
fzavalia, line 25: Maybe abstract the headers into a separate object/function to avoid repetition

fzavalia:
```ts
const noCacheHeaders = { pragma: 'no-cache', 'cache-control': 'no-cache' }
```

fzavalia: or even wrap the fetch function
```ts
const noCacheFetch = (url: Parameters<typeof fetch>[0], init: Parameters<typeof fetch>[1] = {}) =>
  fetch(url, { ...init, headers: { ...init?.headers, pragma: 'no-cache', 'cache-control': 'no-cache' } })
```
But this might be overkill

nachomazzara (approved):
LGTM

PR #1552, opened 2021-09-10
fix: Fix BuilderAPI cyclic dependencies with the sagas

LautaroPetaccio:
This PR fixes a big cyclic dependency where the BuilderAPI module imported the store that later imported all the sagas. This meant that any sagas that used the BuilderAPI was cyclically linked.
The fix was implemented by instantiating the sagas using an instance of the BuilderAPI, generated when the store is generated.


src/modules/item/actions.ts
```diff
@@ -131,3 +131,17 @@ export const deployItemContentsFailure = (collection: Collection, item: Item, er
...
+
+// Fetch rarities
+
+export const FETCH_RARITIES_REQUEST = '[Request] Fetch Rarities'
+export const FETCH_RARITIES_SUCCESS = '[Success] Fetch Rarities'
+export const FETCH_RARITIES_FAILURE = '[Failure] Fetch Rarities'
+
+export const fetchRaritiesRequest = () => action(FETCH_RARITIES_REQUEST)
+export const fetchRaritiesSuccess = (rarities: Rarity[]) => action(FETCH_RARITIES_SUCCESS, { rarities })
+export const fetchRaritiesFailure = (error: string) => action(FETCH_RARITIES_FAILURE, { error })
```
cazala, line 143: What are these `fetchRarity` actions for? All the rarities are already present in the `Rarity` type from `@dcl/schemas`. Also I could not find a single dispatch of the `fetchRaritiesSuccess` and `fetchRaritiesFailure` in this PR.

LautaroPetaccio: The main idea of this change is to replace the cyclic dependency by parametrizing the BuilderAPI in all sagas. This implied that this request, the only one done without a sagas, needed to be done in a sagas, so I created those actions to accomplish that.
The `fetchRaritiesRequest` action is used [here](https://github.com/decentraland/builder/pull/1552/files#diff-a98cd061179e909ac4bd9302ca5c7ef9476b0bbc6bebdad2afb8c74ed670ba0cR27) and replaces the fetch done [here](https://github.com/decentraland/builder/pull/1552/files#diff-9088b58a90f358c685624ad30bde921a26304e094bd9fe43ca4cef8007a82a33L25) that didn't use sagas.
The success and failure actions are used [here](https://github.com/decentraland/builder/pull/1552/files#diff-39d0126acb3863393af8ab7fd948f9b4ebc17e8a74bac80898b6c1d7793d9052R96)

PR #1554, opened 2021-09-10
chore: Solve model types and project constants cyclic dependencies

LautaroPetaccio:
This PR solves two cyclic dependencies, one where the `ModelMetrics`, `Vector3` and `Quaternion` that seem to be exclusive of 3D models now have a types file in the `modules/models` directory.


src/modules/models/types.ts
```diff
@@ -0,0 +1,13 @@
...
+  triangles: number
+  materials: number
+  meshes: number
+  bodies: number
+  entities: number
+  textures: number
+}
+
+// TODO import Vector3 and Quaternion from a library ??? There are multiple implementations of it (decentraland-ecs, babylon, three)
+export type Vector3 = { x: number; y: number; z: number }
```
LautaroPetaccio, line 11: What about all the different Vector3 and Quaternion types that are lying around?

cazala: I _think_ we can import those from `decentraland-ecs`

cazala (approved):

src/modules/models/types.ts
LautaroPetaccio: After further analysis, it seems that the types that come from `decentraland-ecs` define classes, so it seems that we either stick to these values or change wherever we use them to create instances of those clases.

PR #1555, opened 2021-09-10
fix: Cyclic dependency of collections to items

LautaroPetaccio:
This PR removes a cyclic dependency between collections to items.


src/modules/item/utils.ts
```diff
@@ -308,3 +311,16 @@ export function isValidText(text: string) {
 export function isItemSizeError(error: string) {
   return error.search('The deployment is too big. The maximum allowed size per pointer is') !== -1
 }
+
+export function toInitializeItems(items: Item[]): InitializeItem[] {
+  return items.sort(sortByCreatedAt).map(toInitializeItem)
+}
+
+export function toInitializeItem(item: Item): InitializeItem {
```
cazala, line 319: This has been changed by https://github.com/decentraland/builder/pull/1556 I would wait for that one to be merged before merging this one.

This is the change: https://github.com/decentraland/builder/pull/1556/files#diff-294880f7daea062fae2add927da5e496ac51f396bd89cd960673f9da05e3912cR61

PR #1559, opened 2021-09-14
chore: Add tests and CI

LautaroPetaccio:
This PR adds the Jest configs, some tests and the CI process to run tests and report coverage.
The CI only runs the tests because Vercel already builds and deploys the site.

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -1,0 +1,22 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  Build-and-test-webapp:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Use Node.js 14.x
+        uses: actions/setup-node@v1
+        with:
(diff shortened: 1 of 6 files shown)


.github/workflows/node.yml
```diff
@@ -0,0 +1,22 @@
...
+          node-version: 14.x
+      - name: Install
+        run: npm ci
+      - name: Test
+        run: npm run test:coverage
+        if: ${{ always() }}
+      - name: report coverage
+        uses: coverallsapp/github-action@master
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
fzavalia, line 22: It would be nice to also to run the build

fzavalia:
Good job on the tests!

.github/workflows/node.yml
LautaroPetaccio: I left the build out of the Github Actions because we're already building an deploying it in Vercel

cazala (approved):

PR #1567, opened 2021-09-17
fix: Mint message showing an incorrect amount of minted items

LautaroPetaccio:
This PR fixes the transaction message when the mint was/is minted. The amount was computed from the length of the mints, but each mint has an amount of times it should be minted.

This PR also fixes a small issue where the same item could be added twice to be minted.

### DIFF
--- a/src/modules/collection/utils.ts
+++ b/src/modules/collection/utils.ts
@@ -5,7 +5,7 @@ import { getChainIdByNetwork } from 'decentraland-dapps/dist/lib/eth'
 import { Wallet } from 'decentraland-dapps/dist/modules/wallet/types'
 import { Item } from 'modules/item/types'
 import { isEqual, includes } from 'lib/address'
-import { Collection, Access } from './types'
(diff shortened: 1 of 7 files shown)


src/modules/collection/utils.ts
```diff
@@ -88,3 +88,9 @@ export function canManageCollectionItems(collection: Collection, address?: strin
 export function hasReviews(collection: Collection) {
   return collection.reviewedAt !== collection.createdAt
 }
+
+export function getTotalAmountOfMintedItems(mints: Mint[]) {
+  return mints.reduce((total, mint) => {
```
nachomazzara, line 93:
```suggestion
  return mints.reduce((total, mint) => total += mint.amount, 0)
```

nachomazzara, line 93: Maybe we can do

```typescript
  return mints.reduce((total, mint) => total += mint.amount, 0)
```

nachomazzara (approved):
LGTM, just a minor comment

PR #1569, opened 2021-09-20
feat: save published items

### LINKED ISSUE #1544: Allow users to save changes to published items
- Change `handleSavePublishedItemRequest` to dispatch a `saveItemRequest` instead of a  `deployItemRequest`

- If metadata changed do not send a transaction, save item in builder server (the metadata will be saved by the reviewer at the end using rescueItem)

- Change `handleSaveItemRequest` saga to stop checking if the item is published

- Unblock UI of published item (except for item rarity and collection name)

cazala:
Closes #1544

### DIFF
--- a/src/components/Modals/CreateItemModal/CreateItemModal.tsx
+++ b/src/components/Modals/CreateItemModal/CreateItemModal.tsx
@@ -665,7 +665,14 @@ export default class CreateItemModal extends React.PureComponent<Props, State> {
             label={t('create_item_modal.rarity_label')}
             placeholder={t('create_item_modal.rarity_placeholder')}
             value={rarity}
-            options={rarities.map(value => ({ value, text: `${t(`wearable.rarity.${value}`)} (${getMaxSupplyForRarity(value)})` }))}
+            options={rarities.map(value => ({
+              value,
+              label: t(`wearable.supply`, {
+                count: getMaxSupplyForRarity(value),
+                formatted: getMaxSupplyForRarity(value).toLocaleString()
+              }),
+              text: t(`wearable.rarity.${value}`)
+            }))}
             onChange={this.handleRarityChange}
           />
         ) : null}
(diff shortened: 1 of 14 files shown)


src/components/Modals/CreateItemModal/CreateItemModal.tsx
```diff
@@ -746,43 +753,43 @@ export default class CreateItemModal extends React.PureComponent<Props, State> {
...
-                                onChange={this.handleItemChange}
-                                isDisabled={isAddingRepresentation}
-                              />
-                            </Section>
-                          ) : (
-                            this.renderFields()
-                          )}
-                        </>
-                      )}
+                          <>
```
fzavalia, line 756: I think this new whitespaces should not be here. (Locally, Prettier removes them for me)

src/modules/item/sagas.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+        .run({ silenceTimeout: true })
+    })
+
+    it('should throw if file size is larger than 2 MB', () => {
+      const item = {
+        name: 'valid name',
+        description: 'valid description'
+      } as Item
+      return expectSaga(itemSaga, builderAPI)
+        .provide([[matchers.call.fn(calculateFinalSize), Promise.resolve(MAX_FILE_SIZE + 1)]])
```
fzavalia, line 76: You could also use `.provide([[call(calculateFinalSize, expectedItem, expectedContent), Promise.resolve(MAX_FILE_SIZE + 1)]])` To also validate that the function is receiving the expected parameters

src/modules/item/selectors.ts
```diff
@@ -25,9 +25,9 @@ export const getWalletItems = createSelector<RootState, Item[], string | undefin
 )
 
 export const getAuthorizedItems = createSelector<RootState, Collection[], Item[], string | undefined, Item[]>(
-  getAuthorizedCollections,
-  getItems,
-  getAddress,
+  state => getAuthorizedCollections(state),
```
fzavalia, line 28: I'm curious about why it was changed to this

fzavalia (approved):
LGTM!
Comments are for some minor things.

src/modules/item/selectors.ts
cazala: This was due to circular deps between the selectors. Passing `getAuthorizedCollections` as reference there would be undefined, but calling it within that lambda works fine on runtime.

fzavalia (approved):

PR #1579, opened 2021-09-22
feat: Fetch default & created by the user asset packs

LautaroPetaccio:
This PR uses the new query parameter to fetch simultaneously the asset packs created by the user and the default ones.


nachomazzara (approved):
LGTM

cazala (approved):

PR #1580, opened 2021-09-24
feat: show status indicator for collection and items

### LINKED ISSUE #1545: Show status on Collection/Items
- Fetch and store in state hashes of deployed items

- Connect the "green orb" component to check if hashes match or not and show yellow or green color

cazala:
Closes #1545


src/components/ItemStatus/ItemStatus.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+import { Popup } from 'decentraland-ui'
+
+import { SyncStatus } from 'modules/item/types'
+import { Props } from './ItemStatus.types'
+import './ItemStatus.css'
+
+export default class ItemStatus extends React.PureComponent<Props> {
```
fzavalia, line 9: ItemStatus and CollectionStatus share the same css and code. They could be abstracted to a general Status component on which a `type: 'item' | 'collection'` prop is provided.
The container has a minimal difference in both, but it can be handled by using this prop from ownProps

src/components/Modals/CreateItemModal/CreateItemModal.tsx
```diff
@@ -748,43 +755,43 @@ export default class CreateItemModal extends React.PureComponent<Props, State> {
...
-                                onChange={this.handleItemChange}
-                                isDisabled={isAddingRepresentation}
-                              />
-                            </Section>
-                          ) : (
-                            this.renderFields()
-                          )}
-                        </>
-                      )}
+                          <>
```
fzavalia, line 758: We should check why the formatter is doing this.

src/components/CollectionDetailPage/CollectionDetailPage.tsx
```diff
@@ -126,7 +127,7 @@ export default class CollectionDetailPage extends React.PureComponent<Props, Sta
                 <Column className="header-column">
                   <Row className="header-row" onClick={this.handleEditName}>
                     <Header size="huge" className="name">
-                      {collection.name}
+                      {collection.isPublished ? <CollectionStatus collection={collection} /> : null}{collection.name}
```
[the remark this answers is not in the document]
fzavalia: This is just a personal preference thoe

src/components/ItemEditorPage/RightPanel/RightPanel.tsx
```diff
@@ -255,7 +254,7 @@ export default class RightPanel extends React.PureComponent<Props, State> {
       <div className="RightPanel">
         {isConnected ? (
           <ItemProvider id={selectedItemId}>
-            {(item, collection, isItemLoading) => {
+            {(item, _collection, isItemLoading) => {
```
fzavalia, line 257: Why the undescore?

src/modules/collection/selectors.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+import { RootState } from 'modules/common/types'
+import { SyncStatus } from 'modules/item/types'
+import { getStatusByCollectionId } from './selectors'
+
+jest.mock('decentraland-dapps/dist/lib/eth')
+const mockGetChainIdByNetwork = getChainIdByNetwork as jest.Mock
+
+describe('when getting status by item id', () => {
+  it('should return the status for each published item', () => {
+    mockGetChainIdByNetwork.mockReturnValue(ChainId.MATIC_MAINNET)
```
fzavalia, line 12: Lets use `.mockReturnValueOnce` just in case.

fzavalia: In this particular test is unnecessary, but if you were to add a new one with another `mockReturnValue`, the second would not work.

src/components/ItemStatus/ItemStatus.container.ts
```diff
@@ -0,0 +1,17 @@
...
+const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
+  const statusByItemId = getStatusByItemId(state)
+  return {
+    status: statusByItemId[ownProps.item.id]
+  }
+}
+
+const mapDispatch = (_dispatch: MapDispatch): MapDispatchProps => ({})
+
+export default connect(mapState, mapDispatch)(ItemStatus)
```
fzavalia, line 17: Instead of passing an empty mapDispatch you can always pass null or nothing at all.
```
export default connect(mapState)(ItemStatus)
```

src/components/ItemStatus/ItemStatus.container.ts
```diff
@@ -0,0 +1,17 @@
...
+import ItemStatus from './ItemStatus'
+
+const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
+  const statusByItemId = getStatusByItemId(state)
+  return {
+    status: statusByItemId[ownProps.item.id]
+  }
+}
+
+const mapDispatch = (_dispatch: MapDispatch): MapDispatchProps => ({})
```
fzavalia, line 15: The convention for unused variables is just an undescore `_`, no need to declare the name (undescore and name is the convention for private)

src/modules/collection/selectors.ts
```diff
@@ -50,3 +51,22 @@ export const isOnSaleLoading = createSelector<RootState, Transaction[], boolean>
...
+export const getStatusByCollectionId = createSelector<RootState, Item[], Record<string, SyncStatus>, Record<string, SyncStatus>>(
+  state => getItems(state),
+  state => getStatusByItemId(state),
+  (items, itemStatusByItemId) => {
+    const statusByCollectionId: Record<string, SyncStatus> = {}
+    for (const item of items) {
+      const { collectionId } = item
+      if (collectionId) {
+        if (collectionId in statusByCollectionId) {
+          statusByCollectionId[collectionId] = getMostRelevantStauts(statusByCollectionId[collectionId], itemStatusByItemId[item.id])
```
fzavalia, line 64: typo in `getMostRelevantStauts `

src/modules/entity/reducer.ts
```diff
@@ -0,0 +1,59 @@
...
+      }
+    }
+    case FETCH_ENTITIES_SUCCESS: {
+      return {
+        ...state,
+        loading: loadingReducer(state.loading, action),
+        error: null,
+        data: {
+          ...state.data,
+          ...action.payload.entities.reduce((obj, entity) => {
```
fzavalia, line 42: You can always one line this with ```
=> ({...obj, [entity.entityId]: entity })
```

Comment (fzavalia) [reply src/modules/entity/reducer.ts:42]: And even go the extra mile by doing 
```
data: action.payload.entities.reduce((obj, entity) => ({...obj, [entity.entityId]: entity }), state.data)
```

File: src/modules/item/selectors.spec.ts [hunk 10]
```diff
@@ -111,4 +116,126 @@ describe('Item selectors', () => {
       expect(getRarities(state)).toEqual(state.item.rarities)
     })
   })
+
+  describe('when getting status by item id', () => {
+    it('should return the status for each published item', () => {
+      mockGetChainIdByNetwork.mockReturnValue(ChainId.MATIC_MAINNET)
```
fzavalia, line 122: `.mockReturnValueOnce`

src/components/ItemEditorPage/RightPanel/RightPanel.tsx
fzavalia: If you mean to say that collection is an unused variable I recommend you just write an underscore without the name. just undescore means unused, undescore and name means private (in vanilla js)

src/modules/item/selectors.spec.ts
cazala: This one i had to change it back to the `mockReturnValue` because it's called more than once (three times in this test, once of each item).

src/modules/entity/reducer.ts
cazala: Regarding the usage of `state.data` as the initial value of the reduce, that only works when using the spread, but if I'd do that without using the spread within the reduce, it would be mutating

cazala: Oh wait haha it's not 23 seconds (i spread the wrong object 😅  ) it's 71 ms, but still way slower

cazala: 10k entries might sound like a lot, but we have reducer with 10x that (like the tiles on the map), I learned this the hard way once we added a spread in on reducer or selector that handled that amount of entries and it did freeze the UI for a couple seconds every time it got computed lol

src/components/ItemStatus/ItemStatus.container.ts
cazala: mm afaik typescript allows `_` if you don't use any variable but also `_param` if you need to select them individually, ie:

```ts
function (_paramA: any, _paramB: any, paramC: any) {
  // use only paramC
}
```

I think the `_param` is clearer and i think it's been used like that throughout the codebase

src/components/ItemEditorPage/RightPanel/RightPanel.tsx
cazala: in typescript the leading underscore does tell the linter to ignore that param for the unused param rule, without it the linter would complain. Just `_` does work too, but is more vague, and you can't use it if you need to apply it on a few params but not all.

src/components/ItemStatus/ItemStatus.container.ts
cazala: Yea to be honest I just leave them like that to make it easier to connect later on which is usually what happens with most components over time :P (that's why i create the types too even tho they are empty)

src/modules/collection/selectors.spec.ts
cazala: The test fails without it because it calls that mock 3 times

fzavalia (approved):

PR #1583, opened 2021-09-27
fix: export

cazala:
Fixes the export of scenes

### DIFF
--- a/src/modules/project/export.ts
+++ b/src/modules/project/export.ts
@@ -345,21 +345,20 @@ export async function downloadFiles(args: {
   }
 
   // Download models
-  total += Object.keys(mappings).length
+  const paths = Object.keys(mappings).filter(path => (isDeploy ? !path.endsWith('.ts') : !path.endsWith('.js')))
+  total += paths.length
   onProgress({ loaded: progress, total })
 
-  const promises = Object.keys(mappings)
-    .filter(path => (isDeploy ? !path.endsWith('.ts') : !path.endsWith('.js')))
-    .map(path => {
-      const url = mappings[path]
-      return fetch(url, { headers: NO_CACHE_HEADERS })
-        .then(resp => resp.blob())
-        .then(blob => {
-          progress++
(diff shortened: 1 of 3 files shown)


src/modules/project/export.ts
```diff
@@ -345,21 +345,20 @@ export async function downloadFiles(args: {
   }
 
   // Download models
-  total += Object.keys(mappings).length
+  const paths = Object.keys(mappings).filter(path => (isDeploy ? !path.endsWith('.ts') : !path.endsWith('.js')))
```
cazala, line 348: I moved this thingy up here to fix a minor issue where the progress sometimes would not reach 100% because of the mappings filtered out. With this line moved up here now the progress bar always reaches 100%.

src/modules/project/sagas.ts
```diff
@@ -205,13 +204,11 @@ export function* projectSaga(builder: BuilderAPI) {
...
       })
     )
-
     for (const filename of Object.keys(files)) {
       zip.file(filename, files[filename])
     }
 
-    const artifact = yield call(zip.generateAsync, { type: 'blob' })
-    saveAs(artifact, `${sanitizedName}.zip`)
+    yield call(download, zip, sanitizedName)
```
cazala, line 211: This was the line that was breaking the export. I have no idea why it suddenly stopped working.. and I dunno why if I wrap it in an async function it works again, but it does work now 🤷‍♂️ .

fzavalia: Just random guessing, but it might be because of something like this https://redux-saga.js.org/docs/api/#callcontext-fn-args

fzavalia: maybe the generateAsync is not being provided the correct `this` which might be fixed by using the call alternatives

cazala: good catch -- `this` was `null` and causing the function to blow up

fzavalia (approved):
LGTM!

PR #1584, opened 2021-09-29
fix: stats

cazala:
Use the data cdn to show stats. I'm perma caching in memory the promise since it only changes every 24hs. Also I'm caching the promise and not the result because in many cases there are several requests fired simultaneously and awaiting the result before caching would end up in many requests, this way there's only one.

This is a temporary solution until we have a server or something in place, but from my tests this is loading in a reasonable time (500ms, 400kb using brotli compression)

### DIFF
--- a/src/lib/api/builder.ts
+++ b/src/lib/api/builder.ts
@@ -648,10 +648,10 @@ export class BuilderAPI extends BaseAPI {
 
     const weeklyStats: WeeklyStats = {
       base,
-      users: stats ? stats.users : 0,
-      sessions: stats ? stats.sessions : 0,
-      medianSessionTime: stats ? stats.median_session_time : 0,
-      maxConcurrentUsers: stats ? stats.max_concurrent_users || 0 : 0
+      users: stats?.users ?? 0,
+      sessions: stats?.sessions ?? 0,
+      medianSessionTime: stats?.median_session_time ?? 0,
+      maxConcurrentUsers: stats?.max_concurrent_users ?? 0
     }
 
     return weeklyStats
--- a/src/lib/api/builder.ts
+++ b/src/lib/api/builder.ts
@@ -666,7 +666,28 @@ export class BuilderAPI extends BaseAPI {
   }
 
   async fetchWeeklyStats(base: string) {
-    const remoteStats: RemoteWeeklyStats = await this.request('get', `/analytics/weekly?base=${base}`)
+    const resp = await fetch('https://cdn-data.decentraland.org/scenes/scene-stats.json')
+    const json: Record<
+      string,
+      { last_7d: { users: number; sessions: number; median_session_time: number; max_concurrent_users: number | null } }
+    > = await resp.json()
+    const stats = base in json ? json[base]['last_7d'] : null
+    const remoteStats: RemoteWeeklyStats = {
+      week: '',
+      title: '',
+      base,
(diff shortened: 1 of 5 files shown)


src/lib/api/builder.ts
```diff
@@ -666,8 +640,21 @@ export class BuilderAPI extends BaseAPI {
...
+    }
+    const json = await statsPromise
+    const stats = base in json ? json[base]['last_7d'] : null
+
+    const weeklyStats: WeeklyStats = {
+      base,
+      users: stats ? stats.users : 0,
+      sessions: stats ? stats.sessions : 0,
+      medianSessionTime: stats ? stats.median_session_time : 0,
+      maxConcurrentUsers: stats ? stats.max_concurrent_users || 0 : 0
```
LautaroPetaccio, line 654: This is a suggestion, but this might look better with the optional operators:
```suggestion
      users: stats?.users ?? 0,
      sessions: stats?.sessions ?? 0,
      medianSessionTime: stats?.median_session_time ?? 0,
      maxConcurrentUsers: stats?.max_concurrent_users ?? 0
```

src/lib/api/builder.ts
```diff
@@ -666,8 +640,21 @@ export class BuilderAPI extends BaseAPI {
   }
 
   async fetchWeeklyStats(base: string) {
-    const remoteStats: RemoteWeeklyStats = await this.request('get', `/analytics/weekly?base=${base}`)
-    return fromRemoteWeeklyStats(remoteStats)
+    if (!statsPromise) {
+      statsPromise = fetch('https://cdn-data.decentraland.org/scenes/scene-stats.json').then(resp => resp.json())
```
LautaroPetaccio, line 644: Should we have this in the builder api if it doesn't query things from the builder?

src/lib/api/builder.ts
```diff
@@ -26,6 +26,13 @@ export const getContentsStorageUrl = (hash: string = '') => `${BUILDER_SERVER_UR
 export const getAssetPackStorageUrl = (hash: string = '') => `${BUILDER_SERVER_URL}/storage/assetPacks/${hash}`
 export const getPreviewUrl = (projectId: string) => `${BUILDER_SERVER_URL}/projects/${projectId}/media/preview.png`
 
+// caching the stats promise since it should not change for 24hs. Caching the promise instead of the awaited so simultaneous calls to the function would result in a single request.
```
LautaroPetaccio, line 29:
```suggestion
// caching the stats promise since it should not change for 24hs. Caching the promise instead of using await so simultaneous calls to the function would result in a single request.
```

src/lib/api/builder.ts
cazala: I moved this to an `analytics` client that later can be integrated against the server that would handle this

src/modules/stats/sagas.ts
```diff
@@ -8,13 +9,13 @@ import {
 } from './action'
 import { WeeklyStats } from './types'
 
-export function* statsSaga(builder: BuilderAPI) {
+export function* statsSaga(_builder: BuilderAPI) {
```
LautaroPetaccio, line 12: I think that we should remove this if we're not using the builder API here, what do you think?

LautaroPetaccio (changes requested):

LautaroPetaccio (approved):

PR #1664, opened 2021-11-10
feat: Build item slots

### LINKED ISSUE #1573: Buy Item slots
- Add the number of slots available. Add a tooltip saying that the slots are per the third party not from the collection)

- If the user has no item slots available, we should disable the `publish` button and add functionality to buy new ones

- The user should see a modal where they can select the number of item slots desired and the price based on the different tiers. This info will be fetched by the builder-server (the builder-server will use the subgraph). Therefore a new endpoint `/item/tiers` will be needed to get this info.

- Transactions: Approve MANA & Buy slots. The user will need to send a transaction to authorize the TPR smart contract to use the MANA on their behalf. And another transaction to actually buy the item slots.

LautaroPetaccio:
This PR adds the `BuyItemSlotsModal` with the required module to buy slots and attaches it to the `ThirdPartyCollectionDetailPage`.

Screen Shot 2021-11-12 at 14 54 00

Screen Shot 2021-11-12 at 17 42 59

Screen Shot 2021-11-10 at 14 41 29

Screen Shot 2021-11-12 at 14 52 45

Closes #1573


src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
```diff
@@ -39,14 +44,15 @@ export default class ThirdPartyCollectionDetailPage extends React.PureComponent<
   }
 
   hasAccess() {
-    const { wallet, collection } = this.props
-    return collection !== null && canSeeCollection(collection, wallet.address)
+    const { wallet, collection, thirdParty } = this.props
+    return collection && thirdParty && isUserManagerOfThirdParty(wallet.address, thirdParty)
```
LautaroPetaccio, line 48: This is enough to verify that the user has access to the collection. What do you think about this @nicosantangelo ?

src/lib/urn.ts
```diff
@@ -92,6 +100,11 @@ export function toLegacyURN(urn: URN): URN {
   return urn.replace('urn:decentraland:off-chain:base-avatars:', 'dcl://base-avatars/')
 }
 
+export function getThirdPartyId(urn: URN): string {
```
nicosantangelo, line 103: I'm not sure about the name of this method. I think the name can be a bit misleading because it's receiving an URN and returning the thirdPartyId, but it actually it's getting a third party URN and trying to extract the id from there.

Furthermore, the implementation could be reduced to:

```ts
decodeURN(collection.urn).suffix
```

Maybe we can just add a selector that does (and memoizes) this?

`getCollectionThirdParty(state, collection)` ?

src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.container.ts
```diff
@@ -9,21 +9,23 @@ import { getCollectionItems, getLoading as getLoadingItem } from 'modules/item/s
...
   return {
     wallet: getWallet(state)!,
     collection,
     items: getCollectionItems(state, collectionId),
     isLoading:
       isLoadingType(getLoadingCollection(state), FETCH_COLLECTIONS_REQUEST) ||
       isLoadingType(getLoadingCollection(state), DELETE_COLLECTION_REQUEST) ||
-      isLoadingType(getLoadingItem(state), FETCH_ITEMS_REQUEST)
+      isLoadingType(getLoadingItem(state), FETCH_ITEMS_REQUEST),
+    thirdParty: collection && collection.urn ? getThirdParty(state, getThirdPartyId(collection.urn)) : null
```
nicosantangelo, line 28: We usually try to keep selectors that do a direct access without a method. Still this is not necessarily bad thing, but wanted to add the comment to consider [this](https://github.com/decentraland/builder/pull/1664/files#diff-ee2a3e5bcc9f9f86387798ee5c181abbcbbdd0f52fa80a3c7a71983cbce70bfdR103)

src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
nicosantangelo: Yup I think it's ok!

src/modules/tiers/reducer.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const INITIAL_STATE: TiersState = {
+  data: {
+    thirdParty: []
+  },
+  loading: [],
+  error: null
+}
+
+type ThirdPartyReducerAction =
```
nicosantangelo, line 28: `TiersReducerAction` !

src/modules/tiers/reducer.ts
```diff
@@ -0,0 +1,80 @@
...
+  BuyThirdPartyItemTiersRequestAction,
+  BuyThirdPartyItemTiersSuccessAction,
+  FetchThirdPartyItemTiersFailureAction,
+  FetchThirdPartyItemTiersRequestAction,
+  FetchThirdPartyItemTiersSuccessAction
+} from './types'
+
+const INITIAL_STATE: TiersState = {
+  data: {
+    thirdParty: []
```
nicosantangelo, line 22: Just making sure, this property is here so we can, in the future, add tiers to other parts of the state right?

I ask mostly because it's called `thirdParty` but not `thirdPartyItem`s when the type is `ThirdPartyItemTier`

src/modules/tiers/selectors.ts
```diff
@@ -0,0 +1,33 @@
+import { LoadingState } from 'decentraland-dapps/dist/modules/loading/reducer'
+import { isLoadingType } from 'decentraland-dapps/dist/modules/loading/selectors'
+import { RootState } from 'modules/common/types'
+import { getPendingTransactions } from 'modules/transaction/selectors'
+import { FETCH_THIRD_PARTY_ITEM_TIERS_REQUEST, BUY_THIRD_PARTY_ITEM_TIERS_SUCCESS, BUY_THIRD_PARTY_ITEM_TIERS_REQUEST } from './actions'
+import { ThirdPartyItemTier, TiersState } from './types'
+
+function getState(state: RootState): TiersState {
```
nicosantangelo, line 8: I don't really mind this, but if you want, we could change this file to use constants
```ts
export const getState (...)
```

so it looks like all the other selectors in the codebase/other dapps.

Following that path, we can keep consistent and add the four selectors we have in all files:

```ts
export const getState = (state: RootState) => state.item
export const getData = (state: RootState) => getState(state).data
export const getLoading = (state: RootState) => getState(state).loading
export const getError = (state: RootState) => getState(state).error
```

which'd allow for not having to do `.data` below too

src/modules/wallet/selectors.ts
```diff
@@ -0,0 +1,8 @@
+import { RootState } from 'modules/common/types'
+import { getData as getWallet } from 'decentraland-dapps/dist/modules/wallet/selectors'
+
+export function getManaBalance(state: RootState): number {
+  const wallet = getWallet(state)
+
+  return wallet ? wallet.networks.MATIC.mana : 0
+}
```
nicosantangelo, line 8: Careful here, this method is misleading. There's no one mana balance, it always should depend on the network. Furthermore, decentraland dapps already has a `getNetworks` which can be accessed by network to get the mana balance (deprecating [this selector](https://github.com/decentraland/decentraland-dapps/blob/master/src/modules/wallet/selectors.ts#L36))
I think we can just use getNetworks where this selector is used and pass Netowrk.MATIC if necessary

tsconfig.json
```diff
@@ -11,13 +11,13 @@
...
     "noImplicitThis": true,
-    "noUnusedLocals": true,
+    "noUnusedLocals": false,
     "isolatedModules": true,
     "esModuleInterop": true,
     "strictNullChecks": true,
     "noImplicitReturns": true,
     "resolveJsonModule": true,
-    "noUnusedParameters": true,
+    "noUnusedParameters": false,
```
nicosantangelo, line 20: but-why

After seeing this I've opened `BuyItemSlotsModal.container.ts` and saw that Dispatch is imported and not being used :(

nicosantangelo (changes requested):
Things I've noticed using the UI:

*. The loading indicator is too close to the text, which makes it a bit jarring when the data eventually appears
Screen Shot 2021-11-11 at 14 17 43

*. The "you don't have enough MANA notice has more padding at the bottom that at the top which makes it a bit weird. Also, there seem to be two spaces between the MANA symbol and `MANA`

image

3. The same described above happens with the error message. Furthermore, picking a new slot does not clean the error and there's no `X` to do it either which can be confusing.
image
image

*. **MINOR or NOT necessary**: Maybe we can align the MANA symbols vertically?:
image

tsconfig.json
LautaroPetaccio: Ohh, I removed it to make the development a bit faster but I forgot about re-adding it again.

src/modules/tiers/selectors.ts
LautaroPetaccio: We're using lambdas in all of our other selectors, so I don't mind at all! Changed.

src/modules/tiers/reducer.ts
LautaroPetaccio: It is set like that for the purpose you've mentioned.
You're right, it should be called `thirdPartyItems` as it defines the tiers of the items in a more proper way.

src/modules/tiers/reducer.ts
LautaroPetaccio: Done!

src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
```diff
@@ -14,6 +15,7 @@ import CollectionPublishButton from './CollectionPublishButton'
 import { Props } from './ThirdPartyCollectionDetailPage.types'
 
 import './ThirdPartyCollectionDetailPage.css'
+import { isUserManagerOfThirdParty } from 'modules/thirdParty/utils'
```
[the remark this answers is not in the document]
LautaroPetaccio: Done!

src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
```diff
@@ -39,14 +44,15 @@ export default class ThirdPartyCollectionDetailPage extends React.PureComponent<
...
+    const { wallet, collection, thirdParty } = this.props
+    return collection && thirdParty && isUserManagerOfThirdParty(wallet.address, thirdParty)
   }
 
   renderPage() {
-    const { wallet, items } = this.props
+    const { wallet, items, thirdParty } = this.props
     const collection = this.props.collection!
-    const slots = 0
+    const slots = thirdParty ? new BN(thirdParty.maxItems).sub(new BN(thirdParty.totalItems)) : new BN(0)
```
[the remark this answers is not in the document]
LautaroPetaccio: Moved it and tested it.

src/components/Modals/BuyItemSlotsModal/BuyItemSlotsModal.types.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export type Props = ModalProps & {
+  isBuyingItemSlots: boolean
+  isFetchingTiers: boolean
+  thirdParty: ThirdParty
+  onBuyItemSlots: typeof buyThirdPartyItemTiersRequest
+  onFetchThirdPartyItemSlots: typeof fetchThirdPartyItemTiersRequest
+  manaBalance: number
+  tiers: ThirdPartyItemTier[]
+  metadata: { thirdParty: ThirdParty }
```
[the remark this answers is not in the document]
LautaroPetaccio: Moved the callbacks at the end and created the Metadata type!

src/components/Modals/BuyItemSlotsModal/BuyItemSlotsModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+import styles from './BuyItemSlotsModal.module.css'
+
+const sortTiers = (a: ThirdPartyItemTier, b: ThirdPartyItemTier) => {
+  if (a.value < b.value) {
+    return -1
+  } else if (a.value > b.value) {
+    return 1
+  }
+  return 0
+}
```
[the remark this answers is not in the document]
LautaroPetaccio: Moved it to `modules/tiers/utils`. I left it there because it was the only place that it was going to be used.
I also added tests for it.

src/components/Modals/BuyItemSlotsModal/BuyItemSlotsModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+import BN from 'bn.js'
+import { fromWei } from 'web3x/utils'
+import { Button, ModalDescription, ModalHeader, CheckboxProps, Radio, Mana, Loader, Message } from 'decentraland-ui'
+import { T, t } from 'decentraland-dapps/dist/modules/translation/utils'
+import { Modal, NetworkButton } from 'decentraland-dapps/dist/containers'
+import { env } from 'decentraland-commons'
+import { Props } from './BuyItemSlotsModal.types'
+import { Network } from '@dcl/schemas'
+import { ThirdPartyItemTier } from 'modules/tiers/types'
+import styles from './BuyItemSlotsModal.module.css'
```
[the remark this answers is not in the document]
LautaroPetaccio: Done!

src/modules/wallet/selectors.ts
LautaroPetaccio: I took this section of code from [here](https://github.com/decentraland/builder/blob/c5c012568d9cdb5d437857597bd7f3d5b58651ef/src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx#L77).
Can't we have something like `getManaBalanceForNetwork(state:RootState, network: Network)`?  Because as far as I'm aware (I might be mistaken), the contract will live in the MATIC network. This also opens the window for the the question "should we should this modal at all if we're on another network that's not MATIC?" because we could disable the "Buy slots" button in that case.

src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.container.ts
LautaroPetaccio: I'm not entirely sure I understand this comment. Are you saying that `getThirdParty` shouldn't exist?

src/modules/wallet/selectors.ts
LautaroPetaccio: Created the `getManaBalanceForNetwork(state:RootState, network: Network)` method!

src/lib/urn.ts
LautaroPetaccio: Renamed the `getThirdPartyId` method to `extractThirdPartyId`.

src/lib/urn.ts
```diff
@@ -92,6 +100,11 @@ export function toLegacyURN(urn: URN): URN {
...
+  const decodedURN = decodeURN(urn) as DecodedThirdPartyURN
+  return `urn:decentraland:${decodedURN.protocol}:collections-thirdparty:${decodedURN.thirdPartyName}`
+}
```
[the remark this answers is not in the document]
LautaroPetaccio: Changed the implementation to use the conditional as you've mentioned!

src/modules/wallet/selectors.ts
```diff
@@ -0,0 +1,8 @@
+import { Network } from '@dcl/schemas'
+import { RootState } from 'modules/common/types'
+import { getNetworks } from 'decentraland-dapps/dist/modules/wallet/selectors'
```
nicosantangelo, line 3: Can we move this a line above 👉👈

tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "skipLibCheck": true,
     "noImplicitAny": true,
     "noImplicitThis": true,
-    "noUnusedLocals": true,
+    "noUnusedLocals": false,
```
nicosantangelo, line 14: 👓

src/components/Modals/BuyItemSlotsModal/BuyItemSlotsModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+    if (tiers && tiers.length === 0 && !isFetchingTiers) {
+      onFetchThirdPartyItemSlots()
+    }
+  }
+
+  render() {
+    const { onClose, isFetchingTiers, name, isBuyingItemSlots, tiers, error } = this.props
+    const { selectedTierId } = this.state
+    console.log('Starting this shit')
```
nicosantangelo, line 63: 👁️ . 👁️

nicosantangelo (approved):
Minor comments, but great work

src/components/Modals/BuyItemSlotsModal/BuyItemSlotsModal.tsx
LautaroPetaccio: Ups, removed it!

tsconfig.json
LautaroPetaccio: Changed it to the value it had before.

src/modules/wallet/selectors.ts
LautaroPetaccio: Done!

PR #1665, opened 2021-11-14
feat: support asset.json

### LINKED ISSUE #1654: Support asset.json for wearables
When an item is created, the user should be able to provide an `asset.json` file within the zip file that follows the schema of an item and allows the user to programmatically fill the properties of an item without having to use the UI. This would also help to define certain important paths (like the paths to the representations) where more than one GLB is present, or for portable experiences.

[Wearable-Jetpack.zip](https://github.com/decentraland/builder/files/7498005/Wearable-Jetpack.zip)

cazala:
Closes #1654

This PR adds support for adding an `asset.json` within the zip file when creating a new item, and this file can contain the following data to populate the state of the creation modal with data (this avoids having to enter this data manually, and allows to correct wrong data picked but automatically by the creation flow, for example, if you zip contains more than one GLB file you can use the `model` prop to point to the right one, same thing using the `thumbnail` prop to point to the right PNG in case there are many.

The data that can be provided via the `asset.json` file should follow this schema:

```ts
type ItemAssetJson = {
  name?: string
  description?: string
  category?: WearableCategory // "hat", "hair", "upper_body", etc...
  rarity?: ItemRarity // "unique", "mythic", "legendary", etc...
  thumbnail?: string // path/to/thumbnail.png
  model?: string // path/to/model.glb
  bodyShape?: BodyShapeType // "male", "female" or "both"
}

### CODE REVIEW

File: src/components/Modals/CreateItemModal/CreateItemModal.types.ts [hunk 1]
```diff
@@ -42,6 +42,16 @@ export type CreateItemModalMetadata = {
...
 }
 
+export type ItemAssetJson = {
+  name?: string
+  description?: string
+  category?: WearableCategory
+  rarity?: ItemRarity
+  thumbnail?: string
+  model?: string
+  bodyShape?: BodyShapeType
```
LautaroPetaccio, line 52: If you create an `asset.json` that has some things like `bodyShape` as "Female", that is, the value it should have but uppercased, the value is set to "Female" (which is wrong) and the UI seems to let you create the item with it.
It think that we should have a minimal way to check the values input here because it might cause some issues.
Screen Shot 2021-11-15 at 16 03 42

LautaroPetaccio: Following the same idea, if the model path is wrong, the process looks for any of the GLBs but IMO it should throw to let you know that you made a mistake.

LautaroPetaccio (changes requested):

src/components/Modals/CreateItemModal/CreateItemModal.types.ts
cazala: good idea! I added validation for file paths and enum values:
Screen Shot 2021-11-16 at 4 34 39 PM
Screen Shot 2021-11-16 at 4 54 46 PM

src/components/Modals/CreateItemModal/CreateItemModal.types.ts
```diff
@@ -42,6 +42,16 @@ export type CreateItemModalMetadata = {
   changeItemFile?: boolean
 }
 
+export type ItemAssetJson = {
+  name?: string
+  description?: string
```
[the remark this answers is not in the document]
cazala: fixed ✅

LautaroPetaccio (approved):

PR #1669, opened 2021-11-15
fix: edit urn modal

### LINKED ISSUE #1661: Add the Edit URN Modal
Add a modal that can edit both a TP Collection or an Item URN. We could create a generic EditThirdPartyURN and let it decide which URN is being edited using the resource supplied. It should be opened from the Edit URN context menus (already present in the code as `onOpenModal('BuySlotModal', {})`)

nicosantangelo:
https://user-images.githubusercontent.com/692077/141846644-5b7702d2-3c6b-4f12-98aa-40e4934fe36e.mov

Closes #1661


src/modules/collection/sagas.ts
```diff
@@ -134,7 +135,6 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
     try {
       const collections: Collection[] = yield call(() => builder.fetchCollections(address))
       yield put(fetchCollectionsSuccess(collections))
-      yield put(closeModal('CreateCollectionModal'))
```
LautaroPetaccio, line 137: Was this modal mistakenly closed before? Because the `handleSaveCollectionSuccess` only handles the actions of saving collections.

LautaroPetaccio (changes requested):

src/components/Modals/EditURNModal/EditURNModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { buildThirdPartyURN, DecodedURN, decodeURN, URNType } from 'lib/urn'
+import { Item } from 'modules/item/types'
+import { Collection } from 'modules/collection/types'
+import { Props, State, EditURNModalMetadata } from './EditURNModal.types'
+
+export default class EditURNModal extends React.PureComponent<Props, State> {
+  decodedURN: DecodedURN<URNType.COLLECTIONS_THIRDPARTY> = this.decodeURN()
+
+  state: State = {
+    newURNSection: ''
```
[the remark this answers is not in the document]
nicosantangelo: I couldn't figure out a better name for this property, but yes!. It holds either the `collection-id` or the `token-id`, depending on the URN (the last part of the URN)

src/modules/collection/sagas.ts
nicosantangelo: I think so, I checked the commit that added it and didn't seem to make sense:
[`fix: Fix BuilderAPI cyclic dependency`](https://github.com/decentraland/builder/commit/9073f52eff63452731431bd7724a854a30c4fb28)

LautaroPetaccio (approved):

PR #1671, opened 2021-11-15
feat: copy dcl collection urn

### LINKED ISSUE #1564: Display Items and Collections URN
- Display the collection urn

- Display the item urn 
(Same place as the collection urn)

- We need to add the `copy` button

nicosantangelo:
https://user-images.githubusercontent.com/692077/141854898-5d6aea95-5f6a-406f-b085-4c15c4fc71b9.mov

@nachomazzara you think it's better to just hide `Copy URN` for non-published collections?

Closes #1564

### DIFF
--- a/src/modules/translation/languages/zh.json
+++ b/src/modules/translation/languages/zh.json
@@ -1010,6 +1010,7 @@
     "edit_price": "修改价格",
     "see_details": "查看详细信息",
     "edit_urn": "编辑 URN",
+    "copy_urn": "复制 URN",
     "reset_item": "撤消"
   },
   "item": {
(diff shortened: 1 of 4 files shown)


src/modules/translation/languages/zh.json
```diff
@@ -1021,6 +1022,7 @@
     "beneficiary": "受益人",
     "supply": "供应",
     "collection": "采集",
+    "edit_urn": "编辑 URN",
```
LautaroPetaccio, line 1025: Isn't `copy_urn` missing?

LautaroPetaccio (changes requested):

src/modules/translation/languages/zh.json
nicosantangelo: it was, I had a mixup with different branches it seems

LautaroPetaccio (approved):

PR #1673, opened 2021-11-16
feat: View third party collections

LautaroPetaccio:
This PR modifies the `getAuthorizedCollections` selector to show the third party collections owned by the user.


src/modules/collection/selectors.ts
```diff
@@ -30,10 +34,21 @@ export const getWalletCollections = createSelector<RootState, Collection[], stri
...
-  (collections, address) => collections.filter(collection => address && canSeeCollection(collection, address))
+export const getAuthorizedCollections = createSelector<
+  RootState,
+  Collection[],
+  string | undefined,
+  Record<string, ThirdParty>,
+  Collection[]
+>(getCollections, getAddress, getThirdParties, (collections, address, thirdParties) =>
+  collections.filter(collection => {
+    if (decodeURN(collection.urn!).type === URNType.COLLECTIONS_THIRDPARTY) {
```
nicosantangelo, line 45: We can use collection/utils.ts `isThirdParty` method here

src/modules/thirdParty/utils.ts
```diff
@@ -1,8 +1,13 @@
...
+import { Collection } from 'modules/collection/types'
 import { ThirdParty } from './types'
 
 export function isUserManagerOfThirdParty(address: string, thirdParty: ThirdParty): boolean {
   return thirdParty.managers.map(manager => manager.toLowerCase()).includes(address.toLowerCase())
 }
 
 export const getAvailableSlots = (thirdParty: ThirdParty): BN => new BN(thirdParty.maxItems).sub(new BN(thirdParty.totalItems))
+
+export const getThirdPartyOfCollection = (thirdParties: Record<string, ThirdParty>, collection: Collection) =>
```
nicosantangelo, line 12: I think `getThirdPartyForCollection` is a bit clearer of a method name, what do you think?

nicosantangelo (commented):
Minor comments

src/modules/thirdParty/utils.ts
LautaroPetaccio: Renamed it with the suggested name.

src/modules/collection/selectors.ts
LautaroPetaccio: Changed to `isThirdParty`!

src/lib/urn.ts
```diff
@@ -89,6 +89,10 @@ export function buildCatalystItemURN(contractAddress: string, tokenId: string):
   return `urn:decentraland:${getNetworkURNProtocol(Network.MATIC)}:collections-v2:${contractAddress}:${tokenId}`
 }
 
+export function buildDefaultCatalystCollectionURN() {
+  return `urn:decentraland:${getNetworkURNProtocol(Network.MATIC)}:collections-v2:0x0000000000000000000000000000000000000000`
+}
```
nicosantangelo, line 94: Question, would it make sense to generalize `buildCatalystItemURN`?

maybe:

```ts

export function buildCatalystItemURN(contractAddress: string, tokenId?: string): URN {
 let urn = `urn:decentraland:${getNetworkURNProtocol(Network.MATIC)}:collections-v2:${contractAddress}`
  if(tokenId) {
    urn += `:${tokenId}`  
  }
  return urn
}

export function buildDefaultCatalystCollectionURN() {
  return buildCatalystItemURN(Address.ZERO.toString())
}

nicosantangelo (commented):
Comment, but LGTM

nicosantangelo (approved):

src/lib/urn.ts
LautaroPetaccio: I wouldn't generalize it because the catalystItemURN needs to have a token id to be a valid item URN. If we make it optional, we'll open the door to people using it incorrectly.

nicosantangelo: Gotcha, thought that'd be the case

PR #1683, opened 2021-11-24
fix: Tiers types

LautaroPetaccio:
The types in the tiers module weren't written as they're meant to be, this PR fixes that.


nicosantangelo (approved):
❤️

PR #1688, opened 2021-11-29
feat: Update item urns on URN change

LautaroPetaccio:
This PR updates the URN of the items of a collection that was saved.


src/modules/item/reducer.ts
```diff
@@ -305,6 +307,39 @@ export function itemReducer(state: ItemState = INITIAL_STATE, action: ItemReduce
             }
           }
         }
+        case SAVE_COLLECTION_SUCCESS: {
+          const collectionURN = decodeURN(transaction.payload.collection.urn)
+
+          collectionURN.suffix
```
nicosantangelo, line 313: whops!

src/lib/urn.ts
```diff
@@ -34,7 +34,7 @@ const protocolMatcher = '(?<protocol>mainnet|ropsten|matic|mumbai|off-chain)'
 const typeMatcher = '(?<type>base-avatars|collections-v2|collections-thirdparty)'
 
 const baseAvatarsSuffixMatcher = '((?<=base-avatars:)BaseMale|BaseFemale)'
-const collectionsSuffixMatcher = '((?<=collections-v2:)0x[a-fA-F0-9]{40})'
+const collectionsSuffixMatcher = '((?<=collections-v2:)(?<collectionAddress>0x[a-fA-F0-9]{40}))?(:(?<tokenId>[^:|\\s]+))'
```
nicosantangelo, line 37: I think this is missing an `=` after each `<`
```suggestion
const collectionsSuffixMatcher = '((?<=collections-v2:)(?<=collectionAddress>0x[a-fA-F0-9]{40}))?(:(?<=tokenId>[^:|\\s]+))'
```

And that might be breaking the tests

nicosantangelo (changes requested):
Broken tests!

src/modules/item/reducer.ts
LautaroPetaccio: Removed it!

src/lib/urn.ts
LautaroPetaccio: The equal sign matches there the exact string (at least as far as I understood), but that's not needed for the collectionAddress nor for the tokenId.

nicosantangelo: Correct! my bad

PR #1689, opened 2021-11-30
feat: download wearable zip

cazala:
This PR adds a download button to the editor to get the contents of the item as a zip file:

Screen Shot 2021-11-30 at 7 44 52 PM

I was going to add only this feature for the reviewers, but I left it for the creators as well, since they might lose the files and could be handy to get them back.


src/modules/project/sagas.ts
```diff
@@ -205,12 +202,10 @@ export function* projectSaga(builder: BuilderAPI) {
...
     )
-    for (const filename of Object.keys(files)) {
-      zip.file(filename, files[filename])
-    }
 
-    const artifact: Blob = yield call([zip, 'generateAsync'], { type: 'blob' })
-    yield call(saveAs, artifact, `${sanitizedName}.zip`)
+    // download zip
+    const name = project.title.replace(/\s/g, '_')
+    yield call(downloadZip, name, files)
```
LautaroPetaccio, line 208: Would you please add a test for this?

LautaroPetaccio (changes requested):

src/modules/project/sagas.ts
cazala: I tried adding a `modules/project/sagas.spec.ts` like this:

```ts
import { expectSaga } from 'redux-saga-test-plan'
import * as matchers from 'redux-saga-test-plan/matchers'
import { call, select } from 'redux-saga/effects'
import { Scene } from 'modules/scene/types'
import { getData as getPools } from 'modules/pool/selectors'
import { getData as getProjects } from 'modules/project/selectors'
import { getData as getScenes } from 'modules/scene/selectors'
import { setExportProgress } from 'modules/editor/actions'
import { getName } from 'modules/profile/selectors'
import { BuilderAPI } from 'lib/api/builder'
import { projectSaga } from './sagas'
import { Project } from './types'
import { exportProjectSuccess } from './actions'
import { createFiles } from './export'
import { downloadZip } from 'lib/zip'

const builderAPI = {} as BuilderAPI

describe('when exporting a project', () => {
  let project: Project
  let projects: Record<string, Project>
  let scene: Scene
  let scenes: Record<string, Scene>
  let files: Record<string, Blob | string>

  beforeEach(() => {
    project = { id: 'aProject', title: 'Some Project', sceneId: 'aScene' } as Project
    projects = { aProject: project }
    scene = { id: 'aScene' } as Scene
    scenes = { aScene: scene }
    files = { 'model.glb': new Blob(), 'code.ts': 'console.log("hola")' }
  })

  it.skip('should download a zip with all the project files', () => {
    return expectSaga(projectSaga, builderAPI)
      .provide([
        [select(getProjects), projects],
        [select(getPools), {}],
        [select(getScenes), scenes],
        [select(getName), 'pepito'],
        [matchers.call.fn(createFiles), files],
        [call(downloadZip, 'some_project', files), undefined]
      ])
      .put(setExportProgress({ loaded: 0, total: 0 }))
      .put(exportProjectSuccess())
      .run({ silenceTimeout: true })
  })
})
```

But the `import { createFiles } from './export'` is making jest to blow up since it ends up importing babylon, and for some reason jest doesn't like it :(

src/lib/api/builder.ts
```diff
@@ -716,4 +716,15 @@ export class BuilderAPI extends BaseAPI {
   isAxiosError(error: any): error is AxiosError {
     return error.isAxiosError
   }
+
+  async fetchBlob(hash: string) {
```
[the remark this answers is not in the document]
cazala: done ✅

src/modules/item/reducer.ts
```diff
@@ -249,6 +260,14 @@ export function itemReducer(state: ItemState = INITIAL_STATE, action: ItemReduce
       }
     }
 
+    case DOWNLOAD_ITEM_SUCCESS: {
```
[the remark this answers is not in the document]
cazala: done ✅

src/modules/item/utils.ts
```diff
@@ -328,3 +328,13 @@ export function areSynced(item: Item, entity: DeploymentWithMetadataContentAndPo
   }
   return true
 }
+
+export function buildZipFiles(files: Record<string, Blob | string>, areEqual: boolean) {
```
LautaroPetaccio, line 332: What do you think about renaming this method to something like `buildZipContent`? I was expecting it to work with files, but it's just building the content that it should have.

src/modules/item/utils.spec.ts
```diff
@@ -50,3 +50,18 @@ describe('when transforming third party items to be sent to a contract method',
     ])
   })
 })
+
+describe('when building building zip files and representations are equal', () => {
```
LautaroPetaccio, line 54: There's a repeated `building` here haha.
```suggestion
describe('when building zip files and representations are equal', () => {
```

src/modules/item/sagas.spec.ts
```diff
@@ -617,3 +622,85 @@ describe('when publishing third party items', () => {
...
+
+  describe('when an item has two different representations for male and female', () => {
+    it('should download a zip file with both /male and /female directories', () => {
+      const itemId = 'maleAndFemale'
+      const files = { 'male/model_male.glb': 'QmhashMale', 'female/model_female.glb': 'QmhashFemale' }
+      return expectSaga(itemSaga, builderAPI)
+        .provide([
+          [select(getItemsById), itemsById],
+          [call([builderAPI, 'fetchContent'], 'Qmhash'), new Blob()],
+          [matchers.call.fn(buildZipFiles), files],
```
LautaroPetaccio, line 681: It would have been great to have this matcher to do a deep equal comparison. We'll do it next time.

LautaroPetaccio (changes requested):

src/modules/item/sagas.ts
```diff
@@ -334,6 +340,62 @@ export function* itemSaga(builder: BuilderAPI) {
...
+      const maleHashes: string[] = []
+      const femaleHashes: string[] = []
+
+      // download blobs and add them to zip
+      for (const path of Object.keys(item.contents)) {
+        const hash = item.contents[path]
+        // skip downloading already donwloaded hashes
+        if (!hashes.has(hash)) {
+          const blob = yield call([builder, 'fetchContent'], hash)
+          files[path] = blob
```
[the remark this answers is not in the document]
cazala: true, i refactored it to download the files in parallel

src/modules/item/utils.ts
cazala: done ✅

src/modules/item/utils.spec.ts
cazala: fixed :)

LautaroPetaccio (approved):

PR #1698, opened 2021-12-10
fix: compare representations in areSynced

### LINKED ISSUE #1696: Collection is not marked as dirty if the bodyshapes changed
If the an item has both representations but later only one is left. (or If the item has one representation and later both are being added). Both representations will have the same hash and therefore the item is not marked as dirty.

We must check if the same is happen when we modify the name, desc, tags, and category

cazala:
Fixes #1696

The issue was that if you had an item with BOTH representations, and you removed one, the item would not be marked as `unsynced`, so you wouldn't be able to push the changes for review (because we only checked that all the `item.contents` existed on the `entity.contents`). Now we compare that both representations are equal.

### DIFF
--- a/src/modules/item/utils.spec.ts
+++ b/src/modules/item/utils.spec.ts
@@ -1,5 +1,5 @@
-import { Item, ItemMetadataType, ItemType, WearableBodyShape, WearableCategory } from './types'
-import { buildItemMetadata, buildZipContents, toThirdPartyContractItems } from './utils'
+import { Item, ItemMetadataType, ItemType, WearableBodyShape, WearableCategory, WearableRepresentation } from './types'
+import { buildItemMetadata, buildZipContents, toThirdPartyContractItems, areEqualArrays, areEqualRepresentations } from './utils'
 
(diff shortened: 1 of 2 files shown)


fzavalia (changes requested):
Tests for areEqualArrays and areEqualRepresentations?

cazala:
Added tests @fzavalia !

src/modules/item/utils.spec.ts
```diff
@@ -65,3 +65,61 @@ describe('when building zip files and representations are NOT equal', () => {
     })
   })
 })
+
+describe('when comparing two arrays', () => {
+  it('should return true when the elements are the same in the same order', () => {
+    expect(areEqualArrays(['a', 'b', 'c', 'd'], ['c', 'a', 'b', 'd'])).toBe(true)
```
fzavalia, line 71: these are not in the same order

cazala: fixed 😅

fzavalia (approved):

PR #1702, opened 2021-12-17
 feat: add basic support for third party item curation

nicosantangelo:


src/modules/ui/itemEditor/selectors.ts
```diff
@@ -0,0 +1,5 @@
+import { RootState } from 'modules/common/types'
+
+export const getState = (state: RootState) => state.ui.itemEditor
+
+export const getSelectedThirdPartyItemIds = (state: RootState) => getState(state).selectedThirdPartyItemIds
```
LautaroPetaccio, line 5: I know these are simple, but a simple test for each of them would ensure that they're covered from any eventual change.

src/modules/ui/itemEditor/actions.ts
```diff
@@ -0,0 +1,13 @@
...
+// Reset all third party item selections
+
+export const RESET_THIRD_PARTY_ITEMS = 'Reset third party items'
+export const resetThirdPartyItems = () => action(RESET_THIRD_PARTY_ITEMS)
+export type ResetThirdPartyItems = ReturnType<typeof resetThirdPartyItems>
+
+// Toggle third party item selection
+
+export const TOGGLE_THIRD_PARTY_ITEM = 'Toggle third party item'
+export const toggleThirdPartyItem = (itemId: string, isSelected: boolean) => action(TOGGLE_THIRD_PARTY_ITEM, { itemId, isSelected })
```
LautaroPetaccio, line 12: Would you mind testing these actions?

src/components/ItemEditorPage/LeftPanel/LeftPanel.tsx
```diff
@@ -1,66 +1,92 @@
...
 import Header from './Header'
 import Items from './Items'
 import Collections from './Collections'
 import { Props } from './LeftPanel.types'
 import './LeftPanel.css'
+import { getCollectionType } from 'modules/collection/utils'
 
 export default class LeftPanel extends React.PureComponent<Props> {
+  componentDidMount() {
+    this.props.onResetThirdPartyItems()
```
LautaroPetaccio, line 16: Should we deconstruct the props here?

src/lib/urn.ts
```diff
@@ -110,6 +110,17 @@ export function extractThirdPartyTokenId(urn: URN) {
   return `${thirdPartyCollectionId}:${thirdPartyTokenId}`
 }
 
+// TODO: This logic is repeated in collection/util's `getCollectionType`, but being used only for items (item.urn).
+// It should probably be replaced by a getItemType or we should see if it's better to only keep one way of doing this
+export function isThirdParty(urn?: string) {
```
LautaroPetaccio, line 115: Could it be renamed to isThirdPartyItem then?
Can the urns be empty? Shouldn't we say that they could be strings or null values?

LautaroPetaccio (changes requested):

src/modules/ui/itemEditor/selectors.ts
nicosantangelo: I don't mind adding tests at all! but I'm not entirely sold on this one on particular. There's no logic, just property accessors. I'm not sure I'm convinced they'll be more useful than cumbersome.

Do you have a case in mind where they've useful? to me if we change any of these Typescript should be enough.

src/lib/urn.ts
nicosantangelo: > Could it be renamed to isThirdPartyItem then?

not really, it works for both types of URNs, Collections and Items. It is useful as is, but as I tried to say in the comment is not really being used right now for any other use case that for `item.urn`s

> Can the urns be empty? Shouldn't we say that they could be strings or null values?

I think this is well typed in item/types.ts where it says:

```ts
export type Item = BaseItem & {
  urn?: string
  //(...)
}
```

because in builder.ts we're doing the following:

```ts
// fromRemoteItem
  if (remoteItem.urn) item.urn = remoteItem.urn
```

right?

LautaroPetaccio: The builder server should always return a valid URN (string) or a null value. This is done by using the `toFullItem` method.

src/modules/ui/itemEditor/selectors.ts
LautaroPetaccio: We should add something in the testing standards about this.

src/lib/urn.ts
nicosantangelo: OH now I getcha, changing

nicosantangelo: We should change the frontend too

LautaroPetaccio (approved):

PR #1707, opened 2021-12-31
feat: Add MANA approval to the buy items slots flow

LautaroPetaccio:
This PR does the following:
- Adds the MANA approval modal that will be shown if the user has not authorized the MANA contract to operate with the tiers contract.
- Changes the way the modal closes to clear the errors when it does.
- Fixes an issue where errors with empty strings wouldn't show the error message.
- Abstracts the process of building a MANA authorization for a contract.


src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
```diff
@@ -49,11 +58,23 @@ export default class ThirdPartyCollectionDetailPage extends React.PureComponent<
...
+
   handleBuySlot = () => {
-    const { onOpenModal, thirdParty } = this.props
-    onOpenModal('BuyItemSlotsModal', {
-      thirdParty
-    })
+    const { onOpenModal, thirdParty, authorizations } = this.props
+    const manaAuthorization = this.getManaAuthorization()
+    console.log('Authorizations', authorizations)
+    console.log('ManaAuthorization', manaAuthorization)
```
LautaroPetaccio, line 69: These are here to check why the authorizations are not being persisted.

src/components/Modals/BuyItemSlotsModal/BuyItemSlotsModal.tsx
```diff
@@ -16,6 +16,12 @@ export default class BuyItemSlotsModal extends React.PureComponent<Props, State>
     selectedTierId: undefined
   }
 
+  onCloseModal = (): void => {
```
nicosantangelo, line 19: To follow conventions, I think this should be called `handleOnCloseModal` right?

src/components/Modals/BuyItemSlotsModal/BuyItemSlotsModal.container.ts
```diff
@@ -22,6 +22,7 @@ const mapState = (state: RootState): MapStateProps => {
 const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
   onFetchThirdPartyItemSlots: () => dispatch(fetchThirdPartyItemTiersRequest()),
   onTierSelected: () => dispatch(clearTiersError()),
+  onBeforeClose: () => dispatch(clearTiersError()),
```
nicosantangelo, line 25: This might be coincidental, but given that we have two methods with different names poiting to the same dispatched action, what if we just have `onClearTiersError()` and call that instead?

src/lib/mana.ts
```diff
@@ -1,5 +1,23 @@
...
+import { Authorization, AuthorizationType } from 'decentraland-dapps/dist/modules/authorization/types'
+import { ContractName, getContract } from 'decentraland-transactions'
+
 const MANA_SYMBOL = '⏣'
 
 export function addSymbol(num: string | number) {
   return num > 0 ? `${MANA_SYMBOL} ${num.toString()}` : ''
 }
+
+export function buildManaAuthorization(address: string, chainId: ChainId, contractName: ContractName): Authorization {
```
nicosantangelo, line 11: Maybe we can test this method?

nicosantangelo (commented):
A few silly things

src/components/Modals/BuyItemSlotsModal/BuyItemSlotsModal.tsx
LautaroPetaccio: Changed it!

src/components/Modals/BuyItemSlotsModal/BuyItemSlotsModal.container.ts
LautaroPetaccio: I definitely though about it, but from the perspective of the component as an isolated chunk of code, there's nothing to explicitly clear the errors, so it might be a bit confusing.

src/lib/mana.ts
LautaroPetaccio: Oh, I forgot to add tests to it! Added them.

nicosantangelo (approved):

PR #1717, opened 2022-01-11
feat: Use custom selectors to access TP collection and items

LautaroPetaccio:
This PR adds two new selectors, one for the items and the other one for the collections to see if the user has view and edit rights, taking into consideration the possibility of the item or the collection being third party.


src/components/ItemEditorPage/LeftPanel/Header/Header.container.ts
```diff
@@ -19,11 +19,13 @@ const mapState = (state: RootState): MapStateProps => {
...
   }
+  const address = getAddress(state)
   return {
-    address: getAddress(state),
+    address,
     isLoggedIn: isLoggedIn(state),
     isReviewing: isReviewing(state),
-    collection
+    collection,
+    hasEditRights: collection !== undefined && address != undefined && hasViewAndEditRights(state, address, collection)
```
juanigallo, line 28: Hi @LautaroPetaccio, is there any reason why in the first condition we are using `!==` and in the second `!=`? Also which is the benefit of doing that and not directly checking existence?

`hasEditRights: collection && address && hasViewAndEditRights(state, address, collection)`

LautaroPetaccio: Hi Juani! You're right, it was a typo, I forgot to add the other `=`. Fixed it!

nicosantangelo (approved):

Issue #1709, opened 2022-01-04
When the collection name is large the status icon is not displayed. 


nachomazzara:
What about moving it to the left? 

![Screen Shot 2022-01-04 at 15.49.41.png](https://images.zenhubusercontent.com/5e7388701e944359f1d16293/3917498e-7ab6-4475-bdf7-52a95bd2785d)

juanigallo:
Hi @nachomazzara is there any easy way to see `/collections` in local? it is always showing me as a guest user so the collections tabs is hidden

nachomazzara:
@juanigallo  You need to create your own collections in order to see any. Also, you will need to connect your wallet

juanigallo:
Yeah, I tried to do that but for some reason the fetch to get the profile has a wrong URL and of course fails. I will do some more research at night

![image](https://user-images.githubusercontent.com/4969737/149203193-efad0876-b95e-495d-a459-8eb2431b0bc5.png)

juanigallo:
NVM, I was `REACT_APP_PEER_URL` env variable which is in the `.env.example` but not in the README, I will add a PR to fix that later

nachomazzara:
@juanigallo thanks

cazala:
🙌

PR #1718, opened 2022-01-12
Moves status icon to the beginning of the collection name

juanigallo:
Fixes https://github.com/decentraland/builder/issues/1709

left-status


LautaroPetaccio (approved):
Thanks for the fix! Looks good.

PR #1719, opened 2022-01-12
fix: use fetchEntitiesByPointers instead of fetchAllDeployments

cazala:
Replaces the use of the deprecated and throttled `fetchAllDeployments` for `fetchEntitiesByPointers` and `fetchEntitiesByIds`. The first method returns a `DeploymentWithContentAndMetadata` and the latter ones an `Entity`, so I had to make some refactors but all the important data was available.

### DIFF
--- a/src/modules/entity/actions.ts
+++ b/src/modules/entity/actions.ts
@@ -1,23 +1,37 @@
-import { DeploymentOptions, DeploymentPreparationData, DeploymentWithMetadataContentAndPointers } from 'dcl-catalyst-client'
 import { action } from 'typesafe-actions'
+import { DeploymentPreparationData } from 'dcl-catalyst-client'
+import { Entity, EntityType } from 'dcl-catalyst-commons'
 
-// Fetch Item Entity
-export const FETCH_ENTITIES_REQUEST = '[Request] Fetch Entities'
-export const FETCH_ENTITIES_SUCCESS = '[Success] Fetch Entities'
-export const FETCH_ENTITIES_FAILURE = '[Failure] Fetch Entities'
+// Fetch Entities By Pointers
+export const FETCH_ENTITIES_BY_POINTERS_REQUEST = '[Request] Fetch Entities By Pointers'
+export const FETCH_ENTITIES_BY_POINTERS_SUCCESS = '[Success] Fetch Entities By Pointers'
+export const FETCH_ENTITIES_BY_POINTERS_FAILURE = '[Failure] Fetch Entities By Pointers'
 
-export const fetchEntitiesRequest = (options: DeploymentOptions<DeploymentWithMetadataContentAndPointers>) =>
-  action(FETCH_ENTITIES_REQUEST, { options })
-export const fetchEntitiesSuccess = (
-  entities: DeploymentWithMetadataContentAndPointers[],
-  options: DeploymentOptions<DeploymentWithMetadataContentAndPointers>
-) => action(FETCH_ENTITIES_SUCCESS, { entities, options })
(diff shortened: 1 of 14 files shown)


src/modules/item/sagas.ts
```diff
@@ -89,6 +88,7 @@ import { Item, Rarity, CatalystItem, BodyShapeType } from './types'
 import { getData as getItemsById, getItems, getEntityByItemId, getCollectionItems } from './selectors'
 import { ItemTooBigError } from './errors'
 import { buildZipContents, getMetadata, isValidText, MAX_FILE_SIZE, toThirdPartyContractItems } from './utils'
+import { Entity, EntityType } from 'dcl-catalyst-commons'
```
LautaroPetaccio, line 91: Would you mind moving this import to the top?

src/modules/entity/reducer.ts
```diff
@@ -46,21 +56,23 @@ export function entityReducer(state: EntityState = INITIAL_STATE, action: Entity
...
+    case FETCH_ENTITIES_BY_IDS_SUCCESS: {
       return {
         ...state,
         loading: loadingReducer(state.loading, action),
         error: null,
         data: {
           ...state.data,
           ...action.payload.entities.reduce((obj, entity) => {
-            obj[entity.entityId] = entity
+            obj[entity.id] = entity
```
LautaroPetaccio, line 68: Would you mind adding tests for the new actions that are being reduced?

src/modules/entity/actions.ts
```diff
@@ -1,23 +1,37 @@
...
+export const FETCH_ENTITIES_BY_POINTERS_REQUEST = '[Request] Fetch Entities By Pointers'
+export const FETCH_ENTITIES_BY_POINTERS_SUCCESS = '[Success] Fetch Entities By Pointers'
+export const FETCH_ENTITIES_BY_POINTERS_FAILURE = '[Failure] Fetch Entities By Pointers'
+
+export const fetchEntitiesByPointersRequest = (type: EntityType, pointers: string[]) =>
+  action(FETCH_ENTITIES_BY_POINTERS_REQUEST, { type, pointers })
+export const fetchEntitiesByPointersSuccess = (type: EntityType, pointers: string[], entities: Entity[]) =>
+  action(FETCH_ENTITIES_BY_POINTERS_SUCCESS, { entities, type, pointers })
+export const fetchEntitiesByPointersFailure = (type: EntityType, pointers: string[], error: string) =>
+  action(FETCH_ENTITIES_BY_POINTERS_FAILURE, { error, type, pointers })
```
LautaroPetaccio, line 15: Would you mind adding tests for the actions?

LautaroPetaccio (changes requested):

src/modules/entity/reducer.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    const state: EntityState = {
+      data: {},
+      loading: [fetchEntitiesByPointersRequest(EntityType.WEARABLE, entity.pointers)],
+      error: null
+    }
+
+    const action = fetchEntitiesByPointersSuccess(EntityType.WEARABLE, entity.pointers, [entity])
+    const newState = entityReducer(state, action)
+    expect(newState.data[entity.id]).toEqual(entity)
+    expect(newState.loading).toHaveLength(0)
```
LautaroPetaccio, line 54: These expectations are fine, but I would have checked the entire data structure to be sure that it added just one with the id we wanted (it's a bit more robust).

LautaroPetaccio (approved):

src/modules/entity/reducer.spec.ts
cazala: done 👍

PR #1732, opened 2022-01-17
fix: Only show the edit URN link when the URN belongs to a TP Item

LautaroPetaccio:


src/components/ItemDetailPage/ItemDetailPage.tsx
```diff
@@ -175,10 +176,11 @@ export default class ItemDetailPage extends React.PureComponent<Props> {
                     {item.urn}
 
                     <div className="urn-actions">
-                      <span className="link" onClick={this.handleEditURN}>
-                        {t('item.edit_urn')}
-                      </span>
-
+                      {decodeURN(item.urn).type === URNType.COLLECTIONS_THIRDPARTY ? (
```
nicosantangelo, line 179: Maybe we can just use `isThirdParty(item.urn)` here?

LautaroPetaccio: Changed it!

nicosantangelo (approved):

PR #1764, opened 2022-02-02
chore: Refactor the Info popup icon into a InfoIcon

LautaroPetaccio:
This simple PR refactors the Info popup icon into a new component InfoIcon, so we can use the icon elsewhere.


src/components/Info/Info.tsx
```diff
@@ -1,17 +1,19 @@
...
   render() {
-    const { className = '', content = '' } = this.props
+    const { className, content = '' } = this.props
     return (
       <Popup
         className="info-popup"
         content={content}
         position="top center"
-        trigger={<i className={`Info ${className}`} />}
+        trigger={<InfoIcon className={classNames(className, 'infoIcon')} />}
```
nicosantangelo, line 16: are we using the infoIcon class? and, also, why is camelCased?

LautaroPetaccio: We're! it's giving this `InfoIcon` the `cursor: pointer;` property. I'll change it so it's not camel cased.

nicosantangelo (approved):

PR #1770, opened 2022-02-02
feat: Render Not Found page if the collection is not of TP type

### LINKED ISSUE #1763: Check collection types when rendering details
The current implementation allows the user to visit a Standard or Third Party collection using the wrong URL.

For example if the Standard collection of Id X exists, visiting:

`/thirdPartyCollections/X`

will render the page instead of redirecting

juanmahidalgo:
Closes #1763 

image

### DIFF
--- a/src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
+++ b/src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
@@ -15,6 +15,7 @@ import {
   PaginationProps,
   Checkbox
 } from 'decentraland-ui'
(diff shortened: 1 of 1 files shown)


LautaroPetaccio (approved):

LautaroPetaccio (changes requested):
I forgot we had: `getCollectionType` in `src/modules/collection/utils.ts` you could use that instead.

juanmahidalgo:
> I forgot we had: `getCollectionType` in `src/modules/collection/utils.ts` you could use that instead.

Oh nice! Good call @LautaroPetaccio. Code updated using that util method :)

LautaroPetaccio (approved):

PR #1771, opened 2022-02-03
feat: Hide CollectionStatus for TP collections

juanmahidalgo:
## Description

Do not render the `CollectionStatus` badge for TP collections.

### DIFF
--- a/src/components/CollectionStatus/CollectionStatus.tsx
+++ b/src/components/CollectionStatus/CollectionStatus.tsx
@@ -4,12 +4,15 @@ import { Popup } from 'decentraland-ui'
 
 import { SyncStatus } from 'modules/item/types'
 import { Props } from './CollectionStatus.types'
+import { getCollectionType } from 'modules/collection/utils'
+import { CollectionType } from 'modules/collection/types'
 import './CollectionStatus.css'
 
 export default class CollectionStatus extends React.PureComponent<Props> {
   render() {
-    const { status } = this.props
-    return status && status !== SyncStatus.UNPUBLISHED ? (
+    const { status, collection } = this.props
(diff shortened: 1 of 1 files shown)


src/components/CollectionStatus/CollectionStatus.tsx
```diff
@@ -4,12 +4,15 @@ import { Popup } from 'decentraland-ui'
...
+import { getCollectionType } from 'modules/collection/utils'
+import { CollectionType } from 'modules/collection/types'
 import './CollectionStatus.css'
 
 export default class CollectionStatus extends React.PureComponent<Props> {
   render() {
-    const { status } = this.props
-    return status && status !== SyncStatus.UNPUBLISHED ? (
+    const { status, collection } = this.props
+    const shouldRenderStatus = status && status !== SyncStatus.UNPUBLISHED && getCollectionType(collection) !== CollectionType.THIRD_PARTY
```
juanmahidalgo, line 14: As issue suggested, I'm putting the logic inside the component itself instead of each page. This could be moved to the `.container.tsx` file as well, to keep this one just for presentational logic.

LautaroPetaccio: I'm not a big fan on components that decide if they should be rendered or not because reaching the line means that some react's code must have been executed to get there, but the collection status is used in several places and we'll have to repeat code everywhere we use it so this might be the best solution.
What do you think @nicosantangelo ?

nicosantangelo: I agree on js/react code having to be executed without need, but for this particular case, as the Component is being used in quite a few places **and** the logic is non trivial I think it might be better to keep it inside the Component.

I want to reiterate that I am not sure. We could extract the logic into an utils file and reuse it everywhere, but my gut reaction is that it'll be cumbersome when entering the codebase for the first time. Another of those "caveats"

PR #1772, opened 2022-02-03
fix: Info popups

LautaroPetaccio:
This PR fixes the Info popups that:
- Were not opening due to `react-semantic-ui` expecting a semantic component instead of a custom component for the popup trigger.
- Had issues with the styles due to being styled from other components.

### DIFF
--- a/src/components/InfoIcon/InfoIcon.tsx
+++ b/src/components/InfoIcon/InfoIcon.tsx
@@ -1,11 +1,12 @@
 import * as React from 'react'
 import classnames from 'classnames'
+import { omit } from 'decentraland-commons/dist/utils'
 import styles from './InfoIcon.module.css'
 import { Props } from './InfoIcon.types'
 
 export default class InfoIcon extends React.PureComponent<Props> {
   render() {
     const { className } = this.props
-    return <i className={classnames(styles.info, className)} />
+    return <i {...omit(this.props, ['className'])} className={classnames(styles.info, className)} />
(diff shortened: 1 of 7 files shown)


src/components/InfoIcon/InfoIcon.tsx
```diff
@@ -1,11 +1,12 @@
 import * as React from 'react'
 import classnames from 'classnames'
+import { omit } from 'decentraland-commons/dist/utils'
```
nicosantangelo, line 3: for this particular lib, importing for `decentraland-commons` is fine!

src/components/InfoIcon/InfoIcon.tsx
```diff
@@ -1,11 +1,12 @@
...
 import styles from './InfoIcon.module.css'
 import { Props } from './InfoIcon.types'
 
 export default class InfoIcon extends React.PureComponent<Props> {
   render() {
     const { className } = this.props
-    return <i className={classnames(styles.info, className)} />
+    return <i {...omit(this.props, ['className'])} className={classnames(styles.info, className)} />
```
nicosantangelo, line 10: is the omit necessary here? as we're  overwriting it with the next prop

LautaroPetaccio: Yeah, you're right, it gets overwritten, removing the omit.

nicosantangelo (approved):

PR #1776, opened 2022-02-07
Feat: Multiple files upload

### LINKED ISSUE #1563: Create Third Party collection items
- We should allow drag & drop items without collection.
  - If the item is being dragged & dropped into a third party collection, we need to remove the `rarity` (null) and add a default `urn`. This URN is generated based on the third party record, the collection urn and a default generated URN for the item.
  - If the item is being removed from the third party collection, we need to remove the urn (null).
- The item creation modal (ONLY FOR THIRD PARTY ITEMS) should have the following fields:
  - The name of the item
  - The category of the item
  - The URN of the item.
- Once we save the item, we need to fill & check if the URN is valid.
  - We need to check if the urn is well-formed.
  - We need to check if the urn is unique for that specific collection.

### LINKED ISSUE #1570: Upload Collection Items in batch
- Users can submit multiple zip files. Each zip file will be an item.
- Each item must have along with the gltf/glb a `item.json` with the following properties:
```ts
{
  id?: string # i.e: uuid
  name: string
  description: string
  category: string
  urn: string
  tags: string[]
  representations: Representation[]
}

# Representation

{
    "bodyShape": "BaseMale" | "BaseFemale"
    "mainFile": string # "path/to/file"
    "contents": string # "path/to/files"
    "overrideHides": Category[]
    "overrideReplaces": Category[]
}
```

If the user wants to make a `unisex` item, they should upload both representations. 
We need to check that the `item.json` is valid. i.e: bodyShape not repeated.
The files will be processed in the UI to process the JSON and the GLBs to use the same endpoint that we're using today.

We need to display a progress bar in the FE.

We need to add tests

We may check how the scenes and assetpacks are working.

LautaroPetaccio:
This PR does the following:
- Adds the CreateMultipleItemsModal to upload multiple zip files to create items.
- Modifies the CreateSingleItemModal to work with TP items.
- Creates the actions, reducers, sagas and selectors to work with the CreateMultipleItemsModal and upload multiple items.
- Creates the CreateItemsModal to select between creating a single item and multiple items.

Things to consider:
- The CreateItemsModal is used at the moment only on the TP items feature because we'll need first the documentation page to release it to the public and because we'll need to standardize the `asset.json` file support by refactoring the modal to use the functions provided in the builder-client first.

Closes #1563
Closes #1570


src/modules/common/store.ts
```diff
@@ -142,7 +146,18 @@ const store = createStore(rootReducer, enhancer) as RootStore
...
 
-sagasMiddleware.run(rootSaga, builderAPI, catalystClient)
+const getClientAddress = () => getAddress(store.getState())!
+const getClientAuthAuthority = () => {
+  const auths = getData(store.getState())
+  const address = getAddress(store.getState())
+  return auths[address!]
+}
+
+const builderClientUrl: string = BUILDER_SERVER_URL.replace('/v1', '')
```
LautaroPetaccio, line 156: It was either this or adding a new environment variable.

src/components/EditorPage/EditorPage.tsx
```diff
@@ -50,7 +50,7 @@ export default class EditorPage extends React.PureComponent<Props, State> {
   }
 
   handleMouseWheel = (e: Event) => {
-    if ((e as MouseWheelEvent)['ctrlKey']) {
+    if ((e as WheelEvent)['ctrlKey']) {
```
LautaroPetaccio, line 53: Related to the TS update.

src/modules/assetPack/sagas.ts
```diff
@@ -71,8 +71,9 @@ export function* assetPackSaga(builder: BuilderAPI) {
       yield put(setProgress(ProgressStage.UPLOAD_CONTENTS, 0))
 
       const updatableAssets = assetPack.assets.filter(asset => Object.keys(contents[asset.id]).length > 0)
-      const onProgress = yield handleAssetContentsUploadProgress(updatableAssets.length)
-      const uploadEffects = updatableAssets.map(asset => builder.saveAssetContents(asset, contents[asset.id]).then(onProgress))
+      const onProgress: void = yield handleAssetContentsUploadProgress(updatableAssets.length)
+      // TODO: there's a bug here
+      const uploadEffects = updatableAssets.map(asset => builder.saveAssetContents(asset, contents[asset.id]).then(onProgress as any))
```
LautaroPetaccio, line 76: This refers to the `onProgress as any` but I didn't have time to investigate it.

tsconfig.json
```diff
@@ -19,6 +19,7 @@
     "resolveJsonModule": true,
     "noUnusedParameters": true,
     "downlevelIteration": true,
+    "useUnknownInCatchVariables": false,
```
LautaroPetaccio, line 22: This is indeed a great TS feature, but it will require reviewing every exception and typing them correctly. This is out of the scope of this change.

package.json
```diff
@@ -170,7 +171,7 @@
     "ts-loader": "^6.2.1",
     "tslib": "^2.0.1",
     "tslint-plugin-prettier": "^2.0.1",
-    "typescript": "^4.0.2"
+    "typescript": "^4.5.4"
```
nicosantangelo, line 174: 👀

src/components/Modals/CreateMultipleItemsModal/CreateMultipleItemsModal.container.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Collection, CollectionType } from 'modules/collection/types'
+import { getCollection } from 'modules/collection/selectors'
+import { getCollectionType } from 'modules/collection/utils'
+import { MapStateProps, MapDispatchProps, MapDispatch, OwnProps } from './CreateMultipleItemsModal.types'
+import CreateMultipleItemsModal from './CreateMultipleItemsModal'
+
+const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
+  const collection: Collection | null = ownProps.metadata.collectionId ? getCollection(state, ownProps.metadata.collectionId) : null
+  const collectionUrn: string | null =
+    collection?.urn && getCollectionType(collection) === CollectionType.THIRD_PARTY ? collection.urn : null
```
nicosantangelo, line 15: I feel like this should be a method inside the component. My reasoning is that both collection types have URNs and there's no reason for a dev reading `collectionUrn` inside the component to derive that we're ONLY talking about TP URNs.
What do you think?

src/components/Modals/CreateMultipleItemsModal/CreateMultipleItemsModal.tsx
```diff
@@ -0,0 +1,415 @@
...
+  private isViewClosable = (): boolean => {
+    const { view } = this.state
+    return view === ItemCreationView.IMPORTING || view === ItemCreationView.COMPLETED
+  }
+
+  private getValidFiles = (): BuiltFile<Blob>[] => {
+    const { importedFiles } = this.state
+    return Object.values(importedFiles).filter(file => file.type === ImportedFileType.ACCEPTED) as BuiltFile<Blob>[]
+  }
+  private getRejectedFiles = (): RejectedFile[] => {
```
nicosantangelo, line 34:
```suggestion
  }
  
  private getRejectedFiles = (): RejectedFile[] => {
```

pls 👉 👈

src/components/Modals/CreateMultipleItemsModal/CreateMultipleItemsModal.tsx
```diff
@@ -0,0 +1,415 @@
...
+
+const REACT_APP_WEARABLES_ZIP_INFRA_URL = env.get('REACT_APP_WEARABLES_ZIP_INFRA_URL', '')
+export default class CreateMultipleItemsModal extends React.PureComponent<Props, State> {
+  state = {
+    view: ItemCreationView.IMPORT,
+    loadingFilesProgress: 0,
+    importedFiles: {} as Record<string, ImportedFile<Blob>>
+  }
+
+  private isViewClosable = (): boolean => {
```
nicosantangelo, line 25: Not sure: I think `isClosableView` is a bit more natural?

src/components/Modals/CreateMultipleItemsModal/CreateMultipleItemsModal.tsx
```diff
@@ -0,0 +1,415 @@
...
+            onDropAccepted={this.handleFilesImport}
+            onDropRejected={this.handleRejectedFiles}
+            acceptedExtensions={['.zip']}
+            moreInformation={
+              REACT_APP_WEARABLES_ZIP_INFRA_URL ? (
+                <span>
+                  <T
+                    id="create_multiple_items_modal.import_information"
+                    values={{
+                      link: <a href={REACT_APP_WEARABLES_ZIP_INFRA_URL}>{t('create_multiple_items_modal.import_information_link_label')}</a>
```
nicosantangelo, line 294: we could add `rel="noopener noreferrer"` to appease the React Warning gods. Also, maybe a `target="_blank"` ?

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.container.ts
```diff
@@ -4,14 +4,24 @@ import { isLoadingType } from 'decentraland-dapps/dist/modules/loading/selectors
...
 
-const mapState = (state: RootState): MapStateProps => ({
-  address: getAddress(state),
-  error: getError(state),
-  isLoading: isLoadingType(getLoading(state), SAVE_ITEM_REQUEST)
-})
+const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => {
+  const collection: Collection | null = ownProps.metadata.collectionId ? getCollection(state, ownProps.metadata.collectionId) : null
+  const collectionUrn: string | null =
+    collection?.urn && getCollectionType(collection) === CollectionType.THIRD_PARTY ? collection.urn : null
```
nicosantangelo, line 16: Same comment for this variable as for [CreateMultipleItemsModal.container.ts](https://github.com/decentraland/builder/pull/1776/files#diff-75031743d9ed991a0b5782130bf9e81ef4a2403a3a6f2271164ae6a6390fb80a). Plus if you check [here](https://github.com/decentraland/builder/pull/1776/files#diff-14a508dde6f9bf5af42a88ac3c6859b51313837d30291f7c8cee96199399e024R189) it's already confusing as the `isThirdParty` check should be unnecessary

src/lib/debounce.ts
```diff
@@ -29,7 +29,7 @@ export const debounceByKey = <T extends (...args: any[]) => any>(fn: T, ms: numb
           await fn(...args)
           promise.resolve()
         } catch (e) {
-          promise.reject(e.message)
+          promise.reject((e as any).message)
```
nicosantangelo, line 32: This is making me question types lol. Promise.reject (from futures) says that it takes an Error not a string. Not sure if we should change this line to:
`promise.reject((e as Error))`

@cazala @menduz halp

src/lib/api/conversions.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (remoteItem.price) item.price = remoteItem.price
+  if (remoteItem.urn) item.urn = remoteItem.urn
+  if (remoteItem.beneficiary) item.beneficiary = remoteItem.beneficiary
+  if (remoteItem.rarity) item.rarity = rarityToItemRarity(remoteItem.rarity)
+  if (remoteItem.total_supply !== null) item.totalSupply = remoteItem.total_supply // 0 is false
+
+  return item
+}
+
+function rarityToItemRarity(rarity: Rarity): ItemRarity {
```
nicosantangelo, line 35: What if we add a verb or perposition at the beginning of this method to show intent? `fromRarityToItemRarity`, `(convert/transform)RarityToItemRarity`?

src/lib/api/conversions.ts
```diff
@@ -0,0 +1,54 @@
+import { RemoteItem } from '@dcl/builder-client'
```
nicosantangelo, line 1: Are we not moving the rest of the (from/to) methods here to reduce the surface of the PR? If so, maybe add a comment ?

src/modules/common/store.ts
```diff
@@ -34,6 +35,9 @@ import { BuilderAPI, BUILDER_SERVER_URL } from 'lib/api/builder'
 import { Authorization } from 'lib/api/auth'
 import { PEER_URL } from 'lib/api/peer'
 
+import { getData } from 'modules/identity/selectors'
+import { getAddress } from 'decentraland-dapps/dist/modules/wallet/selectors'
```
nicosantangelo, line 39: Maybe this should be higher?

src/modules/common/store.ts
nicosantangelo: What do you think of adding a comment above this line explaining the situation?

src/modules/item/actions.ts
```diff
@@ -64,6 +64,32 @@ export type SaveItemRequestAction = ReturnType<typeof saveItemRequest>
...
 export type SaveItemFailureAction = ReturnType<typeof saveItemFailure>
 
+// Save multiple items
+
+export const SAVE_MULTIPLE_ITEMS_REQUEST = '[Request] Save Multiple Items'
+export const SAVE_MULTIPLE_ITEMS_SUCCESS = '[Success] Save Multiple Items'
+export const SAVE_MULTIPLE_ITEMS_FAILURE = '[Failure] Save Multiple Items'
+export const SAVE_MULTIPLE_ITEMS_CANCELLED = '[Cancelled] Save Multiple Items'
+export const SAVE_MULTIPLE_ITEMS_CANCEL = '[Cancel] Save Multiple Items'
+export const SAVE_MULTIPLE_ITEMS_CLEAR_STATE = '[Clear State] Save Multiple Items'
```
nicosantangelo, line 74: I see the value of having these two actions start with the same prefix as the other parts of the flow. BUT I think it's a bit misleading. They're not part of the async flow of saving items, so maybe starting with the verb is more obvious?

```ts
export const CANCEL_SAVE_MULTIPLE_ITEMS = '[Cancel] Save Multiple Items'
export const CLEAR_SAVE_MULTIPLE_ITEMS = '[Clear] Save Multiple Items'
```

I've purposefully omitted `state` there. I think that, _maybe_, it's not necessary

The types and actions are already following this pattern btw

src/modules/item/actions.ts
```diff
@@ -64,6 +64,32 @@ export type SaveItemRequestAction = ReturnType<typeof saveItemRequest>
...
+
+export const saveMultipleItemsRequest = (builtFiles: BuiltFile<Blob>[]) => action(SAVE_MULTIPLE_ITEMS_REQUEST, { builtFiles })
+export const saveMultipleItemsFailure = (error: string, items: Item[], fileNames: string[]) =>
+  action(SAVE_MULTIPLE_ITEMS_FAILURE, { error, items, fileNames })
+export const saveMultipleItemsCancelled = (items: Item[], fileNames: string[]) =>
+  action(SAVE_MULTIPLE_ITEMS_CANCELLED, { items, fileNames })
+export const saveMultipleItemsSuccess = (items: Item[], fileNames: string[]) => action(SAVE_MULTIPLE_ITEMS_SUCCESS, { items, fileNames })
+
+export const cancelSaveMultipleItems = () => action(SAVE_MULTIPLE_ITEMS_CANCEL)
+export const clearStateSaveMultipleItems = () => action(SAVE_MULTIPLE_ITEMS_CLEAR_STATE)
```
nicosantangelo, line 84: following the comment above, the first action is well named, the second could maybe drop `State`

src/modules/item/sagas.ts
```diff
@@ -90,7 +116,26 @@ import { getData as getItemsById, getItems, getEntityByItemId, getCollectionItem
 import { ItemTooBigError } from './errors'
 import { buildZipContents, getMetadata, isValidText, MAX_FILE_SIZE, toThirdPartyContractItems } from './utils'
 
-export function* itemSaga(builder: BuilderAPI) {
+const takeLatestCancellable = <A extends Action>(
```
nicosantangelo, line 119: I feel like this should maybe live in `common/sagas.ts` ? (looks really cool btw)

src/lib/api/conversions.ts
nicosantangelo, line 1: What do you think of renaming this file to `transformations` or something of the sort? conversions is usually used for units of measurements or monetary values

nicosantangelo (changes requested):
image

src/modules/media/reducer.ts
```diff
@@ -32,11 +32,12 @@ export const mediaReducer = (state = INITIAL_STATE, action: MediaReducerAction):
       return {
         ...state,
         media: {
-          north: window.URL.createObjectURL(north),
-          east: window.URL.createObjectURL(east),
-          south: window.URL.createObjectURL(south),
-          west: window.URL.createObjectURL(west),
-          preview: window.URL.createObjectURL(preview)
+          // TODO: is this null?
```
nicosantangelo, line 35: i dunno, maybe? 😛

src/components/ItemEditorPage/RightPanel/RightPanel.tsx
```diff
@@ -401,14 +402,16 @@ export default class RightPanel extends React.PureComponent<Props, State> {
...
                         />
-                        <Select<ItemRarity>
-                          itemId={item.id}
-                          label={t('global.rarity')}
-                          value={rarity}
-                          options={this.asRaritySelect(rarities)}
-                          disabled={item.isPublished || !canEditItemMetadata}
-                          onChange={this.handleChangeRarity}
-                        />
+                        {!(item.urn && isThirdParty(item.urn)) && (
```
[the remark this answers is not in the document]
LautaroPetaccio: It's pretty subjective, I usually prefer the and logic, even if I have to negate it all.
Changed it!

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.container.ts
LautaroPetaccio: Changed it to be the same as as the `CreateMultipleItemModal` one.

src/lib/debounce.ts
menduz: you should do `promise.reject(e)`, `e.message` is string, `e is Error`

src/components/ItemEditorPage/RightPanel/RightPanel.tsx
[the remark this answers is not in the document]
LautaroPetaccio: The name of the function `isThirdParty` refers to `isURNThirdParty`, so semantically I do prefer it to not receive a null string, as that null string doesn't mean anything for the function to check if an URN is third party or not. Saying that, it's not a bad idea to have a helper that receives either a collection or an item and checks if it is a third party item or collection based on their URN, taking into consideration if the URN is null.

src/components/Modals/CreateMultipleItemsModal/CreateMultipleItemsModal.tsx
nicosantangelo: Can neither confirm nor deny. Thus, dismissed

src/lib/api/conversions.ts
nicosantangelo: Cool!

src/lib/debounce.ts
LautaroPetaccio: Changed it to use `promise.reject(e)`

nicosantangelo (approved):

PR #1782, opened 2022-02-10
feat: Retrieve and show available slots for a Third Party Collection

### LINKED ISSUE #1748: Compute the remaining slots amount
Change the way we're computing the remaining item slots by using the new endpoint to get the amount of free tiers available in the builder server.
Remove the tiers module.

juanmahidalgo:
Closes #1748 

## Description

Consume the new endpoint from the `builder-server` and show the remaining available slots in the existing button

## Screenshot

image


src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
```diff
@@ -201,15 +207,19 @@ export default class ThirdPartyCollectionDetailPage extends React.PureComponent<
...
+                      className={classNames({ empty: areSlotsEmpty && !isLoadingAvailableSlots, slots: !isLoadingAvailableSlots })}
+                      onClick={this.handleBuySlot}
+                    >
+                      <>
+                        {t('third_party_collection_detail_page.slots', { amount: thirdParty?.availableSlots })}
+                        {areSlotsEmpty ? <span className="buy-slots link">{t('global.buy')}</span> : null}
+                      </>
                     </Button>
-                    <CollectionPublishButton collection={collection} items={selectedItems} slots={slots.toNumber()} />
+                    <CollectionPublishButton collection={collection} items={selectedItems} slots={thirdParty?.availableSlots || 0} />
```
nicosantangelo, line 222: is the third party always going to be defined in this component? couldn't this blow up otherwise?
Same goes for [PublishThirdPartyCollectionModal.tsx](https://github.com/decentraland/builder/pull/1782/files#diff-b480305804f9a96c5cd57aadd170b311245effbfa853a5ff6b0233a26a5fc2d2)

src/lib/api/builder.ts
```diff
@@ -704,6 +704,10 @@ export class BuilderAPI extends BaseAPI {
     return this.request('get', '/thirdParties', { manager })
   }
 
+  fetchThirdPartyAvailableSlots(thirdPartyId: string): Promise<number> {
+    return this.request('get', `/thirdParties/${thirdPartyId}/slots`)
```
nicosantangelo, line 708: Question! I'm actually not sure. Does this return an actual number? meaning isn't it a string?

src/modules/thirdParty/types.ts
```diff
@@ -5,4 +5,5 @@ export type ThirdParty = {
   description: string
   maxItems: string
   totalItems: string
+  availableSlots?: number
```
nicosantangelo, line 8: Question!
Does it make sense to have `availableSlots` as a `number | undefined`? couldn't it be zero if it's not defined?

Again, not sure, just asking

nicosantangelo (commented):
Great jobu

src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
juanmahidalgo: That's a good question, it's not that clear in this component if it's always going to be defined. There could be a race condition where the `collection` has been fetch but the `thirdParties` request has not been resolved and thus the `thirdParty` might still be `null`.

I made the following changes to address that:

In the `render` method, before calling `this.renderPage`, I've added the check if `thirdParty` is defined and I'm passing that down to the method instead of reading it from props (to avoid it being typed as `| null`)
```
{shouldRender && thirdParty ? this.renderPage(thirdParty) : <NotFound />}
```
and also moved the `getCollectionType(collection) === CollectionType.THIRD_PARTY` to the `.contatiner` file since I think that having a `thirdParty` object defined (in the render method) covers that case and it would be better to not call the `getCollectionThirdParty` selector (in the container) at all if the collection is not a TP one.

------


For [PublishThirdPartyCollectionModal.tsx](https://github.com/decentraland/builder/pull/1782/files#diff-b480305804f9a96c5cd57aadd170b311245effbfa853a5ff6b0233a26a5fc2d2) I think we're OK because in order to get that modal opened, we need to render the detail page first (and so `thirdParty` would be defined).

src/lib/api/builder.ts
juanmahidalgo: I believe it's coming as a `number` from the request:

image

src/modules/thirdParty/types.ts
juanmahidalgo: Mhmm.. I'm not sure neither. 
I think `availableSlots?: number` is transformed to `availableSlots: number | undefined` tho, so it would be the same.

src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
nicosantangelo: Perfect on both fronts! thanks for the detailed explanation

src/lib/api/builder.ts
nicosantangelo: Awesome then!

nicosantangelo (approved):

src/modules/thirdParty/types.ts
juanmahidalgo: thanks for the suggestion Nico! I've tried that and as you said, it gets a bit difficult to know if the `avaiableSlots` is truly 0 or if they weren't fetch yet. I'm merging it for now, we can come back later to this if we need :)

PR #1790, opened 2022-02-14
feat: Add shortened contract address to title when title is duplicated

fzavalia:
When having 2 collections with the same name but with different whitespaces in the middle (For example: "Hello World" and "Hello   World") the user could publish them but the second one would fail when the forum post is created because the forum removes extra white spaces in the title. This causes both collections to end up with the same name. The second one would end up failing because it would attempt to post a title that already exists.

The solution to this is that whenever this happens, append a shortened version of the collection address to the title to avoid duplication.


src/modules/forum/sagas.ts
```diff
@@ -20,17 +21,31 @@ export function* forumSaga(builder: BuilderAPI) {
     const { collection, forumPost } = action.payload
     try {
       const forumLink: string = yield call([builder, 'createCollectionForumPost'], collection, forumPost)
-      return put(createCollectionForumPostSuccess(collection, forumLink))
+      yield put(createCollectionForumPostSuccess(collection, forumLink))
```
LautaroPetaccio, line 24: Although yield works here, return is correct, as it's the last step of an iterator.

src/modules/forum/sagas.ts
```diff
@@ -20,17 +21,31 @@ export function* forumSaga(builder: BuilderAPI) {
...
     } catch (error) {
       if (builder.isAxiosError(error) && error.response) {
         const forumPostAlreadyExists = error.response.status === 409
         const forumLink = error.response.data?.data?.forum_link
+
         if (forumPostAlreadyExists && forumLink) {
-          return put(createCollectionForumPostSuccess(collection, forumLink))
+          yield put(createCollectionForumPostSuccess(collection, forumLink))
```
LautaroPetaccio, line 31: Return should work here:
```suggestion
          return put(createCollectionForumPostSuccess(collection, forumLink))
```

src/modules/forum/sagas.ts
```diff
@@ -20,17 +21,31 @@ export function* forumSaga(builder: BuilderAPI) {
...
+          return
+        }
+
+        const duplicatedTitle = error.response.status === 500 && error.response.data?.data?.errors?.includes('Title has already been used')
+
+        if (duplicatedTitle) {
+          yield put(
+            createCollectionForumPostRequest(collection, {
+              ...forumPost,
+              title: `${forumPost.title} ${shorten(collection.contractAddress!)}`
```
LautaroPetaccio, line 41: Would it be possible to exceed the length of the title?

LautaroPetaccio (changes requested):

src/modules/forum/sagas.ts
fzavalia: On the added case, `return put` was not working as it was not being handled. Had to write it as `yield put; return` for it to work. Might be a thing with sagas. I used yield everywhere to have a consistent usage in the saga.

src/modules/forum/sagas.ts
fzavalia: Same reason as https://github.com/decentraland/builder/pull/1790/files#r806119241

LautaroPetaccio (approved):

src/modules/forum/sagas.ts
fzavalia: It would be possible, maximum length is 255 characters,

PR #1791, opened 2022-02-14
feat: Introduce Item Curations module

### LINKED ISSUE #1725: Fetch item curations
The current implementation only cares about CollectionCurations, but Third Party collections have an ItemCuration for each item, and a main (virtual) CollectionCuration.

TODO:
- [x] When accessing any type of collection detail, if it's a TP collection, fetch it's itemCurations with `/items/:id/curation`
- [x] Rename the state so `curations` are called `collectionCurations` (or something of the sort)
- [x] Add a `itemCuration` section to the state
- [x] Fill up the state with the new data

Once we do that, we need to update:
- [ ] `ItemStatus` so it's TP enabled
- [x] Update `item/selectors.ts` so `getStatusByItemId` is TP enabled. Something like:
```ts
} else if (item.collectionId && (curationsByCollectionId[item.collectionId]?.status === CurationStatus.PENDING && !isTPW) || (curationByItemId[item.id]?.status === CurationStatus.PENDING && isTPW)  ) {
```

Check any other instance where we need to update the use of itemCurations, like when reviewing a TP Collection.

juanmahidalgo:
## Description
Things included in this PR:
- Introduces the `ItemCuration` module
- Fetch the itemCurations when loading a TP collection detail page
- Update the code to the new distinction of collection and item curations

Closes #1725


src/modules/curations/selectors.ts
```diff
@@ -0,0 +1,29 @@
...
+import { Collection } from 'modules/collection/types'
+import { RootState } from 'modules/common/types'
+import { createSelector } from 'reselect'
+import { getCurationsByCollectionId } from './collectionCuration/selectors'
+import { CollectionCuration } from './collectionCuration/types'
+import { getItemCurationsByCollectionId } from './itemCuration/selectors'
+import { ItemCuration } from './itemCuration/types'
+import { CurationStatus } from './types'
+
+export const getHasPendingCuration = createSelector<
```
juanmahidalgo, line 11: WIP: Not sure if this "generic" method will be used or not. Since the detail page are different, we might not need a generic method since we already know if the collection is TP or not.

src/modules/collection/selectors.ts
```diff
@@ -105,10 +108,22 @@ export const getStatusByCollectionId = createSelector<
         }
       }
     }
-    return statusByCollectionId
+    return statusByCollectionId[collectionId]
```
juanmahidalgo, line 111: I'm updating the selector to return the status for a `collectionId` instead of a map that has to be accessed from the `container`. AFAIK, this is still going to be memoized if the inputs don't change.

src/modules/collection/selectors.ts
```diff
@@ -105,10 +108,22 @@ export const getStatusByCollectionId = createSelector<
...
   }
 )
 
+export const isThirdPartyCollection = createSelector<
```
juanmahidalgo, line 115: I thought it'd be nice to have a selector that memoizes the validation if a collection is a TP based on a specific `collectionId`.

src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
```diff
@@ -48,6 +48,19 @@ export default class ThirdPartyCollectionDetailPage extends React.PureComponent<
     isAuthModalOpen: false
   }
 
+  componentDidUpdate() {
+    const { collection, itemCurations, onFetchItemCurations, isLoadingItemCurations, itemCurationsError } = this.props
+    if (
```
juanmahidalgo, line 53: I'm not sold on this implementation but I'm not convinced on moving the fetching logic to a sagas neither (like one listening to the route changes). 
@LautaroPetaccio @nicosantangelo Any thoughts about where to put this logic guys?

src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
```diff
@@ -47,9 +47,30 @@ export default class ThirdPartyCollectionDetailPage extends React.PureComponent<
     isAuthModalOpen: false
   }
 
+  componentDidMount() {
```
juanmahidalgo, line 50: Not related to the issue, I'm adding this here as a driveby. We should also fetch the available slots on the `componentDidMount` and not only in the `componentDidUpdate` as it's doing atm. 
The `didMount` is necessary when the user comes from the collection list, in that case, the `didUpdate` is not triggered and so the `onFetchAvailableSlots` is never called.

src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
nicosantangelo: tbh I think this is the correct place to do this. The improvement I can come up with is just doing

`if(this.shouldFetchCurations())`

and hide the condition there

src/modules/collection/selectors.ts
nicosantangelo: This seems cool! I'm not entirely sure how reselect handles these cases tho, part of me thinks that's not memoizing but I'm not sure

@cazala @LautaroPetaccio halp

src/modules/collection/selectors.ts
nicosantangelo: I think I commented this somewhere else, sorry if I'm repeating myself.

I don't think this is a bad idea, but it goes against a (admittedly, kinda implicit) convention where the ByXXX selectors return the entire memoized map and let the caller decide which prop to use. I think this was related to how reselect memoizes data and I couldn't find it on the standards (whoops 😅 )

Maybe @cazala can help a bit more here

cazala: same answer as here: https://github.com/decentraland/builder/pull/1791/files/c9a798938003419da21d9a20fb7baac601797781#r808270890

usually we memoize the whole map and then access the map by ID, I dunno how does reselect behave when passing other arguments other than state, if this is still being memoize and not computing more than once of each id then i guess i have no objection on doing it this way, we can verify that by adding a console log of the collection id and checking what comes up

src/modules/collection/selectors.ts
juanmahidalgo: Nevermind guys 😛 , I wrote that selector when I thought the only way to know if an item was a TPW was through the collection so now that I'm using the `item.urn`, this is not longer needed.

To Juan's question tho, I was thinking of using the selector concatenated with other selectors in the `getStatusByItemId` one. Something like:
```
  state => getItems(state),
  state => getEntityByItemId(state),
  state => getCurationsByCollectionId(state),
  (_state, collectionId) => isThirdPartyCollection(collectionId),
  getItemCurationsByItemId,
```

src/modules/item/selectors.ts
```diff
@@ -94,35 +96,57 @@ export const getEntityByItemId = createSelector<RootState, Entity[], Record<stri
...
+  let status = SyncStatus.UNSYNCED
+  if (!entity) {
+    status = SyncStatus.LOADING
+  } else if (areSynced(item, entity)) {
+    status = SyncStatus.SYNCED
+  }
+  return status
+}
+
+const getStatusForTPW = (item: Item, itemCuration: ItemCuration | null, entity: Entity): SyncStatus => {
```
nicosantangelo, line 109: Nitpick but can we drop the W? `getStatusForTP`

this is probably not the only place where we use TPW but it's kinda wrong as the W stands for Wearable and it's a bit weird when we're managing Collections and Items. Third Party (TP) seems a bit better.

This is a mea-culpa as we started this mess with Lautaro lol

src/modules/item/selectors.ts
```diff
@@ -94,35 +96,57 @@ export const getEntityByItemId = createSelector<RootState, Entity[], Record<stri
...
+  let status: SyncStatus
+  if (!item.isPublished) {
+    status = SyncStatus.UNPUBLISHED
+  } else if (!item.isApproved || (collectionCuration && collectionCuration.status === CurationStatus.PENDING)) {
+    status = SyncStatus.UNDER_REVIEW
+  } else {
+    status = getItemSyncedStatus(item, entity)
+  }
+  return status
+}
```
nicosantangelo, line 130: following the current implementation in `master`

```ts
  if (!item.isPublished) {
    statusByItemId[item.id] = SyncStatus.UNPUBLISHED
  } else if (!item.isApproved) {
    statusByItemId[item.id] = SyncStatus.UNDER_REVIEW
  } else {
    const entity = entitiesByItemId[item.id]
    if (!entity) {
      statusByItemId[item.id] = SyncStatus.LOADING
    } else if (areSynced(item, entity)) {
      statusByItemId[item.id] = SyncStatus.SYNCED
    } else if (item.collectionId && curationsByCollectionId[item.collectionId]?.status === CurationStatus.PENDING) {
      statusByItemId[item.id] = SyncStatus.UNDER_REVIEW
    } else {
      statusByItemId[item.id] = SyncStatus.UNSYNCED
    }
  }
```

The original implementation used to favor LOADING over under_review for what I can see. Meaning that if `isApproved = true` and no entity was found the Status'd be LOADING but here could be UNDER_REVIEW If there's a collection status Pending right?

src/modules/item/selectors.ts
```diff
@@ -94,35 +96,57 @@ export const getEntityByItemId = createSelector<RootState, Entity[], Record<stri
...
+    status = SyncStatus.SYNCED
+  }
+  return status
+}
+
+const getStatusForTPW = (item: Item, itemCuration: ItemCuration | null, entity: Entity): SyncStatus => {
+  let status: SyncStatus
+  if (!itemCuration) {
+    status = SyncStatus.UNPUBLISHED
+  } else if (itemCuration.status === CurationStatus.PENDING) return SyncStatus.UNDER_REVIEW
```
nicosantangelo, line 113: the formatting here looks a bit weird without `{}` no? maybe we can add them

nicosantangelo (commented):
Awesome!

We need to try the memoization (as [juan suggested here](https://github.com/decentraland/builder/pull/1791#discussion_r808275357) ) but other than that great work

src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
juanmahidalgo: good idea! Updated to something similar to that :)

src/modules/curations/selectors.ts
juanmahidalgo: I ended up deleting this file, it wasn't necessary. It's a good idea tho, if at some point we need something similar to that selector, we can come back to this.

src/modules/item/selectors.ts
juanmahidalgo: True, that's right. In case there's no `entity`, the current code in `master` returns `LOADING` even after checking if there are curations or not for the collection (talking specifically about DCL collections).
I guess we gotta decide the precedence here, what's more relevant? Or, what should we check first? If it has an `entity` or a pending curation?

src/modules/item/selectors.ts
juanmahidalgo: Glad you asked that 😂 .`W` dropped!

src/modules/item/selectors.ts
juanmahidalgo: Absolutely :)

src/modules/collection/selectors.ts
```diff
@@ -109,6 +109,18 @@ export const getStatusByCollectionId = createSelector<
...
+  RootState,
+  Collection['id'] | undefined,
+  CollectionState['data'],
+  Collection['id'] | undefined,
+  boolean
+>(
+  getData,
+  (_: RootState, collectionId?: string) => collectionId,
+  (data, collectionId?: Collection['id']) => !!collectionId && !!isThirdParty(data[collectionId]?.urn)
```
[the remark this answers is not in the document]
juanmahidalgo: I _think_ that piece of code is gone, we don't need that selector anymore, I think you were checking an outdated version.

nicosantangelo (approved):

PR #1795, opened 2022-02-16
feat: Prevent collectionCard drop if it's a TP

### LINKED ISSUE #1793: Prevent items without collections to be dropped into third party collections
In the UI, there's the possibility of moving items without collections to collections.
The Third Party collections won't accept that.
To complete this task, prevent items from being dropped or moved into Third Party collections.

juanmahidalgo:
Closes #1793 

Please ignore the broken images, it's all mocked in my local environment 😂 

https://user-images.githubusercontent.com/8763687/154249607-d14490b7-9a39-4785-a732-540dae13c3cb.mov


src/components/CollectionsPage/CollectionCard/CollectionCard.dnd.ts
```diff
@@ -14,7 +15,7 @@ export const collectionTarget: DropTargetSpec<Props> = {
     props.onSetCollection(item, props.collection.id)
   },
   canDrop(props) {
-    return !props.collection.isPublished
+    return !props.collection.isPublished && !isThirdParty(props.collection.urn)
```
nicosantangelo, line 18: this is 100% correct, but I think it'd be a good idea to standardize how we check for TP. Looking at the codebase:

- Most of the code is using `getCollectionType(collection)`. I think the method is a bit better, as it lets us add new collection types in the future, the downside is needing the entire collection to decide, but we usually have it  (or we can scope the type). Maybe we can change it here?
  - There's another rogue isThirdParty(collection.urn) on `hasViewAndEditRights` but we can change it later
- Items do not have an ItemType (yet) so it's correct to use `isThirdParty` and refactor later

Tell me what you think

nicosantangelo (commented):
Good work, refactor comment

src/components/CollectionsPage/CollectionCard/CollectionCard.dnd.ts
juanmahidalgo: Yeah, I like the idea of standardising but I don't dislike the idea of having a more specific helper to check if it's specifically a third party one instead of having to import the enum and checking it manually in each component. 

This component has the entire `collection` object so we can use the `getCollectionType(collection)` without problem. We could write another one that receives just the `urn` but I think that if we have the `urn` we probably have the entire collection at that point, so it might not be worth it.
-----
Code updated to use `getCollectionType(collection)`!

nicosantangelo (approved):

PR #1800, opened 2022-02-18
feat: Block Item/Collection URN editing when published

### LINKED ISSUE #1768: Block the editing of URNs for items and collections if they were published
Items and collections can't change their URN if they're published.
- A collection is published if any of their items are published.
- An item is published if it has a curation or if it's in the catalyst.

The server should return if the collection or the item is published.

To complete this task, add the require logic in the components to prevent the user from updating the URNs.

juanmahidalgo:
Closes #1768 

### Collection  Published

image

### Item published (with at least 1 curation)

image

### DIFF
--- a/src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.module.css
+++ b/src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.module.css
@@ -6,3 +6,11 @@
 .ellipsis {
   margin: auto !important;
 }
+
+:global(.ui.dropdown).dropdown :global(.menu > .disabled.item) {
+  pointer-events: auto;
+}
+
+:global(.ui.dropdown).dropdown :global(.menu > .disabled.item a) {
+  cursor: default;
+}
--- a/src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.module.css
+++ b/src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.module.css
@@ -7,6 +7,7 @@
   margin: auto !important;
 }
 
+/* Semantic-UI: https://github.com/Semantic-Org/Semantic-UI/issues/4285#issuecomment-232974723 */
 :global(.ui.dropdown).dropdown :global(.menu > .disabled.item) {
   pointer-events: auto;
 }
--- a/src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.tsx
+++ b/src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.tsx
@@ -24,7 +24,9 @@ export default class CollectionContextMenu extends React.PureComponent<Props> {
 
   handleEditURN = () => {
     const { collection, onOpenModal } = this.props
-    if (!collection.isPublished) onOpenModal('EditCollectionURNModal', { collection })
+    if (!collection.isPublished) {
+      onOpenModal('EditCollectionURNModal', { collection })
+    }
   }
 
   navigateTo = (url: string, target: string = '') => {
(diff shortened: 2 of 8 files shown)


src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.module.css
```diff
@@ -6,3 +6,11 @@
 .ellipsis {
   margin: auto !important;
 }
+
+:global(.ui.dropdown).dropdown :global(.menu > .disabled.item) {
+  pointer-events: auto;
```
juanmahidalgo, line 11: Without this, the `Popup` won't be triggered on a disabled element since the `pointer-events: none` applied on the element will prevent the `:hover` action and so the `Popup` won't be rendered.

src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.tsx
```diff
@@ -24,7 +24,7 @@ export default class CollectionContextMenu extends React.PureComponent<Props> {
 
   handleEditURN = () => {
     const { collection, onOpenModal } = this.props
-    onOpenModal('EditCollectionURNModal', { collection })
+    if (!collection.isPublished) onOpenModal('EditCollectionURNModal', { collection })
```
juanmahidalgo, line 27: The `pointer-events: auto;` allows the user to click the `Dropdown.Item`, so we need to check it manually here before opening the modal. Same for the `EditItemURNModal` modal.

nicosantangelo: 👍 
I know I'm annoying, but to be consistent with the rest of the codebase, would you mind adding `{}` to the if here?

src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.module.css
nicosantangelo: Just tried it, seems you're right!

Now, what about adding these two styles to the `/src/themes/components/Dropdown.css` file, which is the place where we add global, component-based, overrides/styles.
We can remove it from [src/components/ThirdPartyCollectionDetailPage/CollectionItem/CollectionItem.module.css](https://github.com/decentraland/builder/pull/1800/files#diff-ac267dde608f7eeefb22ccb486995c649a76939701ea60c081832101c8507b08) too, and any other place where this was necessary.

NOW, i'm aware that this might be a bit dangerous. We need to do a quick check to see nothing broke. If you're not comfortable with that. This is fine

src/components/ThirdPartyCollectionDetailPage/CollectionItem/CollectionItem.tsx
```diff
@@ -21,7 +21,7 @@ export default class CollectionItem extends React.PureComponent<Props> {
 
   handleEditURN = () => {
     const { item, onOpenModal } = this.props
-    onOpenModal('EditItemURNModal', { item })
+    if (!item.isPublished) onOpenModal('EditItemURNModal', { item })
```
nicosantangelo, line 24: Same as comment [here](https://github.com/decentraland/builder/pull/1800/files#r809825837)

nicosantangelo (commented):
🚀

src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.module.css
juanmahidalgo: Yeah good idea! I created a ticket to track that specifically and make sure it won't break any of the existing stuff :) 
#1810

LautaroPetaccio (approved):

PR #1814, opened 2022-02-28
remove MAX_PUBLISH_ITEM_COUNT

coder-abdo:
no need to add extra information that doesn't use anymore


juanmahidalgo:
Thanks for contributing, this will be fixed in #1819

PR #1820, opened 2022-03-04
feat: User servers' content hash

### LINKED ISSUE #1767: Generate the catalyst's deployment image when saving the item
To complete this task, the following must be done:
- [ ] Remove the image generation process in the catalyst's deployment.
- [ ] Add the removed image generation process to the item's content when upserting the item.

This shouldn't be a breaking change, the server should allow this kind of behavior.
If the item gets downloaded in the future, the deployment of the item in the catalyst should keep working.

LautaroPetaccio:
This PR does the following:
- Updates the builder-client to have the latests types on the `local_content_hash` property.
- Builds the catalyst image when saving the items to store them in the builder-server along its contents.
- Changes the way that the item statuses are computed by including the check between the content hash property (the content hash coming from the blockchain) and the server content hash, coming from the builder-server.
- Changes the way the content hash was computed if there's no server content hash by using the already loaded content hashes of the item and generating a new catalyst image, reducing the time to do this operation.

Closes #1767


src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
```diff
@@ -286,10 +288,16 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
...
-          updatedAt: +new Date()
+          updatedAt: +new Date(),
+          serverContentHash: null
         }
       }
 
+      // Add the catalyst image to the item
+      const catalystImage = await generateCatalystImage(item, { thumbnail: sortedContents.all[THUMBNAIL_PATH] })
+      sortedContents.all[IMAGE_PATH] = catalystImage.content
+      item.contents[IMAGE_PATH] = catalystImage.hash
```
nicosantangelo, line 299: Does it make sense to abstract these 3 lines ? they seem similar to these https://github.com/decentraland/builder/pull/1820/files#diff-58a4f4bd05bdec4de69216a6c54b669acd456a2f36a93f05d47a5315f53febb9R141 but I'm unsure if it's accidental duplication

src/lib/api/builder.ts
```diff
@@ -314,7 +315,8 @@ function toRemoteItem(item: Item): RemoteItem {
     contents: item.contents,
     content_hash: item.contentHash,
     created_at: new Date(item.createdAt),
-    updated_at: new Date(item.updatedAt)
+    updated_at: new Date(item.updatedAt),
+    local_content_hash: item.serverContentHash
```
nicosantangelo, line 319: maybe we want to move this prop (same [here](https://github.com/decentraland/builder/pull/1820/files#diff-9b32bce51ccd6629860aa36212c39f1a85965c2dccd90c6ec0d6c5d7bb8d0159R342)) below content_hash so it's a bit more "content aware"

src/modules/deployment/contentUtils.ts
```diff
@@ -7,25 +7,28 @@ export async function computeHashes(contents: Record<string, Blob>): Promise<Rec
...
     const blob = contents[path]
-    const file = await makeContentFile(path, blob)
-    const cid = await calculateBufferHash(file.content)
-    contentsAsHashes[path] = cid
+    contentsAsHashes[path] = await computeHashOfContent(blob)
   }
   return contentsAsHashes
 }
 
+export async function computeHashOfContent(content: Blob): Promise<string> {
```
nicosantangelo, line 15: maybe
```suggestion
export async function computeHashFromContent(content: Blob): Promise<string> {
```
?

src/modules/item/export.ts
```diff
@@ -111,29 +111,33 @@ export function buildItemEntityMetadata(collection: Collection, item: Item): Cat
...
+    const catalystItem = await generateCatalystImage(item)
+    contents[IMAGE_PATH] = catalystItem.hash
+  }
+
+  return contents
+}
+
+async function buildItemEntityBlobs(item: Item): Promise<Record<string, Blob>> {
+  const [files, image] = await Promise.all([getFiles(item.contents), !item.contents[IMAGE_PATH] ? generateImage(item) : null])
+  files[IMAGE_PATH] = image !== null ? image : files[IMAGE_PATH]
```
nicosantangelo, line 133: Maybe
```suggestion
  files[IMAGE_PATH] = image ?? files[IMAGE_PATH]
```
?

src/modules/item/types.ts
```diff
@@ -146,6 +146,7 @@ export type Item = BaseItem & {
   inCatalyst: boolean
   contents: Record<string, string>
   contentHash: string | null
+  serverContentHash: string | null
```
nicosantangelo, line 149: I was thinking we might be able to name this better. Both here and the server. **I wouldn't change it here**, but I feel like neither local nor server signify the real use of this prop.
Maybe cached_content_hash?

src/modules/item/utils.ts
```diff
@@ -378,6 +400,10 @@ export function areEqualRepresentations(a: WearableRepresentation[], b: Wearable
   return true
 }
 
+export function areSyncedByHash(item: Item): boolean {
+  return Boolean(item.serverContentHash) && item.contentHash === item.serverContentHash
```
nicosantangelo, line 404: This is completely fine, but the rest of the codebase is using `!!` in case you want to change it.
```suggestion
  return !!item.serverContentHash && item.contentHash === item.serverContentHash
```

again, is fine the way it is

nicosantangelo (commented):
Awesome

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
LautaroPetaccio: They seem similar, but they're generating an image and setting its content and hash to different objects. It might not add much to abstract that logic.

src/modules/item/types.ts
LautaroPetaccio: It's definitely a difficult property to name. I wouldn't go with `cached_content_hash`, because it's the content hash of the entity that's stored in the last save to the server. We could rename both content hashed to:
`blockchainContentHash` and `actualContentHash` or `currentContentHash`, what do you think about this?

src/modules/deployment/contentUtils.ts
LautaroPetaccio: Changed it!

src/lib/api/builder.ts
LautaroPetaccio: Moved it!

src/modules/item/utils.ts
LautaroPetaccio: Changed it!

src/lib/api/builder.ts
```diff
@@ -52,6 +52,7 @@ export type RemoteItem = {
   content_hash: string | null
   created_at: Date
   updated_at: Date
+  local_content_hash: string | null
```
nicosantangelo, line 55: You can move this one two lines up too!

src/lib/api/builder.ts
```diff
@@ -336,7 +338,8 @@ function fromRemoteItem(remoteItem: RemoteItem) {
     contentHash: remoteItem.content_hash,
     metrics: remoteItem.metrics,
     createdAt: +new Date(remoteItem.created_at),
-    updatedAt: +new Date(remoteItem.created_at)
+    updatedAt: +new Date(remoteItem.created_at),
+    serverContentHash: remoteItem.local_content_hash
```
nicosantangelo, line 342: You can move this one two lines up too!

nicosantangelo (approved):
I like `blockchainContentHash` and `currentContentHash` !

Approved anyways because the comments I left are just if you want to move it and we can rename afterwards

nicosantangelo (approved):

PR #1821, opened 2022-03-07
feat: Check items with the backend before publishing a collection

fzavalia:


src/modules/collection/selectors.ts
```diff
@@ -22,6 +22,14 @@ export const getData = (state: RootState) => getState(state).data
 export const getLoading = (state: RootState) => getState(state).loading
 export const getError = (state: RootState) => getState(state).error
 
+export const getUnsyncedCollectionError = (state: RootState) => {
```
LautaroPetaccio, line 25: Would you mind adding a test for this selector?

src/modules/translation/languages/en.json
```diff
@@ -1123,7 +1123,8 @@
     "content_policy": "Content Policy",
     "accept_by_publishing": "By clicking Publish you are accepting our terms",
     "not_enough_mana": "You don't have enough {symbol} to publish this collection.",
-    "get_mana": "You can get more using the {link} dapp."
+    "get_mana": "You can get more using the {link} dapp.",
+    "unsynced_collection": "The collection you are trying to publish is different than the one in the server. Refresh the browser and try again."
```
LautaroPetaccio, line 1127: What do you think about something like this?
```suggestion
    "unsynced_collection": "The collection you are trying to publish is different from the one we have in our servers. Please, refresh the browser and try again."
```

src/modules/collection/sagas.ts
```diff
@@ -257,6 +257,24 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
...
+      }
+
+      // TODO: Deeper comparison of browser and server items. Compare metadata for example.
+      serverItems.forEach(serverItem => {
+        const browserItem = items.find(item => item.id === serverItem.id)
+
+        if (!browserItem) {
+          throw new Error(`UnsyncedCollection: No matching item in browser`)
+        }
+      })
```
LautaroPetaccio, line 276: Would you mind adding tests that exercise these exceptions?

src/modules/collection/sagas.ts
```diff
@@ -257,6 +257,24 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
...
+      if (serverItems.length !== items.length) {
+        throw new Error(`UnsyncedCollection: Different items length`)
+      }
+
+      // TODO: Deeper comparison of browser and server items. Compare metadata for example.
+      serverItems.forEach(serverItem => {
+        const browserItem = items.find(item => item.id === serverItem.id)
+
+        if (!browserItem) {
+          throw new Error(`UnsyncedCollection: No matching item in browser`)
```
LautaroPetaccio, line 274: I would narrow the exception to just the existence of the item for now:
```suggestion
          throw new Error(`UnsyncedCollection: Item found in the server but not in the browser`)
```

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
```diff
@@ -214,6 +214,7 @@ export default class PublishCollectionModal extends React.PureComponent<Props, S
             {t('global.publish')}
           </Button>
           <p>{t('publish_collection_modal.accept_by_publishing')}</p>
+          {unsyncedCollectionError && <p className='error'>{t('publish_collection_modal.unsynced_collection')}</p>}
```
LautaroPetaccio, line 217: Should we also prevent the publish button from working when this error is found?

LautaroPetaccio (changes requested):

src/modules/collection/selectors.spec.ts
```diff
@@ -3,12 +3,47 @@ import { getChainIdByNetwork } from 'decentraland-dapps/dist/lib/eth'
...
+describe('when getting the unsynced error message', () => {
+  it('should return null if the base error is null', () => {
+    const result = getUnsyncedCollectionError({
+      collection: {
+        error: null
+      }
+    } as any)
+
+    expect(result).toBeNull()
+  })
```
LautaroPetaccio, line 22: Would you mind changing the tests so they fit with the [testing standards](https://www.notion.so/decentraland/Testing-standards-af326c6776d64d58857429583da41046#8c2d4a1bb0374a05853fc03083567004)?
That is, they should look like this:
```typescript
  let state: RootState
  describe('when the base error is null', () => {
    // We define an describe the state
    beforeEach(() => {
      result = {
        collection: {
          error: null
        }
      } as RootState
    })
  
    it('should return null', () => {
      const result = getUnsyncedCollectionError( as any)

      expect(result).toBeNull()
    })
  })
```

LautaroPetaccio (changes requested):

LautaroPetaccio (approved):

PR #1822, opened 2022-03-08
feat: Create the TP publish or push changes mechanism

### LINKED ISSUE #1752: Create the TP items publish or push changes mechanism
Create the sagas to publish or push changes of a set of TP items.
The sagas should receive an action with the IDs of the items to be published or (inclusive or) have their changes pushed.
This sagas should:
- Make the creator sign a message (a slots debit message) that includes:
  - The quantity of items to be published (items not published before taken into consideration the set of IDs received)
  - The salt value that identifies univocally this message.
  - The third party id.
- Make a POST request to `/collection/:id/publish` that will, knowing that is a TP collection, store the message and use the array of item IDs to create or update the curations for the items accordingly (if it's updating the item or it's publishing it).
- Update the local stats for the collection to show that the collection has items that are still not reviewed.
- Update or create the item curations with the curations returned from the server.
- Create a new publish modal showing a message that tells the user that the action is irreversible and that X item slots will be consumed after publishing.

Related https://github.com/decentraland/builder-server/issues/389

juanmahidalgo:
Closes #1752

### Description

Creates the sagas and the logic to publish and push changes of TP items.

- The publish flow will ask for the signature
image

- The push changes flow calls the endpoint and updates the current itemCuration o creates a new one if needed.

### DIFF
--- a/src/components/ActivityPage/Transaction/Transaction.tsx
+++ b/src/components/ActivityPage/Transaction/Transaction.tsx
@@ -19,7 +19,7 @@ import {
   DISSOLVE_ESTATE_SUCCESS,
   SET_UPDATE_MANAGER_SUCCESS
 } from 'modules/land/actions'
-import { PUBLISH_THIRD_PARTY_ITEMS_SUCCESS, RESCUE_ITEMS_CHUNK_SUCCESS, SET_PRICE_AND_BENEFICIARY_REQUEST } from 'modules/item/actions'
+import { RESCUE_ITEMS_CHUNK_SUCCESS, SET_PRICE_AND_BENEFICIARY_REQUEST } from 'modules/item/actions'
 import {
   MINT_COLLECTION_ITEMS_SUCCESS,
   SET_COLLECTION_MINTERS_SUCCESS,
@@ -373,24 +373,6 @@ const Transaction = (props: Props) => {
       )
     }
 
-    case PUBLISH_THIRD_PARTY_ITEMS_SUCCESS: {
-      const { collectionId, collectionName, items } = tx.payload
-      return (
-        <TransactionDetail
-          tx={tx}
-          collectionId={collectionId}
-          text={
-            <T
-              id="transaction.publish_third_party_items"
-              values={{
-                count: items.length,
-                collectionName: <Link to={locations.thirdPartyCollectionDetail(collectionId)}>{collectionName}</Link>
-              }}
-            />
-          }
-        />
-      )
-    }
     case BUY_THIRD_PARTY_ITEM_SLOT_SUCCESS: {
       const { thirdParty, slotsToyBuy } = tx.payload
       return (
--- a/src/lib/api/builder.ts
+++ b/src/lib/api/builder.ts
@@ -659,7 +659,7 @@ export class BuilderAPI extends BaseAPI {
     return fromRemoteCollection(remoteCollection)
   }
 
-  async publishCollection(collectionId: string) {
+  async publishDCLCollection(collectionId: string) {
     const { collection, items }: { collection: RemoteCollection; items: RemoteItem[] } = await this.request(
       'post',
       `/collections/${collectionId}/publish`
@@ -667,6 +667,29 @@ export class BuilderAPI extends BaseAPI {
     return {
       collection: fromRemoteCollection(collection),
       items: items.map(fromRemoteItem)
+    }
+  }
+
+  async publishCollection(collectionId: string, itemIds: string[], signedMessage: string, signature: string, qty: number, salt: string) {
+    const {
+      collection,
+      items,
+      itemCurations
+    }: { collection: RemoteCollection; items: RemoteItem[]; itemCurations: RemoteItemCuration[] } = await this.request(
+      'post',
+      `/collections/${collectionId}/publish`,
+      {
+        itemIds,
+        signedMessage,
+        signature,
+        qty,
+        salt
+      }
+    )
+    return {
+      collection: fromRemoteCollection(collection),
+      items: items.map(fromRemoteItem),
+      itemCurations: itemCurations.map(fromRemoteItemCuration)
     }
   }
 
@@ -716,6 +739,10 @@ export class BuilderAPI extends BaseAPI {
     return this.request('post', `/collections/${collectionId}/curation`)
   }
 
+  pushItemCuration(itemId: string): Promise<void> {
+    return this.request('post', `/items/${itemId}/curation`)
+  }
+
   fetchCommittee(): Promise<string[]> {
     return this.request('get', '/committee')
   }
@@ -738,6 +765,10 @@ export class BuilderAPI extends BaseAPI {
 
   updateCurationStatus(collectionId: string, status: CurationStatus): Promise<void> {
     return this.request('patch', `/collections/${collectionId}/curation`, { curation: { status } })
+  }
+
+  updateItemCurationStatus(itemId: string, status: CurationStatus): Promise<void> {
+    return this.request('patch', `/items/${itemId}/curation`, { curation: { status } })
   }
 
   isAxiosError(error: any): error is AxiosError {
--- a/src/lib/api/builder.ts
+++ b/src/lib/api/builder.ts
@@ -670,7 +670,7 @@ export class BuilderAPI extends BaseAPI {
     }
   }
 
-  async publishCollection(collectionId: string, itemIds: string[], signedMessage: string, signature: string, qty: number, salt: string) {
+  async publishTPCollection(collectionId: string, itemIds: string[], signedMessage: string, signature: string, qty: number, salt: string) {
     const {
       collection,
       items,
--- a/src/lib/api/builder.ts
+++ b/src/lib/api/builder.ts
@@ -680,10 +680,12 @@ export class BuilderAPI extends BaseAPI {
       `/collections/${collectionId}/publish`,
       {
         itemIds,
-        signedMessage,
-        signature,
-        qty,
-        salt
+        cheque: {
+          signedMessage,
+          signature,
+          qty,
+          salt
+        }
       }
     )
     return {
--- a/src/lib/api/builder.ts
+++ b/src/lib/api/builder.ts
@@ -659,7 +659,7 @@ export class BuilderAPI extends BaseAPI {
     return fromRemoteCollection(remoteCollection)
   }
 
-  async publishDCLCollection(collectionId: string) {
+  async publishStandardCollection(collectionId: string) {
     const { collection, items }: { collection: RemoteCollection; items: RemoteItem[] } = await this.request(
       'post',
       `/collections/${collectionId}/publish`
@@ -741,7 +741,7 @@ export class BuilderAPI extends BaseAPI {
     return this.request('post', `/collections/${collectionId}/curation`)
   }
 
-  pushItemCuration(itemId: string): Promise<void> {
+  pushItemCuration(itemId: string): Promise<ItemCuration> {
     return this.request('post', `/items/${itemId}/curation`)
   }
 
@@ -769,7 +769,7 @@ export class BuilderAPI extends BaseAPI {
     return this.request('patch', `/collections/${collectionId}/curation`, { curation: { status } })
   }
 
-  updateItemCurationStatus(itemId: string, status: CurationStatus): Promise<void> {
+  updateItemCurationStatus(itemId: string, status: CurationStatus): Promise<ItemCuration> {
     return this.request('patch', `/items/${itemId}/curation`, { curation: { status } })
   }
 
--- a/src/lib/api/builder.ts
+++ b/src/lib/api/builder.ts
@@ -52,6 +52,7 @@ export type RemoteItem = {
   content_hash: string | null
   created_at: Date
   updated_at: Date
+  local_content_hash: string | null
 }
 
 export type RemoteCollection = {
@@ -312,7 +313,8 @@ function toRemoteItem(item: Item): RemoteItem {
     data: item.data,
     metrics: item.metrics,
     contents: item.contents,
-    content_hash: item.contentHash,
+    content_hash: item.blockchainContentHash,
+    local_content_hash: item.currentContentHash,
     created_at: new Date(item.createdAt),
     updated_at: new Date(item.updatedAt)
   }
@@ -333,7 +335,8 @@ function fromRemoteItem(remoteItem: RemoteItem) {
     type: remoteItem.type,
     data: remoteItem.data,
     contents: remoteItem.contents,
-    contentHash: remoteItem.content_hash,
+    currentContentHash: remoteItem.local_content_hash,
+    blockchainContentHash: remoteItem.content_hash,
     metrics: remoteItem.metrics,
     createdAt: +new Date(remoteItem.created_at),
     updatedAt: +new Date(remoteItem.created_at)
--- a/src/lib/api/builder.ts
+++ b/src/lib/api/builder.ts
@@ -744,8 +744,10 @@ export class BuilderAPI extends BaseAPI {
     return this.request('post', `/collections/${collectionId}/curation`)
   }
 
-  pushItemCuration(itemId: string): Promise<ItemCuration> {
-    return this.request('post', `/items/${itemId}/curation`)
+  async pushItemCuration(itemId: string): Promise<ItemCuration> {
+    const curation: RemoteItemCuration = await this.request('post', `/items/${itemId}/curation`)
+
+    return fromRemoteItemCuration(curation)
   }
 
   fetchCommittee(): Promise<string[]> {
@@ -772,8 +774,9 @@ export class BuilderAPI extends BaseAPI {
     return this.request('patch', `/collections/${collectionId}/curation`, { curation: { status } })
   }
 
-  updateItemCurationStatus(itemId: string, status: CurationStatus): Promise<ItemCuration> {
-    return this.request('patch', `/items/${itemId}/curation`, { curation: { status } })
+  async updateItemCurationStatus(itemId: string, status: CurationStatus): Promise<ItemCuration> {
+    const curation: RemoteItemCuration = await this.request('patch', `/items/${itemId}/curation`, { curation: { status } })
+    return fromRemoteItemCuration(curation)
   }
 
   isAxiosError(error: any): error is AxiosError {
--- a/src/modules/curations/itemCuration/reducer.ts
+++ b/src/modules/curations/itemCuration/reducer.ts
@@ -1,5 +1,18 @@
 import { loadingReducer, LoadingState } from 'decentraland-dapps/dist/modules/loading/reducer'
 import {
+  PublishAndPushChangesThirdPartyItemsFailureAction,
+  PublishAndPushChangesThirdPartyItemsSuccessAction,
+  PublishThirdPartyItemsFailureAction,
+  PublishThirdPartyItemsSuccessAction,
+  PUBLISH_AND_PUSH_CHANGES_THIRD_PARTY_ITEMS_FAILURE,
+  PUBLISH_THIRD_PARTY_ITEMS_FAILURE,
+  PUBLISH_THIRD_PARTY_ITEMS_SUCCESS,
+  PushChangesThirdPartyItemsFailureAction,
+  PushChangesThirdPartyItemsSuccessAction,
+  PUSH_CHANGES_THIRD_PARTY_ITEMS_FAILURE,
+  PUSH_CHANGES_THIRD_PARTY_ITEMS_SUCCESS
+} from 'modules/thirdParty/actions'
+import {
   FETCH_ITEM_CURATIONS_REQUEST,
   FETCH_ITEM_CURATIONS_SUCCESS,
   FETCH_ITEM_CURATIONS_FAILURE,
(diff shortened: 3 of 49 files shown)


src/components/ActivityPage/Transaction/Transaction.tsx
```diff
@@ -373,24 +373,6 @@ const Transaction = (props: Props) => {
...
 
-    case PUBLISH_THIRD_PARTY_ITEMS_SUCCESS: {
-      const { collectionId, collectionName, items } = tx.payload
-      return (
-        <TransactionDetail
-          tx={tx}
-          collectionId={collectionId}
-          text={
-            <T
-              id="transaction.publish_third_party_items"
```
nicosantangelo, line 384: we'd have to delete the translations too right?

src/lib/api/builder.ts
```diff
@@ -670,6 +670,29 @@ export class BuilderAPI extends BaseAPI {
     }
   }
 
+  async publishCollection(collectionId: string, itemIds: string[], signedMessage: string, signature: string, qty: number, salt: string) {
```
nicosantangelo, line 673: should this method be `publishTPCollection` ?

src/modules/curations/itemCuration/reducer.ts
```diff
@@ -43,7 +60,47 @@ export function itemCurationReducer(state: ItemCurationState = INITIAL_STATE, ac
...
+        data: {
+          ...state.data,
+          [action.payload.collectionId]: mergedItemCurations
+        },
+        error: null
+      }
+    }
+
+    case PUSH_CHANGES_THIRD_PARTY_ITEMS_SUCCESS: {
+      const { itemCurations } = action.payload
```
nicosantangelo, line 81: we can add collectionId here, used below

src/modules/curations/itemCuration/reducer.ts
```diff
@@ -43,7 +60,47 @@ export function itemCurationReducer(state: ItemCurationState = INITIAL_STATE, ac
         },
         error: null
       }
+    }
+
+    case PUBLISH_THIRD_PARTY_ITEMS_SUCCESS: {
+      const { itemCurations } = action.payload
```
nicosantangelo, line 66: we can add collectionId here, used below

src/modules/curations/itemCuration/reducer.ts
```diff
@@ -43,7 +60,47 @@ export function itemCurationReducer(state: ItemCurationState = INITIAL_STATE, ac
...
+      const oldItemCurations = state.data[action.payload.collectionId]
+      const mergedItemCurations = oldItemCurations ? [...oldItemCurations, ...itemCurations] : itemCurations
```
nicosantangelo, line 68: Maybe it's best to also do `[...itemCurations]` to avoid having a pointer to the same array on the state

src/modules/curations/itemCuration/reducer.ts
```diff
@@ -43,7 +60,47 @@ export function itemCurationReducer(state: ItemCurationState = INITIAL_STATE, ac
...
+          [action.payload.collectionId]: mergedItemCurations
+        },
+        error: null
+      }
+    }
+
+    case PUSH_CHANGES_THIRD_PARTY_ITEMS_SUCCESS: {
+      const { itemCurations } = action.payload
+      const oldItemCurations = state.data[action.payload.collectionId]
+      const mergedItemCurations = oldItemCurations.map(ic => {
```
nicosantangelo, line 83: maybe we can rename ic to itemCuration (just for readability sake)

src/modules/thirdParty/sagas.ts
```diff
@@ -41,13 +57,47 @@ export function* getContractInstance(
   return contractInstance
 }
 
+export function* getPublishItemsSignature(thirdPartyId: string, qty: number) {
+  const maticChainId: ChainId = yield call(getChainIdByNetwork, Network.MATIC)
+  const provider = new ethers.providers.Web3Provider((window as any).ethereum)
```
nicosantangelo, line 62: shouldn't this provider come from decentraland-dapps? (instead of using window.ethereum)

src/modules/thirdParty/sagas.ts
```diff
@@ -110,4 +160,56 @@ export function* thirdPartySaga(builder: BuilderAPI) {
...
+      const { signature, signedMessage, salt } = yield call(getPublishItemsSignature, thirdParty.id, items.length)
+
+      const { items: newItems, itemCurations: newItemCurations }: { items: Item[]; itemCurations: ItemCuration[] } = yield call(
+        [builder, 'publishCollection'],
+        collectionId,
+        items.map(i => i.id),
+        signedMessage,
+        signature,
+        items.length, // qty
+        salt
```
nicosantangelo, line 178: Not for now, but I think [this comment](https://github.com/decentraland/builder-server/pull/419#discussion_r822101180) applies for this line too!

src/modules/thirdParty/sagas.ts
```diff
@@ -110,4 +160,56 @@ export function* thirdPartySaga(builder: BuilderAPI) {
...
+    } catch (error) {
+      yield put(publishThirdPartyItemsFailure(error.message))
+    }
+  }
+
+  function* handlePushChangesThirdPartyItemRequest(action: PushChangesThirdPartyItemsRequestAction) {
+    try {
+      const { items } = action.payload
+      const collectionId = items[0].collectionId
+      if (!collectionId) return
```
nicosantangelo, line 192: Question: Should this be a throw? to be able to handle it properly if necessary

src/modules/thirdParty/sagas.ts
```diff
@@ -110,4 +160,56 @@ export function* thirdPartySaga(builder: BuilderAPI) {
...
+
+      const itemCurations: ItemCuration[] = yield select(getItemCurations, collectionId!)
+      const effects: CallEffect<void>[] = items.map(item => {
+        const curation = itemCurations.find(ic => ic.itemId === item.id)
```
nicosantangelo, line 196: same comment as before but maybe we can rename this to itemCuration from ic for readability sake

nicosantangelo (commented):
Great work

src/modules/thirdParty/sagas.ts
juanmahidalgo: I tried! but seems like it's not exposing the `getSigner` method 🤔 
image

src/components/ActivityPage/Transaction/Transaction.tsx
juanmahidalgo: yes! removed, thanks! :)

src/lib/api/builder.ts
juanmahidalgo: agreed! renamed :)

src/modules/curations/itemCuration/reducer.ts
juanmahidalgo: oh yes, totally, updated!

src/modules/curations/itemCuration/reducer.ts
juanmahidalgo: updated!

src/modules/curations/itemCuration/reducer.ts
juanmahidalgo: updated!

src/modules/curations/itemCuration/reducer.ts
juanmahidalgo: for sure! updated

src/modules/curations/itemCuration/reducer.ts
```diff
@@ -43,7 +60,47 @@ export function itemCurationReducer(state: ItemCurationState = INITIAL_STATE, ac
...
+        const newCuration = itemCurations.find(newItemCuration => newItemCuration.itemId === ic.itemId)
+        if (newCuration) {
+          return newCuration
+        }
+        return ic
+      })
```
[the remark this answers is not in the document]
juanmahidalgo: good call, I think yours is better :)

src/modules/thirdParty/sagas.ts
juanmahidalgo: changed to `throw`!

src/modules/thirdParty/sagas.ts
juanmahidalgo: updated!

src/modules/thirdParty/sagas.ts
```diff
@@ -110,4 +160,56 @@ export function* thirdPartySaga(builder: BuilderAPI) {
...
+      yield put(pushChangesThirdPartyItemsSuccess(collectionId, newItemsCurations))
+      yield put(closeModal('PublishThirdPartyCollectionModal'))
+    } catch (error) {
+      yield put(pushChangesThirdPartyItemsFailure(error.message))
+    }
+  }
+
+  function* handlePublishAndPushChangesThirdPartyItemRequest(action: PublishAndPushChangesThirdPartyItemsRequestAction) {
+    const { thirdParty, itemsToPublish, itemsWithChanges } = action.payload
+    yield all([put(publishThirdPartyItemsRequest(thirdParty, itemsToPublish)), put(pushChangesThirdPartyItemsRequest(itemsWithChanges))])
```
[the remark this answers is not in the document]
juanmahidalgo: Mhmmm good question! 
I don't think it will be a problem, it will be called twice but the 2nd one won't do anything. I don't know if there's a way to avoid calling the 2nd one 🤔 .

LautaroPetaccio (changes requested):

src/modules/item/reducer.ts
```diff
@@ -383,6 +373,19 @@ export function itemReducer(state: ItemState = INITIAL_STATE, action: ItemReduce
...
+            ...state,
+            data: {
+              ...state.data,
+              ...items.reduce((accum, item) => {
+                accum[item.id] = item
+                return accum
+              }, {} as ItemState['data'])
+            }
+          }
+        }
```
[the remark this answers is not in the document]
juanmahidalgo: Yeah, but since we're not using the `isPublished` field for the TP items, I thought it might not be necessary to explicitly update the `isPublished` field. The `PUBLISH_THIRD_PARTY_ITEMS_SUCCESS` action will update the item curations so the logic to know if it's public will be there.
That said, you're right, could re-use the `PUBLISH_COLLECTION_SUCCESS` and it should be the same

src/lib/api/builder.ts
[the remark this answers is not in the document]
juanmahidalgo: Good idea! https://github.com/decentraland/builder-client/issues/32 created :)

src/lib/api/builder.ts
```diff
@@ -659,7 +659,7 @@ export class BuilderAPI extends BaseAPI {
     return fromRemoteCollection(remoteCollection)
   }
 
-  async publishCollection(collectionId: string) {
+  async publishDCLCollection(collectionId: string) {
```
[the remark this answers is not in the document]
juanmahidalgo: Done!

src/lib/api/builder.ts
```diff
@@ -716,6 +739,10 @@ export class BuilderAPI extends BaseAPI {
     return this.request('post', `/collections/${collectionId}/curation`)
   }
 
+  pushItemCuration(itemId: string): Promise<void> {
+    return this.request('post', `/items/${itemId}/curation`)
+  }
```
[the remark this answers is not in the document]
juanmahidalgo: YES! good call :)

src/lib/api/builder.ts
```diff
@@ -740,6 +767,10 @@ export class BuilderAPI extends BaseAPI {
     return this.request('patch', `/collections/${collectionId}/curation`, { curation: { status } })
   }
 
+  updateItemCurationStatus(itemId: string, status: CurationStatus): Promise<void> {
```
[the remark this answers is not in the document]
juanmahidalgo: Yup! updated

src/modules/thirdParty/reducer.ts
```diff
@@ -133,27 +132,6 @@ export function thirdPartyReducer(state: ThirdPartyState = INITIAL_STATE, action
...
-      switch (transaction.actionType) {
-        case PUBLISH_THIRD_PARTY_ITEMS_SUCCESS: {
-          const { thirdPartyId, items } = transaction.payload
-          return {
-            ...state,
-            data: {
-              ...state.data,
-              [thirdPartyId]: {
-                ...state.data[thirdPartyId],
-                totalItems: new BN(state.data[thirdPartyId].totalItems).add(new BN(items.length)).toString()
```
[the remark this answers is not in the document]
juanmahidalgo: Added the logic to `put` the `fetchThirdPartyAvailableSlotsRequest` action when the publish action finishes successfully.

src/modules/thirdParty/sagas.spec.ts
```diff
@@ -256,3 +277,172 @@ describe('when fetching third party available slots', () => {
...
+        }
+      ]
+    })
+
+    it('should put the fetch third party success action the response with the new itemCurations and close the PublishThirdPartyCollectionModal modal', () => {
+      return expectSaga(thirdPartySaga, mockBuilder)
+        .provide([
+          [select(getCollection, mockedItem.collectionId), collection],
+          [call(getPublishItemsSignature, thirdParty.id, 1), { signature: '', signedMessage: '', salt: '' }],
+          [matchers.call.fn(mockBuilder.publishTPCollection), { items: [], itemCurations }]
```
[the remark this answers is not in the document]
juanmahidalgo: Updated!

src/modules/thirdParty/sagas.ts
juanmahidalgo: This remains the same here, but it's all wrapped in the `cheque` object in the `publishTPCollection` method.

src/modules/thirdParty/sagas.ts
nicosantangelo: Hm true, what if we use this: https://github.com/decentraland/decentraland-transactions/blob/master/src/utils.ts#L29 ?

with `getConnectedProvider`

The reason why is that getConnectedProvider will correctly assign the provider, even if the user is doing something weird (like wallet connect for example)

src/modules/thirdParty/sagas.ts
```diff
@@ -110,4 +162,106 @@ export function* thirdPartySaga(builder: BuilderAPI) {
...
+      yield put(closeModal('PublishThirdPartyCollectionModal'))
+    } catch (error) {
+      yield put(pushChangesThirdPartyItemsFailure(error.message))
+    }
+  }
+
+  function* handlePublishAndPushChangesThirdPartyItemRequest(action: PublishAndPushChangesThirdPartyItemsRequestAction) {
+    const { thirdParty, itemsToPublish, itemsWithChanges } = action.payload
+    const collectionId = getCollectionId(itemsToPublish)
+    // We need to execute this two in secuence, because the push changes will create a new curation if there was one approved.
```
nicosantangelo, line 243:
```suggestion
    // We need to execute this two methods in sequence, because the push changes will create a new curation if there was one approved.
```

src/modules/thirdParty/sagas.ts
```diff
@@ -110,4 +162,106 @@ export function* thirdPartySaga(builder: BuilderAPI) {
...
+    } catch (error) {
+      yield put(publishThirdPartyItemsFailure(error.message))
+    }
+
+    // Push changes to items
+    try {
+      const resultFromPush: ItemCuration[] = yield call(pushChangesToThirdPartyItems, itemsWithChanges)
+      yield put(pushChangesThirdPartyItemsSuccess(collectionId, resultFromPush))
+    } catch (error) {
+      yield put(pushChangesThirdPartyItemsFailure(error.message))
```
nicosantangelo, line 262: The two methods look good! but I think it's weird to have Success and Failure cases without their corresponding Requests, no?

Now that we have splitted the logic, we could just have a single Request (`PUBLISH_AND_PUSH_CHANGES_THIRD_PARTY_ITEMS_REQUEST`) and a single SUCCESS/FAILURE

```
PUBLISH_AND_PUSH_CHANGES_THIRD_PARTY_ITEMS_SUCCESS
PUBLISH_AND_PUSH_CHANGES_THIRD_PARTY_ITEMS_FAILURE
```

the failure should trigger if any method fails. The success if both work.

What do you think?

nicosantangelo (changes requested):
A few more comments!

src/modules/thirdParty/sagas.ts
LautaroPetaccio: We should be using `these` instead of `this` as there are more than one method.

src/modules/thirdParty/sagas.ts
```diff
@@ -212,6 +239,29 @@ export function* thirdPartySaga(builder: BuilderAPI) {
 
   function* handlePublishAndPushChangesThirdPartyItemRequest(action: PublishAndPushChangesThirdPartyItemsRequestAction) {
     const { thirdParty, itemsToPublish, itemsWithChanges } = action.payload
-    yield all([put(publishThirdPartyItemsRequest(thirdParty, itemsToPublish)), put(pushChangesThirdPartyItemsRequest(itemsWithChanges))])
+    const collectionId = getCollectionId(itemsToPublish)
+    // We need to execute this two in secuence, because the push changes will create a new curation if there was one approved.
+    // It will create it with status PENDING, so the publish will fail if it's executed after that event.
```
LautaroPetaccio, line 244: If we're talking about the curations:
It will create `them` with status PENDING ...

src/modules/item/reducer.ts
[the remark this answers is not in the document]
LautaroPetaccio: If `isPublished` is kind of irrelevant, I would move it up to `PUBLISH_COLLECTION_SUCCESS` as it will avoid you from creating a test for this new behavior in the reducer and it will also make the code shorter.

src/modules/thirdParty/sagas.spec.ts
```diff
@@ -256,3 +277,256 @@ describe('when fetching third party available slots', () => {
...
+      ]
+    })
+
+    it('should put the fetch third party success action with the new itemCurations and close the PublishThirdPartyCollectionModal modal', () => {
+      const mockedItemReturnedByServer = { ...mockedItem, id: 'a new id' }
+      return expectSaga(thirdPartySaga, mockBuilder)
+        .provide([
+          [select(getCollection, item.collectionId), collection],
+          [call(getPublishItemsSignature, thirdParty.id, 1), { signature: '', signedMessage: '', salt: '' }],
+          [matchers.call.fn(mockBuilder.publishTPCollection), { items: [mockedItemReturnedByServer], itemCurations }]
```
LautaroPetaccio, line 330: The matchers of the API calls are not using `call` so their parameters aren't checked, would you mind changing them? At least for the success cases.

LautaroPetaccio (changes requested):

src/modules/thirdParty/sagas.ts
juanmahidalgo: Updated to use `getConnectedProvider` and `eth_signTypedData_v4` as suggested!

src/modules/item/reducer.ts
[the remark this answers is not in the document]
juanmahidalgo: Agreed, moved!

src/modules/curations/itemCuration/reducer.ts
```diff
@@ -43,7 +60,43 @@ export function itemCurationReducer(state: ItemCurationState = INITIAL_STATE, ac
...
+        error: null
+      }
+    }
+
+    case PUSH_CHANGES_THIRD_PARTY_ITEMS_SUCCESS: {
+      const { itemCurations, collectionId } = action.payload
+      const oldItemCurations = state.data[collectionId]
+      const mergedItemCurations = oldItemCurations.map(itemCuration => {
+        return itemCurations.find(newItemCuration => newItemCuration.itemId === itemCuration.itemId) || itemCuration
+      })
```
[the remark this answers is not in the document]
juanmahidalgo: sure :)

src/modules/thirdParty/sagas.ts
juanmahidalgo: I re-introduced `PUBLISH_AND_PUSH_CHANGES_THIRD_PARTY_ITEMS_SUCCESS` and `PUBLISH_AND_PUSH_CHANGES_THIRD_PARTY_ITEMS_FAILURE` 😂 Let me know what you think. Added some tests and the reducer logic as well.

src/modules/thirdParty/sagas.spec.ts
juanmahidalgo: Ohhh, sorry @LautaroPetaccio that's what you meant in your last message. OK, got it, everything updated to use the actual `call` and not the matchers.

src/modules/thirdParty/sagas.ts
```diff
@@ -41,13 +60,62 @@ export function* getContractInstance(
...
+
+  const accounts: string[] = yield call([provider, 'request'], { method: 'eth_requestAccounts', params: [], jsonrpc: '2.0' })
+  const from = accounts[0]
+
+  const signature: string = yield call([provider, 'request'], {
+    method: 'eth_signTypedData_v4',
+    params: [from, msgString],
+    jsonrpc: '2.0'
+  })
+
```
nicosantangelo, line 101: I want to clarify something here. There's a competing implementation between web3x and ethers in this codebase. The reason why you don't see a signer or a request here (or in the marketplace) is because you should do something like:

```
const eth: Eth = yield call(getEth)

const personal = new Personal(eth.provider)
const signature: string = yield personal.sign(...)
```

where `getEth` is a wrapper of the provider for web3x (the old lib we used to use).

I think this implementation is correct, but I wanted to clarify why this came up. An example of a signature can be seen on the ens module's handler `handleSetAlias` here in the builder

src/modules/thirdParty/sagas.ts
```diff
@@ -110,4 +178,103 @@ export function* thirdPartySaga(builder: BuilderAPI) {
...
+      const resultFromPushChanges: ItemCuration[] = yield call(pushChangesToThirdPartyItems, itemsWithChanges)
+      const newItemCurations = [...resultFromPublish.newItemCurations, ...resultFromPushChanges]
+
+      yield put(publishAndPushChangesThirdPartyItemsSuccess(collectionId, resultFromPublish.newItems, newItemCurations))
+      yield put(fetchThirdPartyAvailableSlotsRequest(collectionId)) // re-fetch available slots after publishing
+    } catch (error) {
+      yield put(publishAndPushChangesThirdPartyItemsFailure(error.message)) // TODO: show to the user that something went wrong
+    }
+
+    yield put(closeModal('PublishThirdPartyCollectionModal'))
```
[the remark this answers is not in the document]
juanmahidalgo: good question! moved to the success case :)

nicosantangelo (approved):

LautaroPetaccio (approved):

PR #1826, opened 2022-03-10
fix: Don't request all items at once

fzavalia:
Absolutely every item is being requested when opening the curation page and approving a collection.
Now only required items for the moment are fetched.

### DIFF
--- a/src/components/CurationPage/CollectionRow/CollectionRow.tsx
+++ b/src/components/CurationPage/CollectionRow/CollectionRow.tsx
@@ -14,6 +14,11 @@ import { Props } from './CollectionRow.types'
 import './CollectionRow.css'
 
 export default class CollectionRow extends React.PureComponent<Props> {
+  componentDidMount = () => {
+    const { collection, onFetchCollectionItems } = this.props
+    onFetchCollectionItems(collection.id)
+  }
+
   handleNavigateToForum = (event: React.MouseEvent<HTMLSpanElement, MouseEvent>) => {
     const { collection } = this.props
     window.open(collection.forumLink!, '_blank')
--- a/src/components/CurationPage/CollectionRow/CollectionRow.tsx
+++ b/src/components/CurationPage/CollectionRow/CollectionRow.tsx
@@ -15,8 +15,14 @@ import './CollectionRow.css'
 
 export default class CollectionRow extends React.PureComponent<Props> {
   componentDidMount = () => {
-    const { collection, onFetchCollectionItems } = this.props
-    onFetchCollectionItems(collection.id)
+    const { collection, items, onFetchCollectionItems } = this.props
+    // Only refetch when the collection has no items in store
+    // This way we avoid fetching data too many times
+    // Beware that as data is not refreshed, new changes done by another user
+    // will not be seen and a manual browser refresh will be required.
+    if (items.length === 0) {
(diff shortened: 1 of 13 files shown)


src/modules/item/sagas.ts
```diff
@@ -142,9 +143,16 @@ export function* itemSaga(legacyBuilder: LegacyBuilderAPI, builder: BuilderClien
...
 
   function* handleFetchItemsRequest(action: FetchItemsRequestAction) {
-    const { address } = action.payload
+    const { address, itemIds } = action.payload
     try {
-      const items: Item[] = yield call(() => legacyBuilder.fetchItems(address))
+      let items: Item[]
+
+      if (itemIds) {
+        items = yield call([Promise, 'all'], [itemIds.map(id => legacyBuilder.fetchItem(id))])
```
juanmahidalgo, line 151: Since this is being triggered per collection, couldn't we use the `fetchCollectionItems` so we avoid doing one request per each item of the collection? 
Instead of receiving an array of `itemIds` we could expect a `collectionId` and fetch all the items at once.

fzavalia: You are right. I added that before realizing there was a `fetchCollectionItems` action. Changing it.

LautaroPetaccio: Offtopic: `yield all([])` also works (instead of using a call on Promise.all)

src/components/CurationPage/CurationPage.tsx
```diff
@@ -172,13 +172,15 @@ export default class CurationPage extends React.PureComponent<Props, State> {
...
           </Section>
           {totalPages > 1 && (
-            <Pagination
-              firstItem={null}
-              lastItem={null}
-              totalPages={totalPages}
-              activePage={page}
-              onPageChange={(_event, props) => this.setState({ page: +props.activePage! })}
-            />
+            <div className="pagination">
```
LautaroPetaccio, line 175: Would you mind passing the pagination class as a prop to `Pagination` (in `className`) instead of creating a new div?

src/components/CurationPage/CollectionRow/CollectionRow.tsx
```diff
@@ -14,6 +14,11 @@ import { Props } from './CollectionRow.types'
 import './CollectionRow.css'
 
 export default class CollectionRow extends React.PureComponent<Props> {
+  componentDidMount = () => {
+    const { collection, onFetchCollectionItems } = this.props
+    onFetchCollectionItems(collection.id)
```
LautaroPetaccio, line 19: What if we load the items for the collection once? We can have a selector that checks if there are items for that collection and avoid the loading time for the curations.

LautaroPetaccio (changes requested):

src/components/CurationPage/CurationPage.tsx
fzavalia: `Pagination` does not have a className prop afaik. Will check

src/modules/item/sagas.ts
fzavalia: Nice tip

fzavalia: Ended up removing the `itemIds` as they were unnecessary.

LautaroPetaccio (approved):

PR #1829, opened 2022-03-10
feat: Add push TP changes logic

### LINKED ISSUE #1823: Use the content_hash and local_content_hash to check if the item hash changes to be pushed
Use the `content_hash` and the `local_content_hash` (contentHash and serverContentHash) when selecting the items in the TP collection details component to identify if an item hash changes to be pushed or not.

juanmahidalgo:
Closes #1823

## No item curations

image

## Immediately after publish (button es blocked, there are no changes)

image

## After making changes to the item (button is enabled, the content hash from the item is not the same as in the item curation)

image

## After pushing the changes

image


src/components/Modals/PublishThirdPartyCollectionModal/PublishThirdPartyCollectionModal.tsx
```diff
@@ -3,16 +3,31 @@ import { ModalNavigation, Button } from 'decentraland-ui'
...
-import { isStatusAllowedToPushChanges } from 'modules/item/utils'
+import { isAllowedToPushChanges } from 'modules/item/utils'
 import { PublishButtonAction } from 'components/ThirdPartyCollectionDetailPage/CollectionPublishButton/CollectionPublishButton.types'
 import { getTPButtonActionLabel } from 'components/ThirdPartyCollectionDetailPage/CollectionPublishButton/CollectionPublishButton'
 import { Props } from './PublishThirdPartyCollectionModal.types'
 
 export default class PublishThirdPartyCollectionModal extends React.PureComponent<Props> {
+  getItemsToPublish = () => {
+    const { items, itemsStatus } = this.props
+    return items.filter(item => itemsStatus[item.id] === SyncStatus.UNPUBLISHED)
```
nicosantangelo, line 14: I was thinking that maybe this could be a method somewhere else. Either in the `item/selectors.ts` file or in `item/utils.ts`. Mostly because it seems like something we'll use in more than one place, but also because I think it's repeated [here](https://github.com/decentraland/builder/pull/1832/files#r826330790) ? but with a different implementation.

In case the impl HAS to be different, the find for a itemCuration might be a candidate for abstraction as I think is the same [here](https://github.com/decentraland/builder/pull/1829/files#diff-d678e3d56b94dc53c3a5660c2639e82446a8e15df0169c382f04f3e0cf6e7d90R36) too

What do you think?

src/components/Modals/PublishThirdPartyCollectionModal/PublishThirdPartyCollectionModal.container.ts
```diff
@@ -13,16 +13,19 @@ import { getCollection } from 'modules/collection/selectors'
 import { getLoading, getCollectionItems, getStatusForItemIds } from 'modules/item/selectors'
 import { OwnProps, MapStateProps, MapDispatchProps, MapDispatch } from './PublishThirdPartyCollectionModal.types'
 import PublishCollectionModal from './PublishThirdPartyCollectionModal'
+import { getItemCurations } from 'modules/curations/itemCuration/selectors'
```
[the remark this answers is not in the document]
juanmahidalgo: for sure :)

src/components/Modals/PublishThirdPartyCollectionModal/PublishThirdPartyCollectionModal.tsx
juanmahidalgo: Agreed, it should be outside this component. I moved it to the utils file :) 
I'll update #1832 once this one gets merged!

nicosantangelo (approved):

PR #1832, opened 2022-03-11
feat: Customize TP Under review Popup content

### LINKED ISSUE #1824: Customize Popup label when the publish is blocked
When trying to publish new TP items and there are some items under review, the button Popup should tell the user that the publishing action is blocked due to that reason. Right now it has a message but it's the one used for DCL collections, so it's not specific enough to be clear to the user. 

If there are items with changes to push and the button is blocked as well, communicate that clearly to the user.

juanmahidalgo:
Closes #1824

## Being able to push changes (it has a different contentHash than its curation)

image

## Trying to publish while there's a batch under review

image

## Trying to push changes and publish at the same time while there's a batch under review

image


src/components/ThirdPartyCollectionDetailPage/CollectionPublishButton/CollectionPublishButton.tsx
```diff
@@ -53,11 +53,30 @@ const CollectionPublishButton = (props: Props) => {
...
+        return t('third_party_collection_detail_page.cant_publish')
+    }
+  }, [buttonAction, itemsTryingToPublish])
+
   const hasPendingItemCurations = itemCurations && !!itemCurations.find(ic => ic.status === CurationStatus.PENDING)
   const isTryingToPublish = [PublishButtonAction.PUBLISH, PublishButtonAction.PUBLISH_AND_PUSH_CHANGES].includes(buttonAction)
 
   return !isLoadingItemCurations && isTryingToPublish && hasPendingItemCurations ? (
-    <UnderReview type="publish" />
+    <UnderReview content={underReviewButtonLabel} />
```
nicosantangelo, line 79: This works! but hear me out.

I feel like the original `UnderReview` component made sense because it avoided having to repeat the Popup logic and also decided which translation to show:

```ts
if (collection.isApproved) {
  if (hasPendingCuration) {
    button = <UnderReview type="push" />
  } else if (status === SyncStatus.UNSYNCED) {
    // (...)
  }
} else {
  button = <UnderReview type="publish" />
}
```

in our case, the component doesn't seem to be abstracting much, as we pass the entire text and just let it wrap a button with a popup.

So I was thinking, what about (options):
- Letting the component decide which translation to use by receiving: `itemsTryingToPublish` and `buttonAction`
- Pass `items` and `itemCurations` to the component and move the logic there (both memos)
- Delete underreview and move the code here
- Some combination of the three

I don't know if any of those is better than the others or even than the current solution. There might be other options I haven't considered. What do you think?

src/components/ThirdPartyCollectionDetailPage/CollectionPublishButton/CollectionPublishButton.tsx
```diff
@@ -53,11 +53,30 @@ const CollectionPublishButton = (props: Props) => {
     onClick(collection.id, itemIds, buttonAction)
   }, [collection, items, buttonAction])
 
+  const itemsTryingToPublish = useMemo(
+    () => items.filter(item => !itemCurations.find(itemCuration => itemCuration.itemId === item.id)).length,
```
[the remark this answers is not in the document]
juanmahidalgo: I think it's not necessary to check the status here. I think that in order to get the ones that are gonna be published, we need the items with NO curations at all, because if they are pending/accepted/rejected, the action would be the push changes one.

src/components/ThirdPartyCollectionDetailPage/CollectionPublishButton/CollectionPublishButton.tsx
juanmahidalgo: Agreed! I think we don't need an extra component in this case, so I went with option 3 :) 
`UnderReview` component deleted and `PopUp` included in the jsx from the `CollectionPublishButton`

PR #1843, opened 2022-03-14
feat: Suggest adding collection contract address as part of the TP urn

### LINKED ISSUE #1838: We should support at least an address for the collection name for thirdparties an encourage to use it
@nachomazzara commented on [Fri Mar 11 2022](https://github.com/decentraland/account/issues/196)

Can we add a message saying we encourage the creator to use the collection address as the id?

juanmahidalgo:
## Modal before this changes: 

image

## After these changes:

image

Closes #1838


src/modules/collection/types.ts
```diff
@@ -41,3 +41,4 @@ export type Access = {
 }
 
 export const COLLECTION_NAME_MAX_LENGTH = 32
+export const TP_COLLECTION_NAME_MAX_LENGTH = 64
```
nicosantangelo, line 44: Should we be using this length on the EditCollectionNameModal? it was generic up until this point but it uses `COLLECTION_NAME_MAX_LENGTH`

nicosantangelo: The UI scares me a bit, things are a bit cramped up:

Current:
Screen Shot 2022-03-14 at 12 24 53


Longer:

image

juanmahidalgo: Good call! I updated the modal name to use `TP_COLLECTION_NAME_MAX_LENGTH` as the max length for TPs. 
Also, I updated the styles so the name container doesn't overflow when it's too long:

image

nicosantangelo: telegram-cloud-photo-size-1-5010286617772009953-y

nicosantangelo (approved):

Issue #1852, opened 2022-03-14
Make the transaction to the contract updating the merkle tree root


LautaroPetaccio:
Following the template created in the issue #1847, make a call to the contract (`reviewThirdPartyWithRoot`) using the computed merkle tree #1847 and the signature & message data retrieved from the approval data #1846.

PR #1856, opened 2022-03-15
feat: use the approval data to send the reviewThirdPartyWithRoot tx

### LINKED ISSUE #1846: Implement the gathering of approval data for TP collections
Following the steps created in the #1845 issue, add the mechanism to query the builder-server for the approval data:
Obtain the approval data by doing a GET request to the builder-server’s `/collections/:id/approvalData` endpoint while showing a spinner on the modal. The server should respond with all items that need curation and their content hash. Careful with the casing. The server will probably return the data with `snake_casing` but we need to use `cameCase` for the lib.

nicosantangelo:
Depends on https://github.com/decentraland/builder-server/pull/436 for the correct approval data to be returned from the server

Closes #1846 #1846 #1852


package.json
```diff
@@ -7,6 +7,7 @@
     "@babylonjs/core": "^4.2.0",
     "@babylonjs/loaders": "^4.2.0",
     "@dcl/builder-client": "^1.5.0",
+    "@dcl/content-hash-tree": "^1.1.2",
```
juanmahidalgo, line 10: the `-lock` file has 44,697 additions :O would you mind checking locally if the diff looks good? GH is not showing me the changes cause it's too large

src/modules/collection/sagas.ts
```diff
@@ -545,13 +548,26 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
...
 
       // Open the ApprovalFlowModal with the items to be approved
 
       // 4. Make the transaction to the contract (update of the merkle tree root with the signature and its parameters)
+      const maticChainId: ChainId = yield call(getChainIdByNetwork, Network.MATIC)
+      const thirdPartyId = extractThirdPartyId(collection.urn)
+      const root = tree.merkleRoot
+      const { r, s, v } = ethers.utils.splitSignature(cheque.signature)
+      const thirdPartyContract: ContractData = yield call(getContract, ContractName.ThirdPartyRegistry, maticChainId)
+      const txHash: string = yield call(sendTransaction, thirdPartyContract, contract =>
```
juanmahidalgo, line 566: I think that the `reviewThirdPartyWithRoot`'s 3rd parameter is an array of `ConsumeSlotsParam` so we might be missing the `qty` and `salt` for our cheque.

Would it be something like the following? (I haven't test it)
```
contract.reviewThirdPartyWithRoot(thirdPartyId, root, [{ signR: r, signS: s, signV: v, qty, salt  }])
```


Also related to this, should we use what @LautaroPetaccio did so the `sendTransaction` is easier to test? 
```
await sendTransaction(contract, 'reviewThirdPartyWithRoot', thirdPartyId, root, [cheque]`
```

nicosantangelo: oh shoot! right on both accounts. I'll give a it a go and test it

juanmahidalgo (approved):
excellent 💯

Issue #1849, opened 2022-03-14
Implement the deployment logic for the TP collections in the approval flow


LautaroPetaccio:
Implement the catalyst deployment logic to deploy the TP items following the template in the #1845 issue.

PR #1858, opened 2022-03-15
feat: Update item curation approval

### LINKED ISSUE #1850: Implement the logic to update the item curations in the TP approval flow
Following the template created in the #1845 issue, update the curations of the deployed items to be `APPROVED`.

juanmahidalgo:
Closes #1850 #1849


src/modules/item/types.ts
```diff
@@ -132,6 +132,8 @@ export type CatalystItem = Omit<BaseItem, 'createdAt' | 'updatedAt'> & {
   collectionAddress: string
 }
 
+export type CatalystTPItem = Omit<CatalystItem, 'rarity' | 'collectionAddress'> & {}
```
juanmahidalgo, line 135: This is temporary until we have a proper definition of the entity metadata needed

src/modules/collection/sagas.ts
```diff
@@ -538,6 +540,31 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
...
 
+  function* updateItemCurationsStatus(itemCurations: ItemCuration[], status: CurationStatus) {
+    const effects: CallEffect<ItemCuration>[] = itemCurations.map(curation => {
+      return call([builder, 'updateItemCurationStatus'], curation.itemId, status)
+    })
+    const newItemCuration: ItemCuration[] = yield all(effects)
+    return newItemCuration
+  }
+
+  function* getItemsAndEntitiesToDeploy(collection: Collection, tree?: MerkleDistributorInfo, hashes?: Record<string, string>) {
```
nicosantangelo, line 551: Ok so, this works! but I think it's a miss-leading signature for the method. The problem is not your implementation but the fact that we decided to treat `TP items` the same as `standard Items`. But anyways here we are.

What I mean by miss-leading is that, if I look at this method without *a lot of context*, I can't really tell that the last two parameters mean "it's TP".

```ts
getItemsAndEntitiesToDeploy(standardCollection)

getItemsAndEntitiesToDeploy(tpCollection, tree, hashes)
```

nothing tells me that I can't make this call: `getItemsAndEntitiesToDeploy(tpCollection, tree, hashes)`.
Same thing for `buildItemEntity`.

SO, I was thinking how to fix this aaaand... I can't really come up with a better answer than making three methods. A generic one used internally and two public ones *OR* an override (but I'll be the first time we do it)

```ts
function* getItemsAndEntitiesToDeploy(collection: Collection) {
function* getItemsAndEntitiesToDeploy(collection: Collection,  tree: MerkleDistributorInfo, hashes: Record<string, string>) {

function* getStandardItemsAndEntitiesToDeploy(collection: Collection) {
function* getTPtemsAndEntitiesToDeploy(collection: Collection,  tree: MerkleDistributorInfo, hashes: Record<string, string>) {

// same idea for buildItemEntity
```

What do you think? I'd like to know what @LautaroPetaccio thinks about this too tho

src/modules/item/export.ts
```diff
@@ -118,6 +126,35 @@ function buildItemEntityMetadata(collection: Collection, item: Item): CatalystIt
...
 }
 
+function buildTPItemEntityMetadata(
+  collection: Collection,
+  item: Item,
+  itemHash: string,
+  tree: MerkleDistributorInfo
+): CatalystItem | CatalystTPItem {
+  const baseEntityData = {
+    id: buildThirdPartyURN(extractThirdPartyId(collection.urn), collection.id, item.id),
```
nicosantangelo, line 136: Careful here, the second param is the name, and the third the token id. You can decode the URN and use the data from there

```ts
const decodedUrn = decodeUrn(collection.urn)
buildThirdPartyURN(
  decodedCollectionUrn.thirdPartyName,
  decodedCollectionUrn.thirdPartyCollectionId,
  decodedUrn.thirdPartyTokenId
)
```

src/modules/item/export.ts
```diff
@@ -118,6 +126,35 @@ function buildItemEntityMetadata(collection: Collection, item: Item): CatalystIt
...
+    ...getBaseEntityMetadata(item)
+  }
+  return {
+    merkleProof: getMerkleProof(tree, itemHash, baseEntityData),
+    ...baseEntityData
+  }
```
nicosantangelo, line 142: If you want to avoid a double spread you could:
```ts
{
    ...getBaseEntityMetadata(item)
    merkleProof: getMerkleProof(tree, itemHash, baseEntityData),
    id: buildThirdPartyURN(extractThirdPartyId(collection.urn), collection.id, item.id),
}
```

src/modules/item/export.ts
juanmahidalgo: I updated the code to the following:
```
const decodedURN = decodeURN(collection.urn)
  if (decodedURN.type !== URNType.COLLECTIONS_THIRDPARTY) {
    throw new Error('URN is not a third party URN')
  }
  const { thirdPartyCollectionId, thirdPartyName } = decodedURN
  const baseEntityData = {
    id: buildThirdPartyURN(thirdPartyName, thirdPartyCollectionId!, item.id),
    ...getBaseEntityMetadata(item)
  }
```
But I have some doubts about it...
The first thing I'm not fully bought into, is the boilerplate needed so TypeScript infers the URN correctly and so I can destructure the `thirdPartyCollectionId` and `thirdPartyName` (the `if` does the trick).

And while I was writing the code, I feel that we're decoding a URN and then building it again with the same parameters but the tokenId, which is the last one... so I believe all that process could be replaced by:

```
id: `${collection.urn}:${item.id}`
```

Maybe it can be abstracted in a method like `appendTokenId` so the logic remains in the `urn` utils file in case we change the implementation, but I Think this is more straightforward.

What do you think @nicosantangelo ?

src/modules/item/export.ts
juanmahidalgo: The thing is that I need to send the `baseEntityData` as a parameter to `getMerkleProof`, so I need it to be defined first, that's why I declared it outside the returning object.

src/modules/item/export.ts
```diff
@@ -100,22 +101,30 @@ function calculateFilesSize(files: Array<Blob>) {
...
-    rarity: item.rarity,
     i18n: [{ code: 'en', text: item.name }],
     data: item.data,
     image: IMAGE_PATH,
     thumbnail: THUMBNAIL_PATH,
     metrics: item.metrics
   }
+
+  return isTP
+    ? { ...baseEntityProps, id: buildThirdPartyURN(extractThirdPartyId(collection.urn), collection.id, item.id) }
```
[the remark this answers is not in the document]
juanmahidalgo: Oh, I thought it would extract the id from the URN :O, thanks!

src/modules/collection/sagas.ts
juanmahidalgo: Updated to something similar to what you suggested. It's more code but it's more readable.

LautaroPetaccio: I think that re-using this function will be a problem, specially because:
- We don't need the item entities for the TP items, as the syncing of the TP items can be checked by using the item's `content_hash` and `local_content_hash`
- The `getApprovalData` endpoint was created to not need all the items in the UI, but for the standard items, that is required, as it needs to show them. We're going to show only an amount of them.

src/modules/collection/sagas.ts
```diff
@@ -578,8 +619,50 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
...
+
+          // If cancel exit flow
+        } else if (cancel) {
+          return
+        }
+      }
+
       // 6. If the collection was approved but it had a pending curation, approve the curation
 
+      const itemCurations: ItemCuration[] = yield select(getItemCurations, collection.id)
```
LautaroPetaccio, line 661: Do we need all the item curations to change their status? I'm thinking on when we'll have pagination and multiple items to approve.

LautaroPetaccio (changes requested):

src/modules/item/export.ts
```diff
@@ -118,6 +126,33 @@ function buildItemEntityMetadata(collection: Collection, item: Item): CatalystIt
...
+  if (!item.urn) {
+    throw new Error('Item does not have URN')
+  }
+  const baseEntityData = {
+    id: item.urn,
+    ...getBaseEntityMetadata(item)
+  }
+  return {
+    merkleProof: getMerkleProof(tree, itemHash, baseEntityData),
+    ...baseEntityData
```
LautaroPetaccio, line 139: The TP item will need the `content` property, where its keys are file names and its values are the hashes of the files.

LautaroPetaccio (changes requested):

src/modules/item/export.ts
juanmahidalgo: updated! thanks for catching that it was missing :D

LautaroPetaccio (approved):

src/modules/collection/sagas.ts
```diff
@@ -538,8 +539,45 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
...
+        const entity: DeploymentPreparationData = yield call(buildItemEntity, catalyst, collection, item)
+
+        itemsToDeploy.push(item)
+        entitiesToDeploy.push(entity)
+      }
+    }
+    return { itemsToDeploy, entitiesToDeploy }
+  }
+
+  function* getTPItemsAndEntitiesToDeploy(collection: Collection, tree: MerkleDistributorInfo, hashes: Record<string, string>) {
```
LautaroPetaccio, line 565:
```suggestion
  function* getTPItemsToDeploy(collection: Collection, tree: MerkleDistributorInfo, hashes: Record<string, string>) {
```

PR #1876, opened 2022-03-18
chore: Fetch collectibles via builder-client

### LINKED ISSUE #1816: Move opensea api call to the builder-server and builder-client
- Add [method](https://github.com/decentraland/builder-client/blob/main/src/client/BuilderClient.ts) to hit the server.
- Move endpoint to the builder-server

fzavalia:
Closes #1816


src/modules/asset/sagas.ts
```diff
@@ -6,57 +6,73 @@ import {
...
-      })
+  }
+
+  async function getNFTs(owner: string, cursor?: string): Promise<NFT[]> {
+    const response = await client.getNFTs({ owner, cursor })
+
+    const { next, nfts } = response
+
+    if (next) {
+      await sleep(400)
```
nicosantangelo, line 70: Question: why's this necessary? I see that the old implementation had a `sleep(400)` but we keeping it here for a particular reason?
seems like we could just hit the API right?

fzavalia: I don't know what the reason behind the sleep was. Just kept it there as the backend is still calling the same opensea endpoint. Yet I think its safe to remove it.

nicosantangelo (approved):

PR #1881, opened 2022-03-18
fix: push changes

cazala:
This PR fixes the push changes flow that is currently not working due to what it seems to be a miss-match in the generation of the content hash on the backend. I commented the piece of code that uses the `item.currentContentHash` when available, and it always generates it on the frontend. Since this hash and the one in the backend are different, I had to hardcode the `hasHashes` variable in the `getItemSyncedStatus` helper so it doesn't compare the backend hashes and it fallback to the previous logic that compares all the relevant properties of the item and the entity.


cazala:
closed in favor of #1888 1888

PR #1883, opened 2022-03-21
test: Add missing tests for the TP approval flow

### LINKED ISSUE #1882: Add test for the TP Approval flow
Add tests to the TP approval flow sagas since it's still untested.

juanmahidalgo:
## Description

Adds tests for the TP approval flow, considering all the steps and what happens if they fail.

Closes #1882


src/modules/collection/sagas.spec.ts
```diff
@@ -781,3 +801,240 @@ describe('when executing the approval flow', () => {
...
+
+      describe('when the consume slots transaction fails', () => {
+        it('should open the modal in an error state', () => {
+          const consumeSlotsError = 'Error when consuming slots'
+          const parsedSignature = ethers.utils.splitSignature(cheque.signature)
+          const merkleTree = generateTree(Object.values(contentHashes))
+          return expectSaga(collectionSaga, mockBuilder, mockCatalyst)
+            .provide([
+              [call([mockBuilder, 'fetchApprovalData'], TPCollection.id), { cheque, content_hashes: contentHashes }],
+              [delay(1000), void 0],
```
nicosantangelo, line 951: any reason not tu use `undefined` here?

src/modules/collection/sagas.spec.ts
```diff
@@ -37,16 +41,30 @@ import {
...
 } from './actions'
 import { collectionSaga } from './sagas'
 import { Collection } from './types'
 import { getLatestItemHash, UNSYNCED_COLLECTION_ERROR_PREFIX } from './utils'
-import { getOpenModals } from 'decentraland-dapps/dist/modules/modal/selectors'
+
+jest.mock('./utils', () => ({
+  ...jest.requireActual('./utils'),
+  isTPDeployEnabled: jest.fn().mockImplementation(() => true)
+}))
```
nicosantangelo, line 54: question: are we trying to mock all functions from `./utils` here or is it used more as a way to mock isTPDeployEnabled.

I'm asking because we were already importing `import { getLatestItemHash, UNSYNCED_COLLECTION_ERROR_PREFIX } from './utils'` here but now those will be mocked and if we just need to mock `isTPDeployEnabled` we can do that on a top-level `beforeAll`

nicosantangelo (commented):
Two comments but 💯

src/modules/collection/sagas.spec.ts
juanmahidalgo: Let's continue it in Slack :)

PR #1890, opened 2022-03-22
feat: disallow invalid slots to buy

### LINKED ISSUE #1877: Disallow buying fractional slots
The BuySlotsModal should only accept integer numbers

nicosantangelo:
Closes #1877

### DIFF
--- a/src/components/Modals/BuyItemSlotsModal/BuyItemSlotsModal.tsx
+++ b/src/components/Modals/BuyItemSlotsModal/BuyItemSlotsModal.tsx
@@ -54,14 +54,24 @@ export default class BuyItemSlotsModal extends React.PureComponent<Props, State>
   }
 
   isValidSlotAmount() {
+    const { thirdParty } = this.props.metadata
     const { slotsToBuy } = this.state
-    const slotsAmount = parseInt(slotsToBuy, 10)
-    return Number.isInteger(slotsAmount) && slotsAmount > 0
+    const slotsAmount = Number(slotsToBuy)
+    const availableSlots = Number(thirdParty.availableSlots)
+    return Number.isInteger(slotsAmount) && slotsAmount > 0 && availableSlots >= slotsAmount
+  }
+
+  applySlotBuySlippage() {
+    const { slotPrice } = this.props
+    const { slotsToBuy } = this.state
+    const cost = BigNumber.from(slotPrice).mul(Number(slotsToBuy))
+    return applySlotBuySlippage(cost).toString()
   }
 
   render() {
-    const { name, isBuyingItemSlots, error, isFetchingSlotPrice, slotPrice } = this.props
+    const { name, slotPrice, metadata, isBuyingItemSlots, isFetchingSlotPrice, error } = this.props
     const { slotsToBuy } = this.state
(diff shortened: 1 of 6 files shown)


src/modules/translation/languages/es.json
```diff
@@ -258,8 +258,10 @@
     "not_enough_mana": "No tienes suficiente {symbol} para comprar esa cantidad de slots.",
     "get_mana": "Puedes conseguir más utilizando la {link} dapp.",
     "buy_slots": "Comprar espacio",
-    "how_many_slots_title": "Cuántos slots?",
-    "slots_value": "El costo aproximado de cada espacio es {symbol}{slot_cost}"
+    "how_many_slots_title": "¿Cuántos slots?",
+    "buy_slots_error": "Los slots deben ser un número entero",
+    "slots_value": "El costo aproximado de cada espacio es {symbol}{slot_cost}",
+    "total_cost": "El costo total será {symbol}{total_cost}. Tiene un 3% de deslizamientopara cubrir los cambios de precio del mercado."
```
juanmahidalgo, line 264:
```suggestion
    "total_cost": "El costo total será {symbol}{total_cost}. Tiene un 3% de deslizamiento para cubrir los cambios de precio del mercado."
```

src/components/Modals/BuyItemSlotsModal/BuyItemSlotsModal.tsx
```diff
@@ -69,18 +80,28 @@ export default class BuyItemSlotsModal extends React.PureComponent<Props, State>
...
-                    total_cost: slotPrice && slotsToBuy ? slotPrice * Number(slotsToBuy) : 0
+                    total_cost: slotPrice && this.isValidSlotAmount() ? slotPrice * Number(slotsToBuy) : 0
                   })}
                 </div>
                 <div className={styles.slotValue}>
                   {t('buy_item_slots_modal.total_cost', {
                     symbol: <Mana network={Network.MATIC} size="small" />,
-                    total_cost: slotPrice && slotsToBuy ? applySlotBuySlippage(BigNumber.from(slotPrice).mul(slotsToBuy)).toString() : 0
+                    total_cost:
+                      slotPrice && this.isValidSlotAmount() ? applySlotBuySlippage(BigNumber.from(slotPrice).mul(slotsToBuy)).toString() : 0
```
juanmahidalgo, line 104: I think that if the user introduces something like `12.5`, this will try to apply the slippage to that number, since `slotPrice` is still the input from the user (we might need to call the `parseInt` when setting the state).
When the `parseInt` is called in the `isValidSlotAmount`, the `slotsAmount` will always be an integer, so I believe it will always return `true`, right?

nicosantangelo: You're right, you also made me realize that we weren't checking against the available slots of the third party so I added that too

https://user-images.githubusercontent.com/692077/159593663-2f30240b-b5c5-408f-b751-2efa94b26299.mov

juanmahidalgo: Ohhh nice! looks awesome :) 
Changes approved!

PR #1891, opened 2022-03-22
feat: can't delete published collections

### LINKED ISSUE #1873: Disable the option to delete published TP collections
The server already handles this correctly, but the front-end should block the context option from the collection detail too:

nicosantangelo:
Closes #1873

### DIFF
--- a/src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.tsx
+++ b/src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.tsx
@@ -44,7 +44,6 @@ export default class CollectionContextMenu extends React.PureComponent<Props> {
   render() {
     const { collection } = this.props
     const canDelete = this.canDelete()
-    console.log('canDelete: ', canDelete)
     return (
       <Dropdown
         className={styles.dropdown}
@@ -73,7 +72,7 @@ export default class CollectionContextMenu extends React.PureComponent<Props> {
                   trigger={<Dropdown.Item text={t('global.delete')} />}
                 />
               ) : (
-                <Dropdown.Item text={t('global.delete')} disabled={true} />
+                <Dropdown.Item text={t('global.delete')} disabled />
               )
             }
             hideOnScroll={true}
--- a/src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.tsx
+++ b/src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.tsx
@@ -36,8 +36,15 @@ export default class CollectionContextMenu extends React.PureComponent<Props> {
     }
   }
 
+  canDelete() {
+    const { collection } = this.props
+    return !isLocked(collection) && !collection.isPublished
+  }
+
   render() {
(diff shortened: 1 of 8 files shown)


src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.tsx
```diff
@@ -36,8 +36,15 @@ export default class CollectionContextMenu extends React.PureComponent<Props> {
...
 
+  canDelete() {
+    const { collection } = this.props
+    return !isLocked(collection) && !collection.isPublished
+  }
+
   render() {
     const { collection } = this.props
+    const canDelete = this.canDelete()
+    console.log('canDelete: ', canDelete)
```
juanmahidalgo, line 47: 🧹

src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.tsx
```diff
@@ -53,32 +60,36 @@ export default class CollectionContextMenu extends React.PureComponent<Props> {
...
+            trigger={
+              canDelete ? (
+                // ConfirmDelete breaks the popup, so we just skip it when unnecessary
+                <ConfirmDelete
+                  name={collection.name}
+                  onDelete={this.handleDeleteCollection}
+                  trigger={<Dropdown.Item text={t('global.delete')} />}
+                />
+              ) : (
+                <Dropdown.Item text={t('global.delete')} disabled={true} />
```
juanmahidalgo, line 76:
```suggestion
                <Dropdown.Item text={t('global.delete')} disabled />
```
Maybe? (nitpick)

src/components/ThirdPartyCollectionDetailPage/CollectionContextMenu/CollectionContextMenu.tsx
```diff
@@ -36,8 +36,15 @@ export default class CollectionContextMenu extends React.PureComponent<Props> {
     }
   }
 
+  canDelete() {
+    const { collection } = this.props
+    return !isLocked(collection) && !collection.isPublished
```
juanmahidalgo, line 41: This logic is perfect, but in order to work for TP collections, I think we need to update the `isPublished` attribute once the items are published. We forgot to add that in the collections reducer.
I believe it'd be something like:
```
case PUBLISH_THIRD_PARTY_ITEMS_SUCCESS: {
      const { collectionId } = action.payload
      return {
        ...state,
        data: {
          ...state.data,
          [collectionId]: {
            ...state.data[collectionId],
            isPublished: true
          }
        }
      }
    }
```
In the `collection/reducer.ts` file.
Without this change, the user could publish some items and the button will still be enabled, because the `isPublished` attribute keeps being `false`.

nicosantangelo: good catch! I kinda brushed over that as I though it'd be dealt by https://github.com/decentraland/builder/issues/1874 but I prefer adding it here 👍

juanmahidalgo (approved):
LGTM! 🚀

PR #1893, opened 2022-03-22
feat: redirect to the collection list after TP delete

### LINKED ISSUE #1885: Redirect to /collections when a collection is deleted
When the user deletes a collection, it remains in the same URL, showing the `Not found` page since the collection is now deleted. We should redirect back to `/collections`.

nicosantangelo:
Closes #1885

### DIFF
--- a/src/modules/collection/sagas.ts
+++ b/src/modules/collection/sagas.ts
@@ -91,7 +91,7 @@ import {
 } from 'modules/item/actions'
 import { areSynced, isValidText, toInitializeItems } from 'modules/item/utils'
 import { locations } from 'routing/locations'
-import { getCollectionId, getThirdPartyCollectionId } from 'modules/location/selectors'
+import { getCollectionId } from 'modules/location/selectors'
 import { BuilderAPI } from 'lib/api/builder'
 import { closeModal, CloseModalAction, CLOSE_MODAL, openModal } from 'modules/modal/actions'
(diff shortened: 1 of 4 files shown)


src/modules/collection/sagas.ts
```diff
@@ -255,7 +255,9 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
     try {
       yield call(() => builder.deleteCollection(collection.id))
       yield put(deleteCollectionSuccess(collection))
-      const collectionIdInUriParam: string = yield select(getCollectionId)
+
+      const collectionIdSelector = isTPCollection(collection) ? getThirdPartyCollectionId : getCollectionId
```
juanmahidalgo, line 259: What if we embed this logic in the selector? We can just use the `getCollectionId` and add the `thirdPartyCollectionDetail` to the match selector that we have:
```
const collectionIdMatchSelector = createMatchSelector<
  RootState,
  {
    collectionId: string
  }
>([locations.collectionDetail(), locations.thirdPartyCollectionDetail()])
```
I believe that if we add it there, we can re-use the same selector for both types of collection :) 

Just a suggestion, I'm ok with these changes too!

nicosantangelo: Woo I love this. Looks much better.

juanmahidalgo (approved):
looks great! 👏 👏

PR #1894, opened 2022-03-22
fix: Display the right status for TP items with changes

### LINKED ISSUE #1884: The item status orb doesn't show an unsynced status por TP items
When the item is under review, we've got three content hashes:
- content hash (the content hash in the blockchain)
- local_content_hash (the content hash in the builder server)
- the content hash in the curation

To correctly show the orb in an unsynced state, compare the local_content_hash with the content hash in the curation, if they don't match, change the color of the orb to yellow.

juanmahidalgo:
Closes #1884

image


nicosantangelo (approved):
LGTM

PR #1895, opened 2022-03-22
feat: Do not let push changes until TP collection is approved

### LINKED ISSUE #1878: Do not show the PUSH CHANGES button on under review items
The current implementation will show PUSH CHANGES on the PUBLISH button, but in reality all changes before the item is approved do should not update the curation.

We can discuss this further but the first idea to solve this is to only show PUSH CHANGES if the item is already approved. It's a bit different from what we do on Standard collections, where *published* collections/items are presented with PUSH CHANGES. But the change derives from the fact that published standard elements have already been committed to the blockchain, but TP items have not until approval

juanmahidalgo:
The following screenshots show an item with changes but since there was no approval yet, the push changes is blocked.

image

Closes #1878

### DIFF
--- a/src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
+++ b/src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
@@ -194,158 +194,154 @@ export default class ThirdPartyCollectionDetailPage extends React.PureComponent<
     const totalPages = Math.ceil(total / PAGE_SIZE)
 
     return (
-      <CollectionProvider id={collection.id}>
-        {() => (
-          <>
-            <Section>
-              <Row>
-                <Back absolute onClick={this.handleGoBack} />
-                <Narrow>
-                  <Row>
-                    <Column className="header-column">
-                      <Row className="header-row" onClick={this.handleEditName}>
-                        <Header size="huge" className="name">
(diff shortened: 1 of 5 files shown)


src/components/ThirdPartyCollectionDetailPage/ThirdPartyCollectionDetailPage.tsx
```diff
@@ -354,9 +350,19 @@ export default class ThirdPartyCollectionDetailPage extends React.PureComponent<
...
+            <LoggedInDetailPage
+              className="ThirdPartyCollectionDetailPage"
+              hasNavigation={!hasAccess && !isLoading && !isLoadingCollectionData}
+              isLoading={isLoading || isLoadingCollectionData}
+            >
+              {shouldRender && thirdParty ? this.renderPage(thirdParty) : <NotFound />}
+            </LoggedInDetailPage>
+          )
+        }}
+      </CollectionProvider>
```
nicosantangelo, line 365: Question!
I think we are leaving the `XXXProvider`s after the `LoggedInDetailPage` so we don't fetch the resource for an address that doesn't have the permissions to see it. In this case I can see us fetching the collection before knowing if the address is connected or not.

So! why are we loading here and not on `renderPage`'s level

nicosantangelo (approved):

PR #1898, opened 2022-03-23
fix: Publish changes without cheques

### LINKED ISSUE #1879: Recover from a failed deployment of TP items
We need to be able to recover or re-do a TP Items publication even if the deployment to the catalyst fails somehow. We could:
- Call the contract for the last hash cheque and compare it with the local root we just created (`tree.merkleRoot`)
- Add a new event to the contract so we can index the `messageHash` on thegraph and update thegraph (create related issues if necessary)

LautaroPetaccio:
This PR changes the `handleInitiateTPItemsApprovalFlow` so when the curator approves the TP items using an already consumed cheque, the cheque is not sent in the meta transaction to the ThirdParty registry.

Closes #1879


nicosantangelo (approved):
awesome

PR #1901, opened 2022-03-24
feat: update approve state

### LINKED ISSUE #1875: Update the TP curation page after a successful approval
After a curator finishes Approving a batch of third party items, the items that were just reviewed remain visible and the button for approval enabled.

We need to:
- [ ] Clean the batch of items
- [ ] Add a "No items left to review" notice in the left bar
- [ ] Disable the APPROVE button

nicosantangelo:
Closes #1875

### DIFF
--- a/src/modules/collection/sagas.ts
+++ b/src/modules/collection/sagas.ts
@@ -72,8 +72,7 @@ import {
   ApproveCollectionFailureAction,
   InitiateTPApprovalFlowAction,
   INITIATE_TP_APPROVAL_FLOW,
-  finishTPApprovalFlow,
-  finishApprovalFlow
+  finishTPApprovalFlow
 } from './actions'
 import { getMethodData, getWallet } from 'modules/wallet/utils'
 import { buildCollectionForumPost } from 'modules/forum/utils'
@@ -718,7 +717,6 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
         })
       )
     } catch (error) {
-      console.log(error)
       // Handle error at any point in the flow and show them
       const modalMetadata: ApprovalFlowModalMetadata<ApprovalFlowModalView.ERROR> = {
         view: ApprovalFlowModalView.ERROR,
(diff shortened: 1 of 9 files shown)


src/modules/collection/sagas.ts
```diff
@@ -706,18 +706,19 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
...
+      yield put(finishTPApprovalFlow(collection, itemsToApprove, newItemsCurations))
+
       yield put(
         openModal('ApprovalFlowModal', {
           view: ApprovalFlowModalView.SUCCESS,
           collection
         })
       )
     } catch (error) {
+      console.log(error)
```
LautaroPetaccio, line 721: To remove!
```suggestion
```

src/modules/collection/sagas.ts
```diff
@@ -706,18 +706,19 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
       }
 
       // 6. If the collection was approved but it had a pending curation, approve the curation
-
       const newItemsCurations: ItemCuration[] = yield call(updateItemCurationsStatus, itemsToApprove, CurationStatus.APPROVED)
-      console.log('newItemsCurations: ', newItemsCurations) // TODO: Add this to the success action that will override the curations in the state
 
       // 7. Success 🎉
+      yield put(finishTPApprovalFlow(collection, itemsToApprove, newItemsCurations))
```
LautaroPetaccio, line 712: Would you mind adding an assertion to the current tests to verify that this action was dispatched?

LautaroPetaccio (changes requested):

src/modules/curations/itemCuration/reducer.ts
```diff
@@ -65,7 +67,19 @@ export function itemCurationReducer(state: ItemCurationState = INITIAL_STATE, ac
...
+      const { collection, itemCurations } = action.payload
 
+      return {
+        ...state,
+        loading: loadingReducer(state.loading, action),
+        data: {
+          ...state.data,
+          [collection.id]: [...itemCurations]
+        },
+        error: null
```
LautaroPetaccio, line 80: The reducer clears the errors and the loading action, but this won't be necessary, as the loading reducer won't work (because of the name of the action) and I don't know about the errors, but I think that they shouldn't be cleared either, right?

LautaroPetaccio (changes requested):

src/modules/curations/itemCuration/reducer.ts
nicosantangelo: You're right! got confused with the entire approval flow

LautaroPetaccio (approved):

PR #1905, opened 2022-03-28
feat: collections list view

### LINKED ISSUE #1744: Add a list-view to the /collections list
We need to add a secondary view to the current grid view on the /collections page

A few things that might pose problems:
- If we paginate items inside collections, we'll not have the total amount of items to list
- Same goes for the published count

To mitigate this, we might want to re-use the logic we are trying to develop for the /curations page. There, we have an endpoint dedicated to item stats inside a collection (TP), at least for now: https://github.com/decentraland/builder-server/pull/382/files#diff-2383404c5077a9d06c1bf00c901433559690c98b0a5ba82a9c0e7979e6b249abR45

IF when developing this issue the endpoint is not available we could mock it and add a new issue to track the actual usage of it.

nicosantangelo:
Closes #1744

### DIFF
--- a/src/components/CollectionImage/CollectionImage.tsx
+++ b/src/components/CollectionImage/CollectionImage.tsx
@@ -1,3 +1,4 @@
+import classNames from 'classnames'
 import * as React from 'react'
 import { Loader } from 'decentraland-ui'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
@@ -9,6 +10,10 @@ import './CollectionImage.css'
 const MAX_IMAGES_TO_SHOW = 4
 
 export default class CollectionImage extends React.PureComponent<Props> {
+  static defaultProps = {
+    className: ''
+  }
+
   componentDidMount() {
     const { itemCount } = this.props
     if (itemCount) {
@@ -56,7 +61,7 @@ export default class CollectionImage extends React.PureComponent<Props> {
   }
 
   render() {
-    const { items, itemCount, isLoading } = this.props
+    const { items, className, itemCount, isLoading } = this.props
 
     if (isLoading || itemCount === undefined) {
       return (
@@ -70,7 +75,7 @@ export default class CollectionImage extends React.PureComponent<Props> {
     const hasNoItems = itemCount === 0
 
     return (
-      <div className="CollectionImage is-image">
+      <div className={classNames('CollectionImage', 'is-image', className)}>
         {hasNoItems ? (
           <div className="item-row empty">
             <div className="sparkles" />
--- a/src/components/CollectionImage/CollectionImage.tsx
+++ b/src/components/CollectionImage/CollectionImage.tsx
@@ -63,27 +63,16 @@ export default class CollectionImage extends React.PureComponent<Props> {
   render() {
     const { items, className, itemCount, isLoading } = this.props
 
-    if (isLoading || itemCount === undefined) {
-      return (
-        <div className="CollectionImage is-image">
+    return (
+      <div className={classNames('CollectionImage', 'is-image', className)}>
+        {isLoading || itemCount === undefined ? (
           <div className="item-row loading">
             <Loader active size="tiny" inline />
           </div>
-        </div>
-      )
-    }
-    const hasNoItems = itemCount === 0
-
-    return (
-      <div className={classNames('CollectionImage', 'is-image', className)}>
-        {hasNoItems ? (
+        ) : itemCount === 0 ? (
           <div className="item-row empty">
             <div className="sparkles" />
             <div>{t('collection_image.no_items')}</div>
-          </div>
-        ) : isLoading ? (
-          <div className="item-row loading">
-            <Loader active size="tiny" inline />
           </div>
         ) : (
           this.renderItemRows(items)
--- a/src/components/CollectionsPage/CollectionRow/CollectionRow.container.ts
+++ b/src/components/CollectionsPage/CollectionRow/CollectionRow.container.ts
@@ -1,0 +1,16 @@
+import { connect } from 'react-redux'
+import { push } from 'connected-react-router'
+import { RootState } from 'modules/common/types'
+import { getCollectionItems } from 'modules/item/selectors'
+import { OwnProps, MapStateProps, MapDispatchProps, MapDispatch } from './CollectionRow.types'
+import CollectionRow from './CollectionRow'
+
+const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => ({
+  items: getCollectionItems(state, ownProps.collection.id)
+})
+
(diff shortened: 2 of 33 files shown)


src/components/CollectionImage/CollectionImage.tsx
```diff
@@ -7,21 +8,25 @@ import { Props } from './CollectionImage.types'
...
-    let { items, isLoading } = this.props
+    let { className, items, isLoading } = this.props
 
     const firstItemRow = items.slice(0, 2)
     const secondItemRow = items.slice(2, 4)
     const itemRowStyle = { height: secondItemRow.length ? '50%' : '100%' }
 
     return (
-      <div className="CollectionImage is-image">
+      <div className={classNames('CollectionImage is-image', className)}>
```
LautaroPetaccio, line 27: Although it is the same, should we split them?
```suggestion
      <div className={classNames('CollectionImage', 'is-image', className)}>
```

src/components/CollectionImage/CollectionImage.tsx
```diff
@@ -7,21 +8,25 @@ import { Props } from './CollectionImage.types'
...
+    className: ''
+  }
+
   renderItemRow(items: Item[]) {
     return items.map((item, index) => <ItemImage key={index} item={item} />)
   }
 
   render() {
-    let { items, isLoading } = this.props
+    let { className, items, isLoading } = this.props
```
LautaroPetaccio, line 20: Isn't this a const?
```suggestion
    const { className, items, isLoading } = this.props
```

src/routing/locations.ts
```diff
@@ -27,7 +28,12 @@ export const locations = {
   sceneDetail: (projectId = ':projectId') => `/scenes/${projectId}`,
   collections: () => '/collections',
   itemDetail: (itemId = ':itemId') => `/items/${itemId}`,
-  collectionDetail: (collectionId = ':collectionId') => `/collections/${collectionId}`,
+  collectionDetail: (collectionId = ':collectionId', type: CollectionType = CollectionType.DECENTRALAND) =>
+    type === CollectionType.DECENTRALAND
+      ? `/collections/${collectionId}`
+      : type === CollectionType.THIRD_PARTY
+      ? locations.thirdPartyCollectionDetail(collectionId)
+      : '',
```
LautaroPetaccio, line 36: I'm thinking that it might be better to throw if the type doesn't match any of them to prevent any issues with the generation of urls.

src/modules/ui/collection/reducer.ts
```diff
@@ -0,0 +1,25 @@
...
+  switch (action.type) {
+    case SET_COLLECTION_PAGE_VIEW: {
+      return {
+        ...state,
+        view: action.payload.view
+      }
+    }
+    default:
+      return state
+  }
```
LautaroPetaccio, line 24: Would you mind adding tests for the reducer?

src/components/CollectionsPage/CollectionRow/CollectionRow.container.ts
```diff
@@ -0,0 +1,16 @@
+import { connect } from 'react-redux'
+import { push } from 'connected-react-router'
+import { RootState } from 'modules/common/types'
+import { getCollectionItems } from 'modules/item/selectors'
+import { OwnProps, MapStateProps, MapDispatchProps, MapDispatch } from './CollectionRow.types'
+import CollectionRow from './CollectionRow'
+
+const mapState = (state: RootState, ownProps: OwnProps): MapStateProps => ({
+  items: getCollectionItems(state, ownProps.collection.id)
```
LautaroPetaccio, line 9: The total amount of items is based on the items loaded in the UI, this might break with pagination, as it assumes that all the items are loaded. We could "fix" this by returning the total amount of items for the collections in the builder server.

src/components/CollectionsPage/CollectionRow/CollectionRow.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  render() {
+    const { collection, items } = this.props
+    const type = getCollectionType(collection)
+
+    return (
+      <Table.Row className={styles.CollectionRow} onClick={this.handleTableRowClick}>
+        <Table.Cell width={4}>
+          <div className={styles.imageColumn}>
+            <CollectionImage className={styles.image} collectionId={collection.id} />
```
LautaroPetaccio, line 25: It could happen, although pretty rare (if uploaded using the client), that a thumbnail doesn't have the required height and width and this will make the image look cropped.
Screen Shot 2022-03-31 at 11 47 18
This can be fixed by removing a couple of properties in the classes applied to the image by the `ImageWrapper`
```
.ItemImage.image-wrapper .image
```

src/components/CollectionsPage/ItemRow/ItemRow.tsx
```diff
@@ -0,0 +1,42 @@
...
+          </div>
+        </Table.Cell>
+        <Table.Cell width={3}>{t('global.item')}</Table.Cell>
+        <Table.Cell width={3}>-</Table.Cell>
+        <Table.Cell width={2}>
+          {item.isPublished ? (
+            <div className={styles.published}>
+              {t('global.published')} <Icon name="check" />
+            </div>
+          ) : null}
```
LautaroPetaccio, line 37: Items without collections can't be published, should we add this anyway?

LautaroPetaccio (changes requested):

src/components/CollectionsPage/ItemRow/ItemRow.tsx
nicosantangelo: Didn't realize! removing

src/components/ItemImage/ItemImage.tsx
```diff
@@ -18,9 +18,15 @@ export default class ItemImage extends React.PureComponent<Props> {
...
+    const shouldChangeImage = ['3491d5dd-7c0d-4287-8635-0eecf6957aec', 'b0d65d16-241e-4cd5-8468-ae8464b5869a'].includes(item.id)
+
     return (
       <div className={`ItemImage is-image image-wrapper ${className}`} style={getBackgroundStyle(item.rarity)}>
-        <img className="image" src={src || getThumbnailURL(item)} alt={item.name} />
+        <img
+          className="item-image"
+          src={shouldChangeImage ? 'https://i.stack.imgur.com/ek3Yc.png' : src || getThumbnailURL(item)}
+          alt={item.name}
+        />
```
LautaroPetaccio, line 29: You've left the testing images here hahaha.

LautaroPetaccio (approved):

src/components/ItemImage/ItemImage.tsx
nicosantangelo: what a dork this programmer is

PR #1917, opened 2022-03-30
feat: Enable TP item deployment

LautaroPetaccio:
This PR enables the deployment of TP items.
The deployment of TP items can be tested in `peer-uw-1.decentraland.zone`.

### DIFF
--- a/src/modules/collection/sagas.ts
+++ b/src/modules/collection/sagas.ts
@@ -627,7 +627,7 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
       }
 
       // Open the ApprovalFlowModal with the items to be approved
-      // 4. Make the transaction to the contract (update of the merkle tree root with the signature and its parameters)\
+      // 4. Make the transaction to the contract (update of the merkle tree root with the signature and its parameters)
       if (itemsToApprove.length > 0) {
         const { r, s, v } = ethers.utils.splitSignature(cheque.signature)
         const slot: Slot = {
--- a/src/modules/collection/sagas.ts
+++ b/src/modules/collection/sagas.ts
@@ -130,7 +130,6 @@ import {
   getCollectionType,
   getLatestItemHash,
   UNSYNCED_COLLECTION_ERROR_PREFIX,
-  isTPDeployEnabled,
   isTPCollection
 } from './utils'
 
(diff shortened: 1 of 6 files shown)


src/modules/collection/sagas.ts
```diff
@@ -628,7 +627,7 @@ export function* collectionSaga(builder: BuilderAPI, catalyst: CatalystClient) {
       }
 
       // Open the ApprovalFlowModal with the items to be approved
-      // 4. Make the transaction to the contract (update of the merkle tree root with the signature and its parameters)
+      // 4. Make the transaction to the contract (update of the merkle tree root with the signature and its parameters)\
```
nicosantangelo, line 630: 👀

src/modules/item/export.ts
```diff
@@ -153,14 +154,14 @@ function buildItemEntityMetadata(collection: Collection, item: Item): StandardCa
     name: item.name,
     description: item.description,
     collectionAddress: collection.contractAddress!,
-    rarity: item.rarity!,
-    i18n: [{ code: 'en', text: item.name }],
+    rarity: (item.rarity! as unknown) as Rarity,
+    i18n: [{ code: 'en', text: item.name }] as I18N[],
```
nicosantangelo, line 158: maybe we can save the casting here and use `Locale` everywhere?

```
i18n: [{ code: Locale.EN, text: dbItemMock.name }],
```

src/modules/item/export.ts
```diff
@@ -153,14 +154,14 @@ function buildItemEntityMetadata(collection: Collection, item: Item): StandardCa
     name: item.name,
     description: item.description,
     collectionAddress: collection.contractAddress!,
-    rarity: item.rarity!,
-    i18n: [{ code: 'en', text: item.name }],
+    rarity: (item.rarity! as unknown) as Rarity,
```
nicosantangelo, line 157: Same question as [here](https://github.com/decentraland/builder-server/pull/470#discussion_r839875557) in the server. If we won't change anything there, dismiss!

nicosantangelo (commented):
🔥

src/modules/item/export.ts
LautaroPetaccio: Done!

nicosantangelo (approved):
+1

PR #1926, opened 2022-04-04
feat: Add logic to fetch collection items for images if needed

juanmahidalgo:
## Screenshots

Fetches the images needed in they weren't fetched before:

image

image

### DIFF
--- a/src/components/CollectionImage/CollectionImage.tsx
+++ b/src/components/CollectionImage/CollectionImage.tsx
@@ -1,47 +1,76 @@
 import * as React from 'react'
 import { Loader } from 'decentraland-ui'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+import CollectionProvider from 'components/CollectionProvider'
 import { Item } from 'modules/item/types'
 import ItemImage from 'components/ItemImage'
 import { Props } from './CollectionImage.types'
 import './CollectionImage.css'
+
+const MAX_IMAGES_TO_SHOW = 4
 
 export default class CollectionImage extends React.PureComponent<Props> {
   renderItemRow(items: Item[]) {
     return items.map((item, index) => <ItemImage key={index} item={item} />)
   }
 
-  render() {
-    let { items, isLoading } = this.props
-
+  renderItemRows(items: Item[]) {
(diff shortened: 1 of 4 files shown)


src/lib/api/pagination.ts
```diff
@@ -1,5 +1,5 @@
 export const DEFAULT_PAGE = 1
-export const DEFAULT_PAGE_SIZE = 20
+export const DEFAULT_PAGE_SIZE = 40
```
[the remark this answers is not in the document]
LautaroPetaccio: I would use 60 as the default page size, as there are collections (there might be only one) with more than 50 items.

src/components/CollectionImage/CollectionImage.tsx
```diff
@@ -1,47 +1,76 @@
...
+          </div>
+        ) : null}
+      </>
+    )
+  }
+
+  render() {
+    const { items, itemCount, collectionId, isLoading } = this.props
+
+    if (isLoading || itemCount === null) {
```
LautaroPetaccio, line 40: Item count is typed as undefined or number, it should never take null as its value.

src/components/CollectionImage/CollectionImage.tsx
```diff
@@ -1,47 +1,76 @@
...
+            {({ items, isLoading }) => {
+              return isLoading ? (
+                <div className="item-row loading">
+                  <Loader active size="tiny" inline />
+                </div>
+              ) : (
+                this.renderItemRows(items)
+              )
+            }}
+          </CollectionProvider>
```
LautaroPetaccio, line 71: What do you think about receiving the action to load the first page of items of the collection here instead of using the `CollectionProvider`. The `CollectionProvider` will fetch the collection and the items, but we already have fetched the collection and the items are already received in this component.

LautaroPetaccio (changes requested):

src/components/CollectionImage/CollectionImage.tsx
juanmahidalgo: yeah, good idea, updated!

LautaroPetaccio (approved):

PR #1946, opened 2022-04-07
fix: Allow TP collection names to be longer than 32 characters

LautaroPetaccio:
This PR allows TP collection names to be longer than 32 characters, specifically 42 characters long, the longitude of an Ethereum address with `0x` included.


src/components/CollectionsPage/CollectionCard/CollectionCard.css
```diff
@@ -49,3 +49,9 @@
     max-width: 215px;
   }
 }
+
+.delete-collection-confirm .header {
```
nicosantangelo, line 53: I think this should either be pre-fixed (maybe not easy to do) or added as a top level class for confirm popups on src/themes/Confirm.css

What do you think?

LautaroPetaccio: I think that you're correct, but there's no global class for the `Confirm` modal. We're obligated to either create a wrapper component for it and style it or style it where we we want.
In this case, there shouldn't be any issues with changing the style of only this component, specially because only collection names changed to have a max of 42 characters, but items didn't, so (in theory) there shouldn't be an issue with other assets that are currently being deleted.

nicosantangelo: Ok I understand. My other option was to reuse `ConfirmDelete` and add this class there, so we can reuse it in the other components that use it. But I'm not sure if that Component is too generic for it to work.

I'll approve regardless

nicosantangelo (approved):

src/components/CollectionsPage/CollectionCard/CollectionCard.css
LautaroPetaccio: This change might be a bit bigger than we thought, I've created #1959 to tackle this issue.

PR #1954, opened 2022-04-08
fix: Allow TP manager to edit TP items

### LINKED ISSUE #1953: TP managers can't edit a TP item if they are not owners
## 🎉 Description

As we saw in the demo, as a TP manager, the user can't edit an item in the editor if they are not the owner. 

* [X] 🐛 This is a bug report.

juanmahidalgo:
Closes #1953

image

### DIFF
--- a/src/components/ItemEditorPage/RightPanel/RightPanel.container.ts
+++ b/src/components/ItemEditorPage/RightPanel/RightPanel.container.ts
@@ -5,7 +5,7 @@ import { getItem, getError as getItemError, isDownloading } from 'modules/item/s
 import { deleteItemRequest, downloadItemRequest, saveItemRequest, setCollection } from 'modules/item/actions'
 import { openModal } from 'modules/modal/actions'
 import { getSelectedItemId } from 'modules/location/selectors'
-import { getCollection } from 'modules/collection/selectors'
+import { getCollection, hasViewAndEditRights } from 'modules/collection/selectors'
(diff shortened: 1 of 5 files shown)


src/components/ItemEditorPage/RightPanel/RightPanel.container.ts
```diff
@@ -14,12 +14,13 @@ const mapState = (state: RootState): MapStateProps => {
...
   const address = getAddress(state) || ''
-
+  const collection = selectedItemId && selectedItem && selectedItem.collectionId ? getCollection(state, selectedItem.collectionId) : null
   return {
     address,
-    collection: selectedItemId && selectedItem && selectedItem.collectionId ? getCollection(state, selectedItem.collectionId) : null,
+    collection,
     selectedItem,
     selectedItemId,
+    canEditSelectedItem: !!selectedItem && !!collection && hasViewAndEditRights(state, address, collection),
```
LautaroPetaccio, line 23: Is it considering only the case which the collection exists? We'll be leaving behind the cases where we want to edit a single item that doesn't belong to a collection.

LautaroPetaccio (changes requested):

src/components/ItemEditorPage/RightPanel/RightPanel.container.ts
juanmahidalgo: Oh yeah, you're right! Updated to cover those cases as well!

LautaroPetaccio (approved):

PR #1956, opened 2022-04-08
fix: Show error when editing a URN fails

### LINKED ISSUE #1936: Failing to edit an item URN doesn't show anything in the UI
## 🎉 Description

If the user edits a URN in the `Edit URN` modal and after clicking save the request fails, the UI doesn't act upon it and shows an error. A common use case for this scenario is a user trying to change a URN for another one that is already present.

* [x] 🐛 This is a bug report.

## 🔢 Steps To Reproduce Issue

- Create a TP collection
- Add an item
- Add another item
- Edit one of the item URNs to be the same as the other item
- Click on save and the server will fail, but the UI won't show the error.

LautaroPetaccio:
We need better errors for the case that the URN is already in use and we should think about internationalizing the server errors.

Screen Shot 2022-04-08 at 15 41 41

Screen Shot 2022-04-08 at 15 40 13

Closes #1936


juanmahidalgo (approved):
looks great 👍

PR #1969, opened 2022-04-13
feat: Add check for a valid URN in asset.json when uploading files

### LINKED ISSUE #1955: Throw an error when uploading an item with an URN that's not pointing to the correct TP
## 🎉 Description

Throw an error, or set as failed, items that are uploaded with an `asset.json` that has URNs that don't match with the TP or TP collection id.

* [ ] 🐛 This is a bug report.
* [x] 📈 This is a feature request.

juanmahidalgo:
Closes #1955

image

### DIFF
--- a/src/lib/urn.ts
+++ b/src/lib/urn.ts
@@ -121,14 +121,14 @@ export function isThirdParty(urn?: string) {
   return decodedURN.type === URNType.COLLECTIONS_THIRDPARTY
 }
 
-export function decodeURN(urn: URN): DecodedURN {
+export function decodeURN<T extends URNType>(urn: URN): DecodedURN<T> {
   const matches = urnRegExp.exec(urn)
 
   if (!matches || !matches.groups) {
     throw new Error(`Invalid URN: "${urn}"`)
   }
 
-  return matches.groups as DecodedURN
+  return matches.groups as DecodedURN<T>
 }
 
 function getNetworkURNProtocol(network: Network) {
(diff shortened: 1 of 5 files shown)


src/lib/urn.ts
```diff
@@ -121,14 +121,14 @@ export function isThirdParty(urn?: string) {
   return decodedURN.type === URNType.COLLECTIONS_THIRDPARTY
 }
 
-export function decodeURN(urn: URN): DecodedURN {
+export function decodeURN<T extends URNType>(urn: URN): DecodedURN<T> {
```
LautaroPetaccio, line 124: Although it might get a bit confusing to use, `decodeURN` works without generics due to it being able to parse any kind of URN, that is, if you type it beforehand, your type might be wrong.
The correct way of using it is to decode the URN and then use an if clause on the type so, if the if clause is true about the type, then the block surrounding it will type the URN as the type it is.
For example (taken from `CreateSingleItemModal.tsx:255`):
```
       let decodedCollectionUrn: DecodedURN<any> | null = collection?.urn ? decodeURN(collection.urn) : null
        let urn: string | undefined
        if (
          decodedCollectionUrn &&
          decodedCollectionUrn.type === URNType.COLLECTIONS_THIRDPARTY &&
          decodedCollectionUrn.thirdPartyCollectionId
        ) {
          urn = buildThirdPartyURN(decodedCollectionUrn.thirdPartyName, decodedCollectionUrn.thirdPartyCollectionId, uuid.v4())
        }
```

LautaroPetaccio (changes requested):

src/lib/urn.ts
juanmahidalgo: awesome! thanks for the explanation 💯 , updated!

LautaroPetaccio (approved):

PR #1976, opened 2022-04-21
feat: Use the set alias actions from dcl dapps

LautaroPetaccio:
This PR changes the way we set the alias, using the code in the `decentraland-dapps` repository that contains the latest version of the catalyst-client.


nicosantangelo (approved):
LGTM

PR #2014, opened 2022-05-02
fix: Error when the curator has no profile data

### LINKED ISSUE #2013: [BUG] Error when the curator has no name associated
When accessing the edit curator modal, if the curator has no name associated, it throws an error and the modal is not opened.

juanmahidalgo:
Closes #2013

### DIFF
--- a/src/components/Modals/EditCurationAssigneeModal/EditCurationAssigneeModal.tsx
+++ b/src/components/Modals/EditCurationAssigneeModal/EditCurationAssigneeModal.tsx
@@ -62,7 +62,7 @@ export default class EditCurationAssigneeModal extends React.PureComponent<Props
                     value === UNASSIGN_PLACEHOLDER
                       ? `- ${t('curation_page.assign_modal.unassign')} -`
                       : value === address
-                      ? `${profiles[address]?.avatars[0].name} (${t('collection_row.you')})`
+                      ? `${profiles[address] ? profiles[address].avatars[0].name : value} (${t('collection_row.you')})`
                       : profiles[value]?.avatars[0].name ?? value
                 }))}
                 onChange={this.handleChangeAssignee}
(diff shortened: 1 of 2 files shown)


src/modules/curations/collectionCuration/utils.ts
```diff
@@ -15,9 +15,10 @@ export const getCollectionCurationState = (collection: Collection, curation: Col
       return CurationStatus.DISABLED
     } else if (curation && curation.status === 'rejected') {
       return CurationStatus.REJECTED
-    } else if (curation && curation.status === 'pending' && curation.assignee) {
-      return CurationStatus.UNDER_REVIEW
     }
   }
+  if (curation && curation.status === 'pending' && curation.assignee) {
```
juanmahidalgo, line 20: some collections that have `isApproved` as `true` (because they are in the catalyst) but they have a curation in pending and an assignee were listed as `to review` but they should be listed as `under_review`

src/components/Modals/EditCurationAssigneeModal/EditCurationAssigneeModal.tsx
```diff
@@ -62,7 +62,7 @@ export default class EditCurationAssigneeModal extends React.PureComponent<Props
                     value === UNASSIGN_PLACEHOLDER
                       ? `- ${t('curation_page.assign_modal.unassign')} -`
                       : value === address
-                      ? `${profiles[address].avatars[0].name} (${t('collection_row.you')})`
+                      ? `${profiles[address]?.avatars[0].name} (${t('collection_row.you')})`
```
fzavalia, line 65: If the profile for the given address does not exist,  (${t('collection_row.you')}) will have an empty space in the beginning, is this intentional?

juanmahidalgo: Oh yeah, good catch. Instead of that, I just added to show the address and the `(you)` text.

fzavalia (approved):

PR #2015, opened 2022-05-02
feat: Sorts the assignees list

### LINKED ISSUE #2018: Sort the assignee to be the first one in the list
## 🎉 Description

Put the address that is logged in in the first place so it's easier for the users to find themselfs

* [ ] 🐛 This is a bug report.

juanmahidalgo:
Closes #2018

### DIFF
--- a/src/components/CurationPage/CurationPage.tsx
+++ b/src/components/CurationPage/CurationPage.tsx
@@ -59,7 +59,7 @@ export default class CurationPage extends React.PureComponent<Props, State> {
   }
 
   renderAssigneeFilterDropdown = () => {
-    const { committeeMembers } = this.props
+    const { committeeMembers, wallet } = this.props
     const { assigneeFilter } = this.state
     return (
       <Dropdown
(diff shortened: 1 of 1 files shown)


src/components/CurationPage/CurationPage.tsx
```diff
@@ -59,15 +59,21 @@ export default class CurationPage extends React.PureComponent<Props, State> {
...
+    const { committeeMembers, wallet } = this.props
     const { assigneeFilter } = this.state
     return (
       <Dropdown
         direction="left"
         value={assigneeFilter}
         options={[
           { value: ALL_ASSIGNEES_KEY, text: t('curation_page.filter.all_assignees') },
-          ...committeeMembers.map(address => ({ value: address, text: <Profile textOnly address={address} /> }))
+          ...[
```
fzavalia, line 70: I don't think the first ... is necessary

src/components/CurationPage/CurationPage.tsx
```diff
@@ -59,15 +59,21 @@ export default class CurationPage extends React.PureComponent<Props, State> {
...
+            ...committeeMembers.filter(member => member === wallet.address),
```
fzavalia, line 71: I'd use a reduce instead of 2 filters. Something like:
```
commMembers.reduce((acc, member) => {
  if (member === wallet.address) {
    acc.unshift(member)
  } else {
    acc.push(member)
  }
  return acc
},[])
```

juanmahidalgo: good idea! updated!

src/components/CurationPage/CurationPage.tsx
juanmahidalgo: I believe it's still needed because the result of the `reduce` is an `[]` and it needs to be merged with `{ value: ALL_ASSIGNEES_KEY, text: t('curation_page.filter.all_assignees') }`. Right?

fzavalia (approved):

PR #2021, opened 2022-05-05
feat(raritiesWithOracle): Integrate builder server new rarity endpoints + New Modal

### LINKED ISSUE #1964: Fetch rarity converted prices when the user is going to publish a collection
Currently, rarities are only fetched once as their price was fixed and multiple requests are unnecessary.

```
  // PublishCollectionModal.tsx

  async componentDidMount() {
    const { collection, onClose, onFetchRarities, rarities } = this.props
    if (!collection) {
      onClose()
    }

    if (rarities.length === 0) {
      onFetchRarities()
    }
  }
```

Now that prices are pegged to a USD amount, prices should be fetched anytime the user wants to publish a collection to obtain a more precise value of what it will potentially cost.

### LINKED ISSUE #1965: Show message in publication modal explaining that prices are converted to MANA from a fixed USD price
As prices are not fixed in MANA with the RaritiesWithOracle contract, this conversion should be explained in the publication modal.

fzavalia:
Closes #1964 
Closes #1965


src/modules/item/sagas.ts
```diff
@@ -142,7 +150,33 @@ export function* itemSaga(legacyBuilder: LegacyBuilderAPI, builder: BuilderClien
 
   function* handleFetchRaritiesRequest() {
     try {
-      const rarities: Rarity[] = yield call([legacyBuilder, 'fetchRarities'])
+      const rarities: Rarity[] = yield call([legacyBuilder, legacyBuilder.fetchRarities])
```
juanmahidalgo, line 153: Just out of curiosity, could we move this logic to the backend and make just 1 call using the `currency` that we need to fetch? (`{currency: 'USD'}` if the env flag is one). I'm asking because is not clear why we need to fetch both endpoints and consolidate the data after that.

src/lib/api/builder.ts
```diff
@@ -782,8 +782,8 @@ export class BuilderAPI extends BaseAPI {
     return this.request('post', `/collections/${collection.id}/curation/post`, { forumPost })
   }
 
-  fetchRarities(): Promise<Rarity[]> {
-    return this.request('get', '/rarities')
+  fetchRarities({ currency }: { currency?: 'MANA' | 'USD' } = {}): Promise<Rarity[]> {
+    return this.request('get', `/rarities?currency=${currency || 'MANA'}`)
```
juanmahidalgo, line 786:
```suggestion
    return this.request('get', `/rarities?currency=${currency}`)
```

src/lib/api/builder.ts
```diff
@@ -782,8 +782,8 @@ export class BuilderAPI extends BaseAPI {
     return this.request('post', `/collections/${collection.id}/curation/post`, { forumPost })
   }
 
-  fetchRarities(): Promise<Rarity[]> {
-    return this.request('get', '/rarities')
+  fetchRarities({ currency }: { currency?: 'MANA' | 'USD' } = {}): Promise<Rarity[]> {
```
juanmahidalgo, line 785:
```suggestion
  fetchRarities({ currency }: { currency?: 'MANA' | 'USD' } = { currency: 'MANA' }): Promise<Rarity[]> {
```

src/components/Modals/PublishCollectionModalWithOracle/PublishCollectionModal.tsx
```diff
@@ -0,0 +1,293 @@
...
+    const { collection, onClose, onFetchRarities } = this.props
+
+    if (!collection) {
+      onClose()
+    }
+
+    onFetchRarities()
+  }
+
+  componentDidUpdate() {}
```
juanmahidalgo, line 27: Can we remove this if we won't use it?

src/modules/item/sagas.ts
fzavalia: I need both the price in USD and in MANA in the UI because according to the designs, I need to display the total cost in USD as well as in MANA.

At first I did it like this because I didn't want to change the type returned by the backend. depending on the flag, the only difference would be the price. Still, I ended up adding the currency field to the response.

I'll check out if I can convert it to a single endpoint while keeping compatibility.

fzavalia: Updated so now only 1 request is made to the server.

juanmahidalgo (approved):
🚀

PR #2022, opened 2022-05-05
feat: Add pagination support to the Curation page

juanmahidalgo:
## Guide to review these PRs:
- First https://github.com/decentraland/builder/pull/2035
- Second: https://github.com/decentraland/builder/pull/2027
- Third: https://github.com/decentraland/builder/pull/2022 (this one)

## This PR includes:
- [X] Adds pagination (using the server) to the Curation Page, using all the filter and sort combinations.

Closes #2020

PR #2027, opened 2022-05-06
feat: Add pagination to collections page

### LINKED ISSUE #2026: Add pagination to collections page
## 🎉 Description

We need to add the pagination parameters to the requests made by the collections page. Also, for the items tab.

* [ ] 🐛 This is a bug report.

juanmahidalgo:
## Guide to review these PRs:
- First https://github.com/decentraland/builder/pull/2035
- Second: https://github.com/decentraland/builder/pull/2027 (this one)
- Third: https://github.com/decentraland/builder/pull/2022 

## This PR includes

- [X] Adds pagination support for the collections page. Specifically for both Collections/Single Items tabs.

Closes #2026


src/components/ItemProvider/ItemProvider.tsx
```diff
@@ -3,15 +3,16 @@ import { Props } from './ItemProvider.types'
...
       onFetchItem(id)
     }
   }
 
   componentDidUpdate(prevProps: Props) {
-    const { id, item, onFetchItem } = this.props
-    if (id && id !== prevProps.id && !item) {
+    const { id, item, onFetchItem, isConnected } = this.props
+    const idChanged = id && id !== prevProps.id
+    if (isConnected && id && (idChanged || !item)) {
```
juanmahidalgo, line 15: @cazala I'm not sure the `idChanged` in here is needed. The `&& !item` might be enough. Is there away to navigate between items from the detail page? I want to simulate the `id` prop changing without unmounting the component

cazala: mm i don't think you can navigate between two item detail pages directly, but this `ItemProvider` might be used somewhere else too. The purpose of this component is to recibe an id and provide an item to its children. I think the reason it still fetches an item even if it is already in memory is to update it (in case there were changes since the last time it was fetched).

cazala (approved):

src/components/CollectionsPage/CollectionsPage.tsx
```diff
@@ -15,10 +29,30 @@ import CollectionRow from './CollectionRow'
...
   state = {
-    currentTab: TABS.COLLECTIONS
+    currentTab: TABS.COLLECTIONS,
+    page: 1
+  }
+
+  componentDidMount() {
+    const { onFetchCollections, address } = this.props
+    // fetch if already connected
+    if (address && address) {
```
cazala, line 43: oops, this should be only `address` right?

cazala (changes requested):
A few things i noticed after testing:

1. After creating an item from the empty state, you are not taken to the item detail (maybe this is because that was fixed in master but not in this PR)
2. The empty state for collections shows `0 results` but the one for items shows only `results`: 

Screen Shot 2022-05-13 at 17 01 30

3. Deleting a collection or an item doesn't refresh the page (notice the `2 results`):
Screen Shot 2022-05-13 at 17 05 21

src/components/CollectionsPage/CollectionsPage.tsx
```diff
@@ -29,7 +29,7 @@ import CollectionRow from './CollectionRow'
 import { Props, TABS } from './CollectionsPage.types'
 import './CollectionsPage.css'
 
-const PAGE_SIZE = 20
+const PAGE_SIZE = 6
```
cazala, line 32: doesn't this make the second row cut in half? 
Screen Shot 2022-05-16 at 18 54 42

(disregard the broken images, that's only on my local env)

juanmahidalgo: Oops, I forgot to get that back to `20`, I was testing locally with a smaller page size. I put it back to 20!

cazala (approved):

PR #2035, opened 2022-05-10
feat: Paginate Items and Collections from Left Panel

### LINKED ISSUE #2028: Deprecate fetch collections on log in event
Now that we have the endpoint paginated, we should move the fetch logic to the components, so we can fetch the collections with a page size.

Take a look at these components:
- Collection Dropdown
- LeftPanel (from the item editor)

juanmahidalgo:
## Guide to review these PRs:
- First https://github.com/decentraland/builder/pull/2035 (this one)
- Second: https://github.com/decentraland/builder/pull/2027 
- Third: https://github.com/decentraland/builder/pull/2022 

## This PR includes:

- [X] Adds Tabs to the Left Panel for Collection/Items
- [X] Paginates both Left Panel sections
- [X] Removes the fetch collection call after login in
- [X] Add call to get all collections when opening the CollectionsDropdown (the one that the user uses to move an Item to a collection, we need to show ALL the collections in that dropdown) 

Closes #2028


cazala:
Hey @juanmahidalgo i found an issue which i dunno if it's related to the frontend or the backend. When I go to the `Single Items` tab, i'm always getting all the orphan items in the DB i think (it doesn't matter which account i'm signed in with, i always get this same amount of items, and i'm not the creator of them all):

The request is this one:

```
http://localhost:5001/v1/0xe3547c273ce6c458216aac599b38bdfa4d09a39d/items?page=1&limit=20&collectionId=null
```

The response has always `1572` results no matter which account im using.

And it looks like this (im using the `builder-server` from [this PR](https://github.com/decentraland/builder-server/pull/514) I filled the database with a dump from dev, and the images are broken because im using nico's bucket):
Screen Shot 2022-05-11 at 16 22 45

cazala:
I also noticed that if i refresh on an item detail i get a not found

juanmahidalgo:
thanks @cazala!! The `items` query was messed up, I fixed it in the other PR, so now it's returning just the items of the address and tp items with manager access.
I updated the logic to fix the Not found when refreshing in the item detail page as well, adding the `ItemProvider` so it fetches the item when rendering the page

cazala:
It's working fine now, the only thing is when i refresh the editor page I see no items:
Screen Shot 2022-05-13 at 16 41 05

cazala (approved):
awesome 🚀 !

PR #2049, opened 2022-05-12
feat: Add open in world in the collection detail for items

LautaroPetaccio:
This PR does the following:
- [x] Adds dropdown item menu to the rows of items in the collection detail & third party collection details so the creators can see a specific item in world.
- [x] Changes the link of the `JumpIn` button in the `TopPanel` of the item editor so it loads for the third party collections, the first 30-40 items that it has loaded for curation. This will at least make curators be capable of reviewing the first batch of items in bulk.
- [x] Changes the `See in world` button in the third party collection detail page so the creators can see the the items that are loaded in world by opening the explorer site with the item ids of the items currently loaded.

### DIFF
--- a/src/components/CollectionDetailPage/CollectionItem/CollectionItem.tsx
+++ b/src/components/CollectionDetailPage/CollectionItem/CollectionItem.tsx
@@ -8,7 +8,7 @@ import { locations } from 'routing/locations'
 import { preventDefault } from 'lib/preventDefault'
 import { isComplete, isFree, canMintItem, canManageItem, getMaxSupply } from 'modules/item/utils'
 import ItemStatus from 'components/ItemStatus'
-import { isLocked } from 'modules/collection/utils'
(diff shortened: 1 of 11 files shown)


src/components/CollectionDetailPage/CollectionItem/CollectionItem.tsx
```diff
@@ -137,6 +145,7 @@ export default class CollectionItem extends React.PureComponent<Props> {
                 >
                   <Dropdown.Menu>
                     <Dropdown.Item text={t('collection_item.see_details')} as={Link} to={locations.itemDetail(item.id)} />
+                    <Dropdown.Item text={t('collection_context_menu.see_in_world')} onClick={this.handleNavigateToExplorer} />
```
cazala, line 148: just as a heads up, i think this can be done easier using the `href` prop instead of `onClick`

cazala (approved):

PR #2059, opened 2022-05-13
feat: Hide body shape option from editor

fzavalia:
Prevent the user from selecting body shape as a hideable part

### DIFF
--- a/src/components/CollectionDetailPage/CollectionContextMenu/CollectionContextMenu.tsx
+++ b/src/components/CollectionDetailPage/CollectionContextMenu/CollectionContextMenu.tsx
@@ -19,7 +19,7 @@ export default class CollectionContextMenu extends React.PureComponent<Props> {
 
   handleNavigateToExplorer = () => {
     const { collection } = this.props
-    this.navigateTo(getExplorerURL(collection), '_blank')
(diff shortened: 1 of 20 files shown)


cazala (changes requested):
mm i think this should be done here, where we already hide the head category: https://github.com/decentraland/builder/blob/master/src/modules/item/utils.ts#L337-L343

fzavalia:
> mm i think this should be done here, where we already hide the head category: https://github.com/decentraland/builder/blob/master/src/modules/item/utils.ts#L337-L343

You are right!

LautaroPetaccio (approved):

LautaroPetaccio (changes requested):
Dismissed my review due to the PR already having a review.

fzavalia:
> Dismissed my review due to the PR already having a review.

Reviewedn't

fzavalia:
> mm i think this should be done here, where we already hide the head category: https://github.com/decentraland/builder/blob/master/src/modules/item/utils.ts#L337-L343

The thing is that getWearableCategories is not used to get what can be overriden and what not.

cazala (approved):

PR #2061, opened 2022-05-17
feat: Track wallet connection with dapps lib function

fzavalia:


fzavalia:
Gonna reopen in another PR

PR #2074, opened 2022-05-20
fix: Queue TP deployments and curation updates

LautaroPetaccio:
This PR does the following:
- Creates a new view in the ApprovalFlowModal to show a limited set of items to be deployed and to execute the new batched deployment actions
- Changes the way the TP approval flow works by:
  - Moving the procedure that builds the entities of the items next to the deployment of them, removing the necessity of storing all of the items contents before deploying.
  - Batching the deployment so they don't get rejected by the Catalyst or the builder-server.
  - Batching the update of the curations so they don't get rejected by the builder-server.


src/modules/thirdParty/sagas.ts
```diff
@@ -215,4 +224,34 @@ export function* thirdPartySaga(builder: BuilderAPI) {
...
+      if (!identity) {
+        throw new Error('Invalid Identity')
+      }
+
+      const promisesOfItemsBeingDeployed: (() => Promise<DeploymentPreparationData>)[] = items.map((item: Item) => async () => {
+        console.log('Tried to execute this')
+        const entity: DeploymentPreparationData = await buildTPItemEntity(catalyst, builder, collection, item, tree, hashes[item.id])
+        console.log('Entity built', entity.entityId)
+        await catalyst.deployEntity({ ...entity, authChain: Authenticator.signPayload(identity, entity.entityId) })
+        console.log('Deployed entity', entity.entityId)
```
cazala, line 245: oopsie

cazala (approved):

PR #2077, opened 2022-05-23
feat: Maintenance by feature flag

fzavalia:
Added features module polling for feature values every 60 seconds


src/modules/features/selectors.ts
```diff
@@ -0,0 +1,14 @@
+import { getIsFeatureEnabled } from 'decentraland-dapps/dist/modules/features/selectors'
+import { ApplicationName } from 'decentraland-dapps/dist/modules/features/types'
+import { RootState } from 'modules/common/types'
+import { FeatureName } from './types'
+
+export const getIsMaintenanceEnabled = (state: RootState) => {
```
LautaroPetaccio, line 6: Would you mind adding a test for this selector?

LautaroPetaccio (changes requested):

src/modules/features/selectors.ts
fzavalia: surelicious

fzavalia: Added tests

Not testing the internal function that is being used because it is tested thoroughly in the lib https://github.com/decentraland/decentraland-dapps/blob/master/src/modules/features/selectors.spec.ts

Issue #2044, opened 2022-05-11
Cleanup rarities with oracle feature flag when the feature has been fully deployed on all envs


fzavalia:

fzavalia:
The main things to remove is the function to get the rarities contract according to the flag and just keep the publish collection modal with oracle.

PR #2079, opened 2022-05-23
feat: Feature flags for Old Rarities / Rarities pegged to USD

fzavalia:


src/modules/collection/utils.ts
```diff
@@ -136,8 +138,12 @@ export function getCollectionFactoryContract(chainId: ChainId) {
...
-    ? getContract(ContractName.RaritiesWithOracle, chainId)
-    : getContract(ContractName.Rarities, chainId)
+export function* getRaritiesContract(chainId: ChainId) {
+  const enabled: boolean = yield select(getIsRaritiesWithOracleEnabled)
+
+  if (enabled) {
+    return getContract(ContractName.RaritiesWithOracle, chainId)
+  }
+
+  return getContract(ContractName.Rarities, chainId)
```
LautaroPetaccio, line 148: What if instead of being a generator, it is a selector? It's used from a sagas, so we could use the `select` effect there and make this simpler.
It should be moved to selectors too.

```suggestion
export function getRaritiesContract(state: RootState, chainId: ChainId) {
  const enabled: boolean = getIsRaritiesWithOracleEnabled(state)

  if (enabled) {
    return getContract(ContractName.RaritiesWithOracle, chainId)
  }

  return getContract(ContractName.Rarities, chainId)
```

src/components/Modals/PublishCollectionModalOld/PublishCollectionModal.container.ts
```diff
@@ -0,0 +1,33 @@
+import { connect } from 'react-redux'
```
LautaroPetaccio, line 1: Would you mind (if it doesn't exist) creating a ticket to remove this modal?

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
```diff
@@ -1,257 +1,10 @@
...
-import { t, T } from 'decentraland-dapps/dist/modules/translation/utils'
-import { ItemRarity } from 'modules/item/types'
-import { getBackgroundStyle } from 'modules/item/utils'
-import { emailRegex } from 'lib/validators'
-import { Props, State } from './PublishCollectionModal.types'
-import './PublishCollectionModal.css'
+import React from 'react'
+import { Props } from './PublishCollectionModal.types'
+import PublishCollectionModalOld from '../PublishCollectionModalOld'
+import PublishCollectionModalWithOracle from '../PublishCollectionModalWithOracle'
```
LautaroPetaccio, line 4: I don't know what's going on with Github, but I'm not seeing this file in this PR.

LautaroPetaccio (changes requested):

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
fzavalia: https://github.com/decentraland/builder/issues/2044 The issue was already created but I added and extra comment for better context

src/components/Modals/PublishCollectionModalOld/PublishCollectionModal.container.ts
fzavalia: https://github.com/decentraland/builder/issues/2044 The issue was already there but I added an extra comment for more context

src/modules/collection/utils.ts
fzavalia: I think that is a good idea. Had my doubts about adding a selector that returns something that does not belong directly to the store but there is actually no rule against it

src/components/Modals/PublishCollectionModal/PublishCollectionModal.tsx
fzavalia: The whole file is just:
```
import React from 'react'
import { Props } from './PublishCollectionModal.types'
import PublishCollectionModalOld from '../PublishCollectionModalOld'
import PublishCollectionModalWithOracle from '../PublishCollectionModalWithOracle'

const PublishCollectionModal = ({ isRaritiesWithOracleEnabled, ...rest }: Props) => {
  return isRaritiesWithOracleEnabled ? <PublishCollectionModalWithOracle {...rest} /> : <PublishCollectionModalOld {...rest} />
}

export default React.memo(PublishCollectionModal)

```

src/modules/collection/utils.ts
LautaroPetaccio: That is a great question, that seem to be a gray area. What made me recommend this change is that I tend to avoid whenever it's possible, `selects` in generators that are not action handlers. This is because to the problem of these types of function possibly being re-used with un-expected side-effects due to doing a select in a different moment in time where the state might have changed.

LautaroPetaccio (approved):

src/modules/collection/utils.ts
fzavalia: Moved it

src/modules/collection/utils.spec.ts
```diff
@@ -164,25 +161,3 @@ describe('when checking if a collection is of type third party', () => {
     })
   })
 })
-
```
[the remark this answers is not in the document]
fzavalia: Added the test in the selectors test file

PR #2085, opened 2022-05-25
fix: show the correct status for TP collections at curation page

### LINKED ISSUE #2082: TP Collections not showing the right status in the Curations Page
## 🎉 Description

They are showing the status "Under Review" even if they are already approved.

* [X] 🐛 This is a bug report.

juanmahidalgo:
Closes #2082

image

image

### DIFF
--- a/src/modules/curations/collectionCuration/utils.ts
+++ b/src/modules/curations/collectionCuration/utils.ts
@@ -1,11 +1,11 @@
 import { Collection } from 'modules/collection/types'
-import { hasReviews } from 'modules/collection/utils'
+import { hasReviews, isTPCollection } from 'modules/collection/utils'
 import { CurationStatus } from '../types'
 import { CollectionCuration } from './types'
 
 export const getCollectionCurationState = (collection: Collection, curation: CollectionCuration | null) => {
   if (collection.isApproved) {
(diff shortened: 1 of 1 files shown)


src/modules/curations/collectionCuration/utils.ts
```diff
@@ -1,11 +1,11 @@
 import { Collection } from 'modules/collection/types'
-import { hasReviews } from 'modules/collection/utils'
+import { hasReviews, isTPCollection } from 'modules/collection/utils'
 import { CurationStatus } from '../types'
 import { CollectionCuration } from './types'
 
 export const getCollectionCurationState = (collection: Collection, curation: CollectionCuration | null) => {
   if (collection.isApproved) {
-    if (!curation || curation.status === 'approved') {
+    if (!curation || curation.status === 'approved' || isTPCollection(collection)) {
```
LautaroPetaccio, line 8: After reviewing this change it think that it might cause issues if there's a push changes to be reviewed or if there's a new publication. If we keep saying that it's approved, it might be confusing for the curators that are about to review it. I'm thinking that we might want to update or re-create the collection curation each time a push changes occurs or there's a new publication.

LautaroPetaccio (changes requested):

PR #2096, opened 2022-05-30
fix: Fix the math for calcuating the approval threshold for TP collections

### LINKED ISSUE #2095: Shuffle hangs when curating TP collections with lower than 100 items & items to review count is incorrect
## 🎉 Description

Three issues occur when reviewing Third Party collections:
- To check how many items need to be curated, the UI is multiplying the threshold (0,01) and later getting the floor of that to get the amount of items available to curate. This is causing collections with < 100 items to end up having to curate 0 items, as (for example) 50 items * 0,01 = 0,5 and the floor of 0,5 is 0.
- The shuffle button is active when there's no more items to be curated.
- The DAO proposal says that a minimum amount of 50 items must be curated per collection and then a variable amount, based in a percentage of items, up to 300 items. We're just applying a percentage. This should be changed.

* [x] 🐛 This is a bug report.

## 🔢 Steps To Reproduce

1. Create a third party collection
2. Add 5 items and publish them
3. Go to the curation tab, shuffle should be active and the counter should mark 0/0 (both incorrect)
4. Click on shuffle and the website will hang.

juanmahidalgo:
Closes #2095


src/modules/collection/utils.ts
```diff
@@ -135,3 +136,14 @@ export function isTPCollection(collection: Collection): boolean {
 export function getCollectionFactoryContract(chainId: ChainId) {
   return getContract(ContractName.CollectionFactoryV3, chainId)
 }
+
+export const getTPThresholdToReview = (totalItems: number) => {
+  // Reference: https://governance.decentraland.org/proposal/?id=f69c4d40-aaaf-11ec-87a7-6d2a41508231
+  // Max to review, 300 items, min to review is 50 or 1% of the collection
+  if (totalItems! < MIN_TP_ITEMS_TO_REVIEW) {
```
LautaroPetaccio, line 143: Does it have a `!` because it used to be undefined?
```suggestion
  if (totalItems < MIN_TP_ITEMS_TO_REVIEW) {
```

src/modules/collection/utils.ts
```diff
@@ -135,3 +136,14 @@ export function isTPCollection(collection: Collection): boolean {
...
+    return totalItems
+  } else if (totalItems! > MIN_TP_ITEMS_TO_REVIEW && totalItems! < MAX_TP_ITEMS_TO_REVIEW) {
+    return MIN_TP_ITEMS_TO_REVIEW
+  }
+  return Math.min(Math.floor(totalItems! * TP_TRESHOLD_TO_REVIEW), MAX_TP_ITEMS_TO_REVIEW)
```
LautaroPetaccio, line 148: I think that the whole function can be reduced to:
```suggestion
  return Math.min(Math.max(Math.floor(totalItems! * TP_TRESHOLD_TO_REVIEW), MIN_TP_ITEMS_TO_REVIEW), MAX_TP_ITEMS_TO_REVIEW)
```
That is, if the minimum amount of items to review according to the threshold is less than 50, then review 50 items, but if it's greater than 50, review as many items as 300.

LautaroPetaccio (changes requested):

src/modules/collection/utils.spec.ts
```diff
@@ -161,3 +162,47 @@ describe('when checking if a collection is of type third party', () => {
...
+  })
+  describe('and the collection has more than the minium to review', () => {
+    beforeEach(() => {
+      totalItems = MIN_TP_ITEMS_TO_REVIEW + 10
+    })
+    it('should return the minimum quantity to review', () => {
+      expect(getTPThresholdToReview(totalItems)).toBe(MIN_TP_ITEMS_TO_REVIEW)
+    })
+  })
+  describe('and the collection somewhere in betwen of the minimun and the minimum percentage', () => {
```
LautaroPetaccio, line 184:
```suggestion
  describe('and the collection has an amount of items somewhere in between of the minimum amount of items and the minimum percentage', () => {
```

src/modules/collection/utils.spec.ts
```diff
@@ -161,3 +162,47 @@ describe('when checking if a collection is of type third party', () => {
...
+    beforeEach(() => {
+      totalItems = 500
+    })
+    it('should return the minimum quantity to review', () => {
+      expect(getTPThresholdToReview(totalItems)).toBe(50)
+    })
+  })
+  describe('and the collection somewhere in betwen of the 1% and the max percentage', () => {
```
LautaroPetaccio, line 192:
```suggestion
  describe('and the collection has an amount of items somewhere in between the 1% and the max percentage', () => {
```

LautaroPetaccio (changes requested):

LautaroPetaccio (approved):

PR #2122, opened 2022-06-06
fix: make emotes follow the wearable schema to allow deployments

cazala:
Currently we can't deploy emotes because the metadata doesn't follow the wearable schema (missing hides/replaces/overrides and category is a different enum).

Screen Shot 2022-06-06 at 17 03 17

So I had to hack it to make it pass through. I deployed and minted the pride emotes in ropsten and tested them on play.decentraland.org?NETWORK=ropsten and they worked fine:

 
Screen Shot 2022-06-06 at 17 44 05

Screen Shot 2022-06-06 at 17 45 50

Eventually once the EMOTE entity type is live these emotes will need to be redeployed with the new entity type, and that one is going to have it's own schema so these hacks could be removed.

### DIFF
--- a/src/modules/item/export.ts
+++ b/src/modules/item/export.ts
@@ -176,6 +176,15 @@ function buildItemEntityMetadata(collection: Collection, item: Item): StandardCa
     catalystItem.emoteDataV0 = {
       loop: (item.data as EmoteData).category === EmoteCategory.LOOP
     }
+    // add missing properties from wearable schema so catalyst wont reject the deployment
+    catalystItem.data.category = WearableCategory.HAT
+    catalystItem.data.hides = []
+    catalystItem.data.replaces = []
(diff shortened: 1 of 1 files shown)


src/modules/item/export.ts
```diff
@@ -176,6 +176,15 @@ function buildItemEntityMetadata(collection: Collection, item: Item): StandardCa
     catalystItem.emoteDataV0 = {
       loop: (item.data as EmoteData).category === EmoteCategory.LOOP
     }
+    // add missing properties from wearable schema so catalyst wont reject the deployment
+    catalystItem.data.category = WearableCategory.HAT
```
cazala, line 180: I used the `hat` category randomly, it's not used by the explorer aparently, any wearable category would work, but leaving the `simple` or `loop` values would make the schema reject it.

LautaroPetaccio (approved):

PR #2124, opened 2022-06-08
ci: use config instead of env vars

cazala:
This PR removes all the environment variables that were different on each environment and moved them to the `config` object, which uses `@dcl/ui-env` to pick the right configuration. All those env variables were moved to the `src/config/env` folder under three different jsons: `dev.json`, `stg`.json` and `prod.json`.


.env.example
```diff
@@ -1,64 +1,7 @@
...
-REACT_APP_RESOLVER_CONTRACT_ADDRESS=0x4976fb03c32e5b8cfe2b6ccb31c09ba78ebaba41
-REACT_APP_CONTROLLER_ADDRESS=0xd4e92f6ea18412dbae8b2d2b69e498a2d1064fa2
-REACT_APP_REGISTRAR_ADDRESS=0xeb6f5d94d79f0750781cc962908b161b95192f53
-
-# Others
-
-REACT_APP_PROMO_URL=
-REACT_APP_SHARE_SCENE_URL=
+NODE_PATH=development
+REACT_APP_DCL_DEFAULT_ENV=dev
```
LautaroPetaccio, line 2: Should we include the other env variables that the envs have?
```
GEN_STATIC_LOCAL=true
PUBLIC_URL=/
```

.env.example
```diff
@@ -1,64 +1,7 @@
...
-REACT_APP_ENS_CONTRACT_ADDRESS=0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e
-REACT_APP_RESOLVER_CONTRACT_ADDRESS=0x4976fb03c32e5b8cfe2b6ccb31c09ba78ebaba41
-REACT_APP_CONTROLLER_ADDRESS=0xd4e92f6ea18412dbae8b2d2b69e498a2d1064fa2
-REACT_APP_REGISTRAR_ADDRESS=0xeb6f5d94d79f0750781cc962908b161b95192f53
-
-# Others
-
-REACT_APP_PROMO_URL=
-REACT_APP_SHARE_SCENE_URL=
+NODE_PATH=development
```
LautaroPetaccio, line 1: Shouldn't `NODE_PATH` be src?
```suggestion
NODE_PATH=src
```
Shouldn't it be in the other environment files?

src/config/env/prod.json
```diff
@@ -0,0 +1,33 @@
+{
+  "CHAIN_ID": "1",
+  "LOCAL_STORAGE_KEY": "builder",
+  "EMAIL_SERVER_URL": "https://decentraland.org/subscribe",
+  "PEER_URL": "https://peer-lb.decentraland.org",
```
LautaroPetaccio, line 5: We could change it to peer:
```suggestion
  "PEER_URL": "https://peer.decentraland.org",
```

src/config/env/stg.json
```diff
@@ -0,0 +1,33 @@
+{
+  "CHAIN_ID": "1",
+  "LOCAL_STORAGE_KEY": "builder",
+  "EMAIL_SERVER_URL": "https://decentraland.org/subscribe",
+  "PEER_URL": "https://peer-lb.decentraland.org",
```
LautaroPetaccio, line 5: We could change it to peer:
```suggestion
  "PEER_URL": "https://peer.decentraland.org",
```

src/modules/analytics/delighted.ts
```diff
@@ -1,9 +1,9 @@
 import { delay, call } from 'redux-saga/effects'
 import { getAnalytics } from 'decentraland-dapps/dist/modules/analytics/utils'
-import { env } from 'decentraland-commons'
+import { config } from 'config'
 
 const DELIGHTED_DELAY = 180 * 1000
-const DELIGHTED_API_KEY = env.get('REACT_APP_DELIGHTED_API_KEY', '')
+const DELIGHTED_API_KEY = config.get('DELIGHTED_API_KEY', '')
```
LautaroPetaccio, line 6: Should we remove delighted if we're not using it?

src/modules/analytics/hotjar.ts
```diff
@@ -1,7 +1,7 @@
-import { env } from 'decentraland-commons'
+import { config } from 'config'
 import { isProduction } from 'lib/environment'
 
-const HOTJAR_ID = parseInt(env.get('REACT_APP_HOTJAR_ID', ''), 10)
+const HOTJAR_ID = parseInt(config.get('HOTJAR_ID', ''), 10)
```
LautaroPetaccio, line 4: Should we remove hotjar if we're not using it?

vercel.json
```diff
@@ -0,0 +1,22 @@
+{
+  "version": 2,
+  "name": "builder",
+  "build": {
+    "env": {
+      "NODE_PATH": "src",
+      "NODE_ENV": "development",
+      "GEN_STATIC_LOCAL": "true",
+      "PUBLIC_URL": "/",
+      "REACT_APP_DCL_DEFAULT_ENV": "dev"
```
LautaroPetaccio, line 10: This is only defined in Vercel and in the examples, what is it for?

src/modules/collection/utils.ts
```diff
@@ -81,7 +81,8 @@ export function getCollectionSymbol(collection: Collection) {
 
 export function toCollectionObject(collections: Collection[]) {
   return collections.reduce((obj, collection) => {
-    obj[collection.id] = utils.omit<Collection>(collection, ['items'])
+    const { items: _, ...rest } = collection as Collection & { items?: any[] }
```
LautaroPetaccio, line 84: Does this work?
```suggestion
    const { items: _, ...rest } = collection as Omit<Collection, 'items'>
```

.env.example
cazala: yes 😅

.env.example
cazala: sure lets add those (they are not necessary for local development but they are used to generate the build for deployment)

vercel.json
cazala: It overrides the default environment used. The behavioir is:

- Look for an `env` query param and use that as the env
- If no param present, check the TLD (if .io/.zone -> DEV, .today/.net -> STG, .org -> PROD)
- If no TLD found, use default (which is by default PROD, but with this env we override it to DEV).

This is so vercel deployments use by default the dev env (we can always change it via query param).

The same applies for `localhost`, if no query param it will default to PROD, but with that env var we can make it default to DEV.

src/modules/collection/utils.ts
cazala: It's not an omit actually, we are adding an optional property to the type (`Collection` doesn't have  an `items` property)

src/modules/analytics/delighted.ts
cazala: done ✅

.env.example
cazala: done ✅

src/modules/analytics/hotjar.ts
cazala: done ✅

LautaroPetaccio (approved):

PR #2131, opened 2022-06-10
ci: remove old infra and decentraland-commons

cazala:
This PR removes the files from the old CI and `decentraland-commons` since most of its usage was replaced with `@dcl/ui-env`


setupEnv.sh
```diff
@@ -1,6 +0,0 @@
-#! /bin/bash
-
-set -e
-
-source dcl-env
-cp .ci/.env.${ENVIRONMENT} .env
```
LautaroPetaccio, line 6: We don't have the .env files per environment anymore, but, we still need a .env file to be loaded to load some variables, for example, the `PUBLIC_PATH`. We still need something to build the `.env` file, don't we?

LautaroPetaccio (changes requested):

setupEnv.sh
cazala: This was only used by the old CI, the current one doesn't need this script, it uses the `prebuild.js` one: https://github.com/decentraland/builder/blob/master/scripts/prebuild.js

LautaroPetaccio (approved):

PR #2147, opened 2022-07-01
fix: use as alias

nicosantangelo:
There was very niche dependency problem causing EntityType to be `undefined` **inside** decentraland-dapps. Because the (not major 😠 ) version of @dcl/schemas removes EntityType I had to force-install 4.14.0 for it to work.

That said, I'm now hitting this error:

`"Failed to fetch https://peer.decentraland.zone/content/contents/QmV6f9iNyPGTFsFrWuv5AnQkjWcHkt8WXNZnjXSAGtofdy. Got status 404. Response was ''"`

in dev. Pls reviewer could you check with your names? as I'm not sure if this is a me thing

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "@dcl/content-hash-tree": "^1.1.3",
         "@dcl/crypto": "^3.0.1",
         "@dcl/hashing": "^1.1.0",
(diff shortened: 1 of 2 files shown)


LautaroPetaccio (approved):

LautaroPetaccio:
I've tested it and it's working.

Issue #2143, opened 2022-06-27
Filter the 'Move To' collections list when moving an orphan item


juanmahidalgo:
## 🎉 Description

When trying to move a "Single Item" to a collection, we're showing all the collections for the address. We should filter the list down to the "allowed" ones (not published).
It might be better to add a query param to the endpoint and move the logic to the builder-server.

* [ ] 🐛 This is a bug report.
* [X] 📈 This is a feature request.

## 🔢 Steps To Reproduce Issue [ Good To Have ]

Please remember that with baby steps it's easier to reproduce the bug and it's much faster to fix it.

```
- Step 1: Go to `http://localhost:3000/collections`
- Step 2: Click the `Single Items` tab
- Step 3: When hovering a single item card, click the `...` button
- Step 4: Click `Move to collection` option
- Step 5: the list should only contain collections where it can be moved.
...
```

PR #2148, opened 2022-07-04
feat: Filter collections to only show allowed ones when moving a single item to a collection

cyaiox:
This PR allows filtering the available collections to show when moving an item to a collection.

Closes: #2143

### DIFF
--- a/src/components/CollectionDropdown/CollectionDropdown.tsx
+++ b/src/components/CollectionDropdown/CollectionDropdown.tsx
@@ -6,9 +6,9 @@ import './CollectionDropdown.css'
 
 export default class CollectionDropdown extends React.PureComponent<Props> {
   componentDidMount() {
-    const { address, isCollectionPublished, onFetchCollections } = this.props
+    const { address, onFetchCollections, fetchCollectionParams } = this.props
     if (address) {
-      onFetchCollections(address, {isPublished: isCollectionPublished})
+      onFetchCollections(address, fetchCollectionParams)
     }
   }
 
--- a/src/components/CollectionDropdown/CollectionDropdown.tsx
+++ b/src/components/CollectionDropdown/CollectionDropdown.tsx
@@ -6,9 +6,9 @@ import './CollectionDropdown.css'
 
 export default class CollectionDropdown extends React.PureComponent<Props> {
   componentDidMount() {
-    const { address, onFetchCollections } = this.props
+    const { address, isCollectionPublished, onFetchCollections } = this.props
     if (address) {
-      onFetchCollections(address)
+      onFetchCollections(address, {isPublished: isCollectionPublished})
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/components/CollectionDropdown/CollectionDropdown.tsx
```diff
@@ -6,9 +6,9 @@ import './CollectionDropdown.css'
 
 export default class CollectionDropdown extends React.PureComponent<Props> {
   componentDidMount() {
-    const { address, onFetchCollections } = this.props
+    const { address, isCollectionPublished, onFetchCollections } = this.props
```
juanmahidalgo, line 9: What do you think if we pass an entire object of the possible params for the fetch? it could be called `fetchCollectionParam`
```suggestion
    const { address, onFetchCollections, fetchCollectionParams } = this.props
```

src/components/CollectionDropdown/CollectionDropdown.tsx
```diff
@@ -6,9 +6,9 @@ import './CollectionDropdown.css'
...
     if (address) {
-      onFetchCollections(address)
+      onFetchCollections(address, {isPublished: isCollectionPublished})
```
juanmahidalgo, line 11: so here we could just pass it directly
```suggestion
      onFetchCollections(address, fetchCollectionParams)
```

src/components/CollectionDropdown/CollectionDropdown.types.ts
```diff
@@ -8,6 +8,7 @@ export type Props = {
   value?: Collection
   placeholder?: string
   filter?: (collection: Collection) => boolean
+  isCollectionPublished?: boolean
```
juanmahidalgo, line 11:
```suggestion
  fetchCollectionParams?: FetchCollectionsParams
```

src/components/Modals/MoveItemToCollectionModal/MoveItemToCollectionModal.tsx
```diff
@@ -29,7 +29,8 @@ export default class MoveItemToCollectionModal extends React.PureComponent<Props
           <CollectionDropdown
             value={collection}
             onChange={this.handleChangeCollection}
-            filter={collection => !isTPCollection(collection)}
+            filter={collection => !isTPCollection(collection) && !collection.isPublished}
+            isCollectionPublished={false}
```
juanmahidalgo, line 33: and here we could pass the object (so it gets more generic and more extensible)
```suggestion
            fetchCollectionParams ={{ isPublished: false }}
```

src/components/CollectionDropdown/CollectionDropdown.tsx
cyaiox: Nice proposal! doing it now 🚀

juanmahidalgo (approved):
LGTM! good job 👏

PR #2151, opened 2022-07-06
fix: add existing assignee when pushing a new curation

### LINKED ISSUE #2150: Collection assignee gets removed on push changes event
## 🎉 Description
When the creator pushes a change, the curation assignee gets removed.

* [X] 🐛 This is a bug report.

juanmahidalgo:
Closes #2150


src/modules/curations/collectionCuration/sagas.ts
```diff
@@ -70,9 +71,10 @@ export function* collectionCurationSaga(builder: BuilderAPI) {
 
   function* handlePushCurationRequest(action: PushCollectionCurationRequestAction) {
     const { collectionId } = action.payload
+    const curation: CollectionCuration | undefined = yield select(getCuration, collectionId)
 
     try {
-      yield call([builder, builder.pushCuration], collectionId)
+      yield call([builder, builder.pushCuration], collectionId, curation?.assignee)
```
LautaroPetaccio, line 77: Should we allow the server to set assignees when they're not curators? I think that this should be prevented and the server should do this automatically when a new collection curation is created.
Leaving the implementation thing aside, I have a couple of questions on how the curation process works:
- Do we want the same curator to be always assigned? We shouldn't re-assign curators that are not curators anymore and
- Is the responsibility of a curator that already curated a collection to always curate their changes?

LautaroPetaccio (changes requested):

src/modules/curations/collectionCuration/sagas.ts
juanmahidalgo: I believe that the curation should be the same through all the collection curation process. If the curation ask for a change and a new curation is created, I'd say to keep the same curation. This is what the PR does. The curators then can unassign themselves if they don't want to curate the update.

I didn't understand the first part of the question, asking of the server should set assignees when they're not curators, is it related to the PR or more of a general question?

Issue #2152, opened 2022-07-06
Moving an item out of a collection doesn't remove it from the single items lists


LautaroPetaccio:
## 🎉 Description

Moving an orphan item in the `Single items` section into a collection doesn't remove the item from the list of items when it should.

* [x] 🐛 This is a bug report.
* [ ] 📈 This is a feature request.

## 🔢 Steps To Reproduce Issue

1. Create a collection or be sure to have one created that is not published.
2. Create an orphaned item.
3. Go to the `Single items` list, click on item's options and move it into the collection and the item is still in the list.

PR #2175, opened 2022-07-06
fix: Fetch the items list again when moving a single item to a collection successfully

cyaiox:
This PR allows fetching the items list again when moving an orphan item to an unpublished collection successfully.

Closes: #2152

### DIFF
--- a/src/modules/item/sagas.ts
+++ b/src/modules/item/sagas.ts
@@ -451,6 +451,8 @@ export function* itemSaga(legacyBuilder: LegacyBuilderAPI, builder: BuilderClien
     yield put(saveItemRequest(newItem, {}))
     yield take(SAVE_ITEM_SUCCESS)
     yield put(closeModal('AddExistingItemModal'))
+    const address: string = yield select(getAddress)
+    yield put(fetchItemsRequest(address))
   }
 
   function* handleSetItemsTokenIdRequest(action: SetItemsTokenIdRequestAction) {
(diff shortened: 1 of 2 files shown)


src/modules/item/sagas.ts
```diff
@@ -451,6 +451,8 @@ export function* itemSaga(legacyBuilder: LegacyBuilderAPI, builder: BuilderClien
     yield put(saveItemRequest(newItem, {}))
     yield take(SAVE_ITEM_SUCCESS)
     yield put(closeModal('AddExistingItemModal'))
+    const address: string = yield select(getAddress)
```
LautaroPetaccio, line 454: Would you mind moving the select as up as possible in this method? This is not he case, but re-dispatching actions might result in undesired behavior if the data is not retrieved before putting other actions.
Also, would you mind adding a test that exercises this?

cyaiox: Thanks for your review. Sure, on it!

src/modules/item/sagas.ts
```diff
@@ -443,14 +443,15 @@ export function* itemSaga(legacyBuilder: LegacyBuilderAPI, builder: BuilderClien
...
     const newItem = { ...item }
+    const address: string = yield select(getAddress)
     if (collectionId === null) {
       delete newItem.collectionId
     } else {
       newItem.collectionId = collectionId
     }
     yield put(saveItemRequest(newItem, {}))
     yield take(SAVE_ITEM_SUCCESS)
-    yield put(closeModal('AddExistingItemModal'))
```
cyaiox, line 453: This `closeModal` is not required because the module: `modal.sagas` takes the action: `SAVE_ITEM_SUCCESS` and execute the method: `handleCloseAllModals`.
https://github.com/decentraland/builder/blob/master/src/modules/modal/sagas.ts#L13

LautaroPetaccio (approved):

PR #2176, opened 2022-07-07
feat: use emote entity type and schema

### LINKED ISSUE #2171: [Publish] Use EntityType.EMOTE
## 🎉 Description

Once the platform team introduces the new entity type for emotes, we need to stop deploying all items as wearable entities, use the right type (whether wearable or emote).

## 📝 Details

The `EntityType` is set here: https://github.com/decentraland/builder/blob/master/src/modules/item/export.ts#L224, we will need to change that to `EntityType.WEARABLE` or `EntityType.EMOTE` according to the `item.type`.

Also the schema for the emote metadata will be different to the wearable schema, we will need to make it fit making changes to this function: https://github.com/decentraland/builder/blob/master/src/modules/item/export.ts#L150-L190 otherwise the catalyst will reject the deployment.

A few of the changes i can recall is that the categories for the emotes are different to the wearable ones, the data and representations don't have hides/replaces/overrides, and there a `loop` boolean in the emote data.

cazala:
This PR makes the necessary changes to deploy emotes as `EntityType.EMOTE` instead of `EntityType.WEARABLE`. 

I created this PR as draft until the platform team enables a catalyst where we can test deploying these new entities.

Closes #2171


src/modules/item/utils.ts
```diff
@@ -461,6 +458,51 @@ export function areSynced(item: Item, entity: Entity) {
   return true
 }
 
+export function isEmoteSynced(item: Item, entity: Entity) {
+  if (item.type !== ItemType.EMOTE) {
+    throw new Error('Item must be EMOTE')
+  }
+
+  // check if metadata has the new schema from ADR 74
+  const isADR74 = 'emoteDataADR74' in entity.metadata
```
cazala, line 467: This is here because there are already some emotes deployed that don;t have the new schema from ADR 74. We will mark those as unsynced, so the owner (which is only Shibu so far) has to do a Push Chages + Approve, this will upload the new entities to the catalyst with the rht type and schema

juanmahidalgo (approved):
🚀

src/modules/item/utils.ts
```diff
@@ -461,6 +463,51 @@ export function areSynced(item: Item, entity: Entity) {
...
+    return false
+  }
+
+  // check if metadata is synced
+  const catalystItem = entity.metadata! as Emote
+
+  const hasMetadataChanged =
+    item.name !== catalystItem.name ||
+    item.description !== catalystItem.description ||
+    (item.data.category as string) !== catalystItem.emoteDataADR74.category ||
```
cazala, line 483: just a heads up, this is throwing for me, not being able to access `category` of `undefined`. This is because I have some legacy emotes deployed, and the `isEmotesFeatureFlagOn` is set to `false`. I think if that flag is false, then we should NOT look for the `emoteDataADR74`, otherwise owners of legacy emotes won't be able to use the builder until the flag is on. Instead we can look for the `emoteDataV0` prop while the flag is off, or just return false, which is not great, but is better than getting an error.

cc @juanmahidalgo 

Screen Shot 2022-08-10 at 16 04 55

PR #2180, opened 2022-07-11
feat: Add progress bar for pushing changes to a TP collection

### LINKED ISSUE #2139: TPW - Progress bar when pushing changes
## 🎉 Description

Add a progress bar to the action of pushing changes so it's clear how the process is going.

- Add a progress bar for the process.
- Prevent the user from leaving the modal.

* [ ] 🐛 This is a bug report.

juanmahidalgo:
Closes #2139

### DIFF
--- a/src/modules/thirdParty/sagas.ts
+++ b/src/modules/thirdParty/sagas.ts
@@ -174,7 +174,7 @@ export function* thirdPartySaga(builder: BuilderAPI, catalyst: CatalystClient) {
   }
 
   function* resetThirdPartyProgressAction() {
-    actionProgressChannel.put({
+    yield actionProgressChannel.put({
       progress: 0,
       tpAction: ThirdPartyAction.PUSH_CHANGES
     })
--- a/src/modules/thirdParty/sagas.ts
+++ b/src/modules/thirdParty/sagas.ts
@@ -60,7 +60,11 @@ import {
   deployBatchedThirdPartyItemsSuccess,
   deployBatchedThirdPartyItemsFailure,
   DEPLOY_BATCHED_THIRD_PARTY_ITEMS_REQUEST,
-  DeployBatchedThirdPartyItemsRequestAction
+  DeployBatchedThirdPartyItemsRequestAction,
+  PUSH_CHANGES_THIRD_PARTY_ITEMS_SUCCESS,
+  PUSH_CHANGES_THIRD_PARTY_ITEMS_FAILURE,
+  PUBLISH_AND_PUSH_CHANGES_THIRD_PARTY_ITEMS_FAILURE,
+  PUBLISH_AND_PUSH_CHANGES_THIRD_PARTY_ITEMS_SUCCESS
 } from './actions'
 import { getPublishItemsSignature } from './utils'
 import { ThirdParty } from './types'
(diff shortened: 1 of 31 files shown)


src/modules/ui/thirdparty/action.ts
```diff
@@ -1,8 +1,10 @@
...
+import { ThirdPartyAction } from './types'
 
-// TP Approval Flow
-export const APPROVAL_FLOW_UPDATE_PROGRESS = '[Update progress] Approval flow '
+// Update TP Action progress
+export const THIRD_PARTY_ACTION_UPDATE_PROGRESS = '[Update progress] Third Party Action'
 
-export const updateApprovalFlowProgress = (progress: number) => action(APPROVAL_FLOW_UPDATE_PROGRESS, { progress })
+export const updateThirdPartyActionProgress = (progress: number, tpAction: ThirdPartyAction) =>
+  action(THIRD_PARTY_ACTION_UPDATE_PROGRESS, { progress, action: tpAction })
```
LautaroPetaccio, line 8: Just to simplify it hahaha
```suggestion
export const updateThirdPartyActionProgress = (progress: number, action: ThirdPartyAction) =>
  action(THIRD_PARTY_ACTION_UPDATE_PROGRESS, { progress, action })
```

src/modules/thirdParty/sagas.ts
```diff
@@ -236,8 +241,8 @@ export function* thirdPartySaga(builder: BuilderAPI, catalyst: CatalystClient) {
     }
   }
 
-  function* handleUpdateApprovalFlowProgress(action: { progress: number }) {
-    yield put(updateApprovalFlowProgress(action.progress))
+  function* handleUpdateApprovalFlowProgress(action: { progress: number; tpAction: ThirdPartyAction }) {
+    yield put(updateThirdPartyActionProgress(action.progress, action.tpAction))
```
LautaroPetaccio, line 245: I think, as we're doing sagas integration tests, that we should be able to place a `put` in the tests to be sure that the update is being done, would you mind doing that?

src/modules/thirdParty/sagas.ts
```diff
@@ -167,6 +168,10 @@ export function* thirdPartySaga(builder: BuilderAPI, catalyst: CatalystClient) {
     const queue = new PQueue({ concurrency: MAX_CONCURRENT_REQUESTS })
     const promisesOfItemsBeingUpdated: (() => Promise<ItemCuration>)[] = items.map((item: Item) => {
       const curation = itemCurations.find(itemCuration => itemCuration.itemId === item.id)
+      actionProgressChannel.put({
+        progress: Math.round(((items.length - (queue.size + queue.pending)) / items.length) * 100),
+        tpAction: ThirdPartyAction.PUSH_CHANGES
+      })
```
LautaroPetaccio, line 174: If I'm not wrong, we're not clearing the progress when the modal closes, are we still showing the progress when the process starts until a request is completed?

LautaroPetaccio (changes requested):

src/modules/ui/thirdparty/action.ts
juanmahidalgo: I would love to, but it conflicts with the `action` fall from the next line 😅, that's why I had to call it with a different name
image

src/modules/thirdParty/sagas.ts
juanmahidalgo: Good point! Added!

PR #2181, opened 2022-07-11
fix: Incorrect collection curation fetch when the collection is not published

LautaroPetaccio:
This PR prevents the UI from requesting collection curations when the collection is not published.


nicosantangelo (approved):
LGTM

Issue #2179, opened 2022-07-08
Show an error when rarities cannot be fetched from the server


fzavalia:
![Screen Shot 2022-07-08 at 15.18.59.png](https://images.zenhubusercontent.com/5ea09af31c23030cb8e249b8/237fba90-906b-475b-a6bd-99695fb4fb5b)

PR #2183, opened 2022-07-11
feat: Show error when fetching rarities fails

fzavalia:
Closes https://github.com/decentraland/builder/issues/2179

Screen Shot 2022-07-11 at 15 20 12

### DIFF
--- a/src/components/Modals/PublishCollectionModalWithOracle/PublishCollectionModal.tsx
+++ b/src/components/Modals/PublishCollectionModalWithOracle/PublishCollectionModal.tsx
@@ -79,10 +79,15 @@ export default class PublishCollectionModal extends React.PureComponent<Props, S
       <>
         <ModalNavigation title={t('publish_collection_modal_with_oracle.title')} onClose={onClose} />
         <Modal.Content className="first-step">
-          {isFetchingItems || isFetchingRarities || !refRarity ? (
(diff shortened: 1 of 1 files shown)


src/components/Modals/PublishCollectionModalWithOracle/PublishCollectionModal.tsx
```diff
@@ -80,10 +80,15 @@ export default class PublishCollectionModal extends React.PureComponent<Props, S
       <>
         <ModalNavigation title={t('publish_collection_modal_with_oracle.title')} onClose={onClose} />
         <Modal.Content className="first-step">
-          {isFetchingItems || isFetchingRarities || !refRarity ? (
+          {isFetchingItems || isFetchingRarities ? (
             <div className="loader-wrapper">
               <Loader size="big" active={isFetchingItems || isFetchingRarities} />
             </div>
+          ) : !!itemError || !refRarity ? (
```
juanmahidalgo, line 87: Will `itemError` only hold rarities related errors? 
I'm asking because if there's an error but `refRarity` has value, it will show the rarities fetch error.

fzavalia: rarities is inside the item sagas so it includes both item and rarities errors

juanmahidalgo (approved):
LGTM!

PR #2185, opened 2022-07-12
feat: Show Error Toast when a TP action fails

### LINKED ISSUE #1830: Show error to the user if publish/push changes fails
Born from https://github.com/decentraland/builder/pull/1822. 

If either the publish or push changes action fail, we should show some kind of message/feedback to the user instead of keep the modal open.

juanmahidalgo:
Closes #1830

image

### DIFF
--- a/src/modules/thirdParty/sagas.ts
+++ b/src/modules/thirdParty/sagas.ts
@@ -1,14 +1,17 @@
 import PQueue from 'p-queue'
 import { channel } from 'redux-saga'
 import { takeLatest, takeEvery, call, put, select } from 'redux-saga/effects'
-import { Authenticator, AuthIdentity } from '@dcl/crypto'
-import { CatalystClient, DeploymentPreparationData } from 'dcl-catalyst-client'
 import { Contract, providers } from 'ethers'
+import { Authenticator, AuthIdentity } from '@dcl/crypto'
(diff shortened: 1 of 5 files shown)


src/modules/thirdParty/sagas.ts
```diff
@@ -153,9 +168,11 @@ export function* thirdPartySaga(builder: BuilderAPI, catalyst: CatalystClient) {
       )
 
       yield put(publishThirdPartyItemsSuccess(thirdParty.id, collectionId, newItems, newItemCurations))
-      yield put(closeModal('PublishThirdPartyCollectionModal'))
     } catch (error) {
+      yield showActionErrorToast()
```
LautaroPetaccio, line 172: Should we show some specific message about the error that occurred?

juanmahidalgo: Mhmm, I don't know what else could we show to be honest, I don't think it will be useful for the user to show the BE's error.
Do you have any ideas?

LautaroPetaccio (approved):

src/modules/thirdParty/sagas.ts
LautaroPetaccio: Let's go with this for the time being and if we need to get specific, we can change it.

Issue #2125, opened 2022-06-09
Disable the "USE AS ALIAS" button when the user does not have a profile


fzavalia:
When clicking this button when the user has not created a profile yet will:

![image.png](https://images.zenhubusercontent.com/5ea09af31c23030cb8e249b8/fb837606-a714-4f7f-8602-b0e8c4970e96)

Show a message without the name that is replacing.

![image.png](https://images.zenhubusercontent.com/5ea09af31c23030cb8e249b8/e493b199-9b19-4108-a987-aba861dcef9a)

Fail with a 404 when confirming because there is no profile to assign that name.

TASK:

Disable the button and add a tooltip or message indicating that you have to create you avatar in [play.decentraland.org](https://play.decentraland.org) first

VeitchM:
I started working in this issue

PR #2186, opened 2022-07-13
fix: Disable the "USE AS ALIAS" button when the user does not have a profile

cyaiox:
This PR disables the `Use as Alias button` when the user doesn't have a profile and shows a tooltip.

Screen Shot 2022-07-13 at 17 22 45

Closes: #2125


src/modules/translation/languages/en.json
```diff
@@ -1502,4 +1503,4 @@
       "subtitle": "Thank you for revieweing this collection"
     }
   }
-}
+}
```
juanmahidalgo, line 1506: mhmm, is your IDE removing the extra newline in here? I think this change shouldn't be here 😅

juanmahidalgo: same for other `.json` files!

src/components/ENSListPage/ENSListPage.tsx
```diff
@@ -89,8 +89,18 @@ export default class ENSListPage extends React.PureComponent<Props, State> {
...
+        <Button className="ui basic button" onClick={() => this.handleOpenModal(ens.name)} disabled={!hasProfileCreated}>
+          {t('ens_list_page.button.use_as_alias')}
+        </Button>
+      </div>
+    )
+  }
+
   renderEnsList() {
-    const { ensList } = this.props
+    const { ensList, hasProfileCreated } = this.props
```
juanmahidalgo, line 103:
```suggestion
    const { ensList, avatar } = this.props
    const hasProfileCreated = !!avatar
```
what do you think about this alternative? so we keep the prop as the selector

src/modules/translation/languages/en.json
cyaiox: Thanks for the review! Updated

juanmahidalgo (approved):
LGTM 🚀

Issue #2187, opened 2022-07-14
Materials count is not right for some wearables


juanmahidalgo:
## 🎉 Description

It was reported that for some wearables, the Material count shown when uploading the wearable is not right.

In the original file it was showing 10 correctly:
![image](https://user-images.githubusercontent.com/8763687/179000176-1340fe9e-e5c5-44f1-a981-052bbe80035a.png)

But the modified with (the one with issues and attached here) is showing 4 instead of 10:

![image](https://user-images.githubusercontent.com/8763687/179000383-491e6b6a-6c1b-4da6-892c-910dff47017c.png)

File:
[meterial_count_10.glb.zip](https://github.com/decentraland/builder/files/9112397/meterial_count_10.glb.zip)

* [X] 🐛 This is a bug report.
* [ ] 📈 This is a feature request.

PR #2188, opened 2022-07-14
fix: Update wearable materials count strategy

cyaiox:
This PR updates the wearable materials count strategy to have a more accurate number of materials.

Closes: #2187

### DIFF
--- a/src/lib/getModelData.ts
+++ b/src/lib/getModelData.ts
@@ -11,7 +11,8 @@ import {
   DirectionalLight,
   AmbientLight,
   RectAreaLight,
-  MeshStandardMaterial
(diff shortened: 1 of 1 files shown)


src/lib/getModelData.ts
```diff
@@ -90,7 +91,7 @@ export async function getModelData(url: string, options: Partial<Options> = {})
       if (node instanceof Mesh) {
         bodies++
         if (node.material) {
-          materials++
+          materials.add((node.material as Material).name)
```
cyaiox, line 94: The idea behind this set is to avoid counting multiple times the reused materials for different objects

PR #2191, opened 2022-07-18
feat: Adapt collection empty state to the new Emote one

### LINKED ISSUE #2156: [Creation] Update new item button to new designs
## 🎉 Description

Based on the fact that we can detect if a file is a wearable or an emote, let's go with the design that has only one button and the code will detect if the file is either an emote or wearable and launch the corresponding flow.

- Update the 'New Item' actual button to be 'Open in editor' since the new item action will be in the middle of the screen.

- The "Open in editor" button should be disabled if the collection is empty.

- Load ZIP and detect if it's an Emote

juanmahidalgo:
image

Closes #2156


nicosantangelo (commented):
Looks good but I don't understand how I am supposed to add items to a collection once I already added one?

image


(maybe this comes after this PR)

juanmahidalgo:
> Looks good but I don't understand how I am supposed to add items to a collection once I already added one?
> 
> image
> 
> (maybe this comes after this PR)

good point! it was a few mocks ahead of the current one but it should be updated here.
Empty state:
image

With > 1 items:
image

nicosantangelo (commented):
Two things!

Now the we have both buttons I think we need to accommodate for long collection names. This is with the max 32 chars:

image

image


Also! do we need to add something like this to Third Parties?

nicosantangelo:
I think we've might ran into another problem. I opened a collection named `Pipi` and got this:

image

and for a longer but not that long name:

juanmahidalgo:
> I think we've might ran into another problem. I opened a collection named `Pipi` and got this:
> 
> image
> 
> and for a longer but not that long name: image image

I couldn't reproduce it! I created the same collection but I see it just fine:

(at the smaller breaking before getting to the mobile version message)
image


bigger viewport:

image

juanmahidalgo:
With these latest changes:

## With short name
image

## With long name and published

image

## Long name and not publish

image

nicosantangelo:
Just a tiny thing left, the ON SALE toggle seems to overlap with the button:

image

image

nicosantangelo (approved):
LGTM !

PR #2194, opened 2022-07-19
feat: introduces a new FF for emotes and wraps the breaking change code

### LINKED ISSUE #2193: Create new Emotes flow FF and wrap the breaking change code with it
So we can continue to deliver code to `master`, let's wrap everything that is a breaking change with a FF.
Enable de FF in dev but not prod until it's done.

juanmahidalgo:
Closes #2193

### DIFF
--- a/src/modules/item/utils.ts
+++ b/src/modules/item/utils.ts
@@ -458,14 +458,14 @@ export function isWearableSynced(item: Item, entity: Entity) {
   return true
 }
 
-export function isEmoteSynced(item: Item, entity: Entity) {
+export function isEmoteSynced(item: Item, entity: Entity, isEmotesFeatureFlagOn: boolean) {
   if (item.type !== ItemType.EMOTE) {
     throw new Error('Item must be EMOTE')
   }
 
   // check if metadata has the new schema from ADR 74
   const isADR74 = 'emoteDataADR74' in entity.metadata
-  if (!isADR74) {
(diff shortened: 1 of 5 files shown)


src/modules/item/utils.ts
```diff
@@ -458,14 +458,14 @@ export function isWearableSynced(item: Item, entity: Entity) {
...
-export function isEmoteSynced(item: Item, entity: Entity) {
+export function isEmoteSynced(item: Item, entity: Entity, isEmotesFeatureFlagOn: boolean) {
   if (item.type !== ItemType.EMOTE) {
     throw new Error('Item must be EMOTE')
   }
 
   // check if metadata has the new schema from ADR 74
   const isADR74 = 'emoteDataADR74' in entity.metadata
-  if (!isADR74) {
+  if (!isADR74 && isEmotesFeatureFlagOn) {
```
nicosantangelo, line 468: Is there a way to get this boolean here without depending on an argument? I see it comes from another selector but looking at the code it seems it'd be nice to keep it locally to this file right?

nicosantangelo (approved):
LGTM but a comment I don't know if we can implement

src/modules/item/utils.ts
juanmahidalgo: Yeah, I thought the same thing but I didn't find another way since it comes from the state, I believe it must be passed as parameter :(

PR #2205, opened 2022-07-27
feat: use hashv1 for asset files

nicosantangelo:
depends on https://github.com/decentraland/builder-server/pull/573


src/modules/asset/utils.ts
```diff
@@ -52,8 +52,7 @@ export async function getContentsCID(asset: RawAsset): Promise<Record<string, st
 
   for (let filePath of Object.keys(contents)) {
     const file = await makeContentFile(filePath, contents[filePath])
-    const fileCID: string = await getCID([{ path: path.basename(file.path), content: file.content, size: file.size }], false)
```
LautaroPetaccio, line 55: Should we call the function getContentsHash?

src/modules/deployment/utils.ts
```diff
@@ -42,14 +41,14 @@ export const getDefaultGroundAsset = (): Asset => ({
   actions: []
 })
 
-export async function getCID(files: ContentServiceFile[], shareRoot: boolean): Promise<string> {
+export async function getCID(files: ContentServiceFile[]): Promise<string> {
```
LautaroPetaccio, line 44: Should we call this function getHash?

nicosantangelo: This function didn't actually change! I just used `hasV1` when needed

src/modules/asset/utils.ts
nicosantangelo: Good catch!

LautaroPetaccio (approved):

Issue #2160, opened 2022-07-06
[Creation] Set Emote Price


juanmahidalgo:
## 🎉 Description

After confirming the emote data, show the Set Price & Beneficiary modal (re-use the same modal that we have for wearables nowadays)

* [ ] 🐛 This is a bug report.
* [ ] 📈 This is a feature request.

![image.png](https://images.zenhubusercontent.com/61e56f8bfe61d4326945f317/d8b5a3de-28b5-4210-9c6c-4f23eb7d74fe)

PR #2206, opened 2022-07-27
feat: Set price modal

cyaiox:
This PR adds the set price and beneficiary step to the emote creation flow, allowing define the price and beneficiary or skip the step.

Closes: #2160

### DIFF
--- a/src/modules/item/sagas.ts
+++ b/src/modules/item/sagas.ts
@@ -105,7 +105,7 @@ import { getMethodData } from 'modules/wallet/utils'
 import { getCatalystContentUrl } from 'lib/api/peer'
 import { downloadZip } from 'lib/zip'
 import { calculateFinalSize, reHashOlderContents } from './export'
-import { Item, Rarity, CatalystItem, BodyShapeType, IMAGE_PATH, THUMBNAIL_PATH, WearableData } from './types'
+import { Item, Rarity, CatalystItem, BodyShapeType, IMAGE_PATH, THUMBNAIL_PATH, WearableData, ItemType } from './types'
 import { getData as getItemsById, getItems, getEntityByItemId, getCollectionItems, getItem, getPaginationData } from './selectors'
 import { ItemTooBigError } from './errors'
 import { buildZipContents, getMetadata, groupsOf, isValidText, generateCatalystImage, MAX_FILE_SIZE } from './utils'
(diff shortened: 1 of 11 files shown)


src/modules/item/sagas.ts
```diff
@@ -356,14 +356,23 @@ export function* itemSaga(legacyBuilder: LegacyBuilderAPI, builder: BuilderClien
...
+      yield put(closeModal('EditPriceAndBeneficiaryModal'))
+    } else if (openModals['CreateSingleItemModal'] && location.pathname === locations.collections() && item.type === ItemType.WEARABLE) {
       // Redirect to the newly created item detail
-      const { item } = action.payload
       yield put(push(locations.itemDetail(item.id)))
+    } else if (
+      openModals['CreateSingleItemModal'] &&
+      location.pathname === locations.collections() &&
+      item.type === ItemType.EMOTE &&
+      item.beneficiary
```
juanmahidalgo, line 371: Question: do we need the item to have a `beneficiary` to redirect the `itemEditor`?

src/modules/item/sagas.ts
```diff
@@ -454,6 +463,7 @@ export function* itemSaga(legacyBuilder: LegacyBuilderAPI, builder: BuilderClien
     }
     yield put(saveItemRequest(newItem, {}))
     yield take(SAVE_ITEM_SUCCESS)
+    yield put(closeModal('MoveItemToCollectionModal'))
```
juanmahidalgo, line 466: Question: is this `MoveItemToCollectionModal` related to this feature? Why do we need to close it? 🤔

src/modules/item/sagas.ts
juanmahidalgo: also, could we add a test for this new behavior? 😃

src/modules/item/sagas.ts
cyaiox: It's not related to this feature, but when removing the action: `SAVE_ITEM_SUCCESS` from the [modalSaga]((https://github.com/decentraland/builder/pull/2206/files#diff-36e062192786423af2f7cc0356f14a15104beff8969e0d53957d623e277f795eL5)), the modals that dispatch a `saveItemRequest` action are not closing automatically anymore

src/modules/item/sagas.ts
cyaiox: > Question: do we need the item to have a beneficiary to redirect the itemEditor?
This validation avoids redirecting when saving on the first step`(Name, Rarity, Category, Play Mode)` and redirects when the User reaches the last step: `Set a price` where I get the beneficiary(or could be price) to know is the last step.

Could another option be to pass the step in the url and validate with the `pathname`?

> also, could we add a test for this new behavior? 😃
Sure!

src/modules/item/sagas.ts
```diff
@@ -356,14 +356,18 @@ export function* itemSaga(legacyBuilder: LegacyBuilderAPI, builder: BuilderClien
...
     if (openModals['EditItemURNModal']) {
       yield put(closeModal('EditItemURNModal'))
-    } else if (openModals['CreateSingleItemModal'] && location.pathname === locations.collections()) {
+    } else if (openModals['EditPriceAndBeneficiaryModal']) {
+      yield put(closeModal('EditPriceAndBeneficiaryModal'))
+    } else if (openModals['CreateSingleItemModal'] && location.pathname === locations.collections() && item.type === ItemType.WEARABLE) {
       // Redirect to the newly created item detail
-      const { item } = action.payload
       yield put(push(locations.itemDetail(item.id)))
+    } else if (openModals['CreateSingleItemModal'] && location.pathname === locations.collections() && item.type === ItemType.EMOTE) {
```
juanmahidalgo, line 367: Question: Should we add the emote's feature flag here as well?

cyaiox: Good catch. We should add the emote's FF here.

Doing...

PR #2229, opened 2022-08-10
feat: support wearable.json

nicosantangelo:
Ditch `asset.json` jsons in favour of the new (alpha) `wearable.json` file.

Examples:
[portable.zip](https://github.com/decentraland/builder/files/9304295/portable.zip)
[banana.zip](https://github.com/decentraland/builder/files/9304298/banana.zip)
[hat.zip](https://github.com/decentraland/builder/files/9304301/hat.zip)


src/components/Modals/CreateSingleItemModal/ImportStep/ImportStep.tsx
```diff
@@ -44,58 +43,34 @@ export default class ImportStep extends React.PureComponent<Props, State> {
...
-
-    // asset.json contains data to populate parts of the state
-    const assetJsonPath = fileNames.find(path => basename(path) === ASSET_MANIFEST)
     let assetJson: ItemAssetJson | undefined
-
-    if (assetJsonPath) {
-      const assetRaw = zip.file(assetJsonPath)
-      const content = await assetRaw.async('text')
-      assetJson = JSON.parse(content)
+    let modelPath: string | undefined
```
LautaroPetaccio, line 51: I think we shouldn't keep the concept of `assetJson` and its type, as it refers to the older mechanism of importing assets.

src/components/Modals/CreateSingleItemModal/ImportStep/ImportStep.tsx
```diff
@@ -44,58 +43,34 @@ export default class ImportStep extends React.PureComponent<Props, State> {
...
-      contents[file.name] = file.blob
-      return contents
-    }, {})
-
     if (!modelPath) {
       throw new MissingModelFileError()
     }
 
-    const result = await this.processModel(modelPath, contents)
+    const result = await this.processModel(modelPath, content)
```
LautaroPetaccio, line 71: The result of `processModel` will ignore one of the representations, I think we should change the code so we can return from this function all the data we can to use it later when building the item.

LautaroPetaccio (approved):

src/modules/item/utils.ts
```diff
@@ -296,7 +300,7 @@ export function getThumbnailURL(item: Item) {
 }
 
 export function getRarities() {
-  return Object.values(ItemRarity)
+  return ItemRarity.schema.enum as ItemRarity[]
```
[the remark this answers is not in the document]
nicosantangelo: I think enum is already the `Object.values` version of the enum: https://github.com/decentraland/common-schemas/blob/main/src/dapps/rarity.ts#L20

src/components/Modals/CreateSingleItemModal/ImportStep/ImportStep.tsx
nicosantangelo: Maybe we can expand ModelData (the return type) to return all the properties we currently have on assetJSON which are:

```ts
name
description
rarity
category
bodyShape
```

another option is to just rename the prop?

I'm hesitant to rewrite everything as this also is being checked later on in `getAssetJsonProps`

What did you have in mind?

src/components/Modals/CreateSingleItemModal/ImportStep/ImportStep.tsx
nicosantangelo: Sorry a bit confused here. I think process model does not deal with representations?

src/modules/item/utils.ts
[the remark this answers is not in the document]
LautaroPetaccio: You're right, my mistake!

PR #2233, opened 2022-08-12
feat: Show Emote Metrics in the ItemDetailPage

juanmahidalgo:


src/components/CollectionDetailPage/CollectionDetailPage.tsx
```diff
@@ -93,8 +93,11 @@ export default class CollectionDetailPage extends React.PureComponent<Props, Sta
     const canMint = canMintCollectionItems(collection, wallet.address)
     const isOnSale = isCollectionOnSale(collection, wallet)
     const isLocked = isCollectionLocked(collection)
-    const filteredItems = items.filter(item => item.type === tab)
-    const showShowTabs = items.some(item => item.type === ItemType.WEARABLE) && items.some(item => item.type === ItemType.EMOTE)
+    const hasEmotes = items.some(item => item.type === ItemType.EMOTE)
```
juanmahidalgo, line 96: not related to the Emote Metrics, I'm using this PR to include this fix as well :) 
We should show the tabs in the collection detail page only if the collection has both type of Items, if not, show only the needed table.

PR #2234, opened 2022-08-14
fix: Use lootie-react instead of react-lootie

LautaroPetaccio:
This PR changes lootie-react for react-lootie. The old library supported up to React 16 and needed to be replaced.


nicosantangelo (approved):
🙏

PR #2239, opened 2022-08-17
fix: Fix the missing prop `bodyshape` when adding an item representation

cyaiox:
This PR fixes the missing prop `bodyShape` when adding an item representation.


juanmahidalgo (approved):
🚀 🚀 🚀

PR #2283, opened 2022-09-01
fix: ENS not working when there's no LAND

LautaroPetaccio:
This PR fixes the issue of the names not showing up when the user has no lands.
The issue was caused due to having an empty array of coordinates being sent to the endpoint, making the endpoint fail thus preventing the rest of the operation to succeed.


cyaiox (approved):

fzavalia:
🚀

PR #2293, opened 2022-09-07
chore: Add pre-commit/pre-push hooks

cyaiox:
This PR adds a pre-commit/pre-push hooks to apply linting, prettify the code and run tests.


.eslintrc.json
```diff
@@ -0,0 +1,33 @@
...
+  "ignorePatterns": ["src/contracts/*.ts", "src/ecsScene/scene.js", "package.json", "package-lock.json"],
+  "plugins": ["@typescript-eslint"],
+  "extends": ["eslint:recommended", "plugin:@typescript-eslint/recommended", "prettier"],
+  "env": {
+    "browser": true,
+    "es6": true
+  },
+  "rules": {
+    "@typescript-eslint/no-non-null-assertion": "off",
+    "@typescript-eslint/no-unused-vars": "off",
```
LautaroPetaccio, line 13: I would prefer this to be enabled to disabled the no-unused-vars option in the compiler that is usually pretty annoying. What do you think about it?

.eslintrc.json
```diff
@@ -0,0 +1,33 @@
...
+    "@typescript-eslint/ban-types": "off",
```
LautaroPetaccio, line 14: Does enabling this implies a lot of changes?

.eslintrc.json
```diff
@@ -0,0 +1,33 @@
...
+    "@typescript-eslint/quotemark": "off",
+    "@typescript-eslint/no-commented-out-code": "off",
```
LautaroPetaccio, line 16: I would set this as on, we shouldn't have commented code, right?

.eslintrc.json
```diff
@@ -0,0 +1,33 @@
...
+    "@typescript-eslint/no-unnecessary-type-assertion": "off"
```
LautaroPetaccio, line 17: Are there a lot of places that need this rule to be off?

.husky/pre-push
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+echo "Running tests"
+CI=true npm run test -- --all --silent
```
LautaroPetaccio, line 5: I would not force tests to be ran as tests can be pretty time consuming.

.eslintrc.json
LautaroPetaccio, line 13: I would set this as `on` and remove the unused var restriction on the compiler so we can develop without restrictions.

LautaroPetaccio (commented):
Great contribution, I love it 🚀 
I think we need to fix all the codebase at once before merging it so we don't end up with PRs that have a ton of unwanted changes.

LautaroPetaccio (approved):

PR #2295, opened 2022-09-08
fix: Stop emote animation when pressing button stop

### LINKED ISSUE #2289: [BUG][Editor][CenterPanel] Update button Play Emote stop method
## 🎉 Description

Stopping an emote should stop the animation and go back to the original Idle position. Today it seems it works like a Pause because if you stop and reproduce the same emote, it continues the animation (instead of starting over)

* [x] 🐛 This is a bug report.

cyaiox:
This PR sets the avatar to IDLE when pressing the stop button.

Closes #2289

### DIFF
--- a/src/components/ItemEditorPage/CenterPanel/CenterPanel.tsx
+++ b/src/components/ItemEditorPage/CenterPanel/CenterPanel.tsx
@@ -211,7 +211,6 @@ export default class CenterPanel extends React.PureComponent<Props, State> {
     return (
       <div className={`CenterPanel ${isImportFilesModalOpen ? 'import-files-modal-is-open' : ''}`}>
         <WearablePreview
-          baseUrl="http://localhost:3001"
           id="wearable-editor"
           profile="default"
           bodyShape={bodyShape}
--- a/src/components/ItemEditorPage/CenterPanel/CenterPanel.tsx
+++ b/src/components/ItemEditorPage/CenterPanel/CenterPanel.tsx
@@ -111,20 +111,25 @@ export default class CenterPanel extends React.PureComponent<Props, State> {
   }
 
   handleWearablePreviewLoad = () => {
-    const { wearableController, onSetWearablePreviewController } = this.props
+    // const { wearableController, onSetWearablePreviewController } = this.props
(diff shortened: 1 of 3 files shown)


src/components/ItemEditorPage/CenterPanel/CenterPanel.tsx
```diff
@@ -47,6 +47,11 @@ export default class CenterPanel extends React.PureComponent<Props, State> {
     this.setState({ isLoading: true })
   }
 
+  componentWillUnmount() {
+    const { onSetWearablePreviewController } = this.props
+    onSetWearablePreviewController(null)
```
LautaroPetaccio, line 52: Setting the wearable preview controller as null will work, but it won't close the channel because the unsubscribe function won't be ever executed.
It might be harmless but the correct way to do it is to emit an END message in the controller when the wearable preview component gets unmounted.

src/modules/editor/sagas.ts
```diff
@@ -499,24 +497,27 @@ function createWearablePreviewChannel(controller: IPreviewController) {
...
-        case PreviewEmoteEventType.ANIMATION_PLAY:
-          yield put(setEmotePlaying(true))
-          break
-        case PreviewEmoteEventType.ANIMATION_PAUSE:
-          yield put(setEmotePlaying(false))
-          break
-        case PreviewEmoteEventType.ANIMATION_END:
-          yield put(setEmotePlaying(false))
-          break
+  if (controller) {
```
LautaroPetaccio, line 501: This is just a question about the line 499, can't the controller be get from the action itself instead of the redux-state?

cyaiox: yep, updating!

src/components/ItemEditorPage/CenterPanel/CenterPanel.tsx
cyaiox: thanks for the feedback! working on it

LautaroPetaccio (approved):
Approving this as it's not extremely required for this bug to be fixed to remove the channel, but it's something we might want to look in the future.

PR #2312, opened 2022-09-15
fix: looped emotes

cazala:
The condition was wrong, because the `isEmotePlayModeFeatureFlagOn` is used to DISABLE the `EmotePlayMode.LOOP` from the dropdown. [So when the flag is ON, the feature is OFF](https://github.com/decentraland/builder/blob/master/src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx#L779). 

What I did is I removed the ternary when setting the value of the `loop` prop to always be what is in `state`, since the we already have the check of the flag to disable the dropdown.

### DIFF
--- a/src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
+++ b/src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
@@ -269,7 +269,7 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
   }
 
   modifyItem = async (pristineItem: Item, sortedContents: SortedContent, representations: WearableRepresentation[]) => {
-    const { isEmotePlayModeFeatureFlagOn, onSave } = this.props
+    const { onSave } = this.props
     const { name, bodyShape, type, metrics, category, playMode } = this.state as StateData
 
     let data: WearableData | EmoteDataADR74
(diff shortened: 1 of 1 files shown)


cyaiox:
good catch @cazala! 

is it possible to update here: https://github.com/decentraland/builder/pull/2312/files#diff-14a508dde6f9bf5af42a88ac3c6859b51313837d30291f7c8cee96199399e024R779 the condition to:

`const isDisabled = !isEmotePlayModeFeatureFlagOn ? value === EmotePlayMode.LOOP : false`

cazala:
Hey @cyaiox I dunno if i followed what you said. If you mean changing the feature flag to work the other way around (so the flag ON means the feature is ON) it's kind of late since that's in production now.. so we would need to flip the value and deploy at the same time. And since we are about to launch I wouldn't do any deployments until after the announcement.

Regarding the second point, I think that link doesnt take me to any particular line, and i don't know what you meant by `and use the value from the state here`. I changed it so it does use the value from state always.

cyaiox:
Hey @cazala! My bad, I copy/paste a wrong url... here is the line: https://github.com/decentraland/builder/blob/master/src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx#L287

cyaiox (approved):

PR #2313, opened 2022-09-15
fix: create emote from editor

cazala:
When inside the item editor, but without opening any collection or item, the url is `/item-editor`, and there's no `paginationData`, but there is a `+` button to create an item. If you create an item in this particular page, the success saga at the end fails because `paginationSaga` is undefined. I added a check to make sure it exists.

### DIFF
--- a/src/modules/item/sagas.ts
+++ b/src/modules/item/sagas.ts
@@ -93,7 +93,7 @@ import { updateProgressSaveMultipleItems } from 'modules/ui/createMultipleItems/
 import { isLocked } from 'modules/collection/utils'
 import { locations } from 'routing/locations'
 import { BuilderAPI as LegacyBuilderAPI, FetchCollectionsParams } from 'lib/api/builder'
-import { DEFAULT_PAGE, PaginatedResource, PaginationStats } from 'lib/api/pagination'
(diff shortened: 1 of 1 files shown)


src/modules/item/sagas.ts
```diff
@@ -381,10 +381,14 @@ export function* itemSaga(legacyBuilder: LegacyBuilderAPI, builder: BuilderClien
...
           if (collectionId) {
             const paginationData: ItemPaginationData = yield select(getPaginationData, collectionId)
-            yield put(fetchCollectionItemsRequest(collectionId, { page: paginationData.currentPage, limit: paginationData.limit }))
+            if (paginationData) {
+              yield put(fetchCollectionItemsRequest(collectionId, { page: paginationData.currentPage, limit: paginationData.limit }))
+            }
           } else {
             const paginationData: ItemPaginationData = yield select(getPaginationData, address)
-            yield put(fetchItemsRequest(address, { page: paginationData.currentPage, limit: paginationData.limit }))
+            if (paginationData) {
```
cyaiox, line 389: if there's no pagination, could we pass the DEFAULT_PAGE and the DEFAULT_PAGE_SIZE?

yield put(fetchCollectionItemsRequest(collectionId, { page: paginationData.currentPage || DEFAULT_PAGE, limit: paginationData.limit || DEFAULT_PAGE_SIZE}))

cyaiox (approved):

PR #2314, opened 2022-09-15
feat: support blobs that are ADR74 emotes

cazala:
~This PR depends on this one to be released before merging: https://github.com/decentraland/ui/pull/199~ Ready ✅


src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
```diff
@@ -603,7 +603,7 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
   renderWearablePreview = () => {
     const { type, contents } = this.state
     const isEmote = type === ItemType.EMOTE
-    const blob = contents ? toWearableWithBlobs({ contents, isEmote }) : undefined
+    const blob = contents ? toEmoteWithBlobs({ contents }) : undefined
```
cyaiox, line 606: the method `renderWearablePreview` also computes the thumbnail for wearables. 
the blob should be something like this:
`const blob = contents ? (isEmote ? toEmoteWithBlobs({ contents }) : toWearableWithBlobs({ contents })) : undefined`

cazala: good catch!

cyaiox (approved):
🚀

cazala:
@cyaiox just a heads up I added one more commit to remove an unnecesary part of the code that checked for a main image file on an emote file, since emotes can't be images.

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
```diff
@@ -945,7 +945,7 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
     return (
       <EditThumbnailStep
         isLoading={!!isLoading}
-        blob={contents ? toWearableWithBlobs({ contents, isEmote: true }) : undefined}
+        blob={contents ? toEmoteWithBlobs({ contents }) : undefined}
```
juanmahidalgo, line 948: This should be 
```suggestion
        blob={blob}
```
right?

src/components/Modals/EditThumbnailModal/EditThumbnailModal.tsx
```diff
@@ -61,7 +61,7 @@ export default class EditThumbnailModal extends React.PureComponent<Props, State
 
     let wearablePreviewProp
     if (file) {
-      wearablePreviewProp = { blob: toWearableWithBlobs({ file, isEmote: true }) }
+      wearablePreviewProp = { blob: toEmoteWithBlobs({ file }) }
```
juanmahidalgo, line 64: we should add some logic here to know if it should use `toEmoteWithBlobs` or `toWearableWithBlobs`, right?

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
cazala: There’s no ‘blob’ variable in that scope 👀

src/components/Modals/EditThumbnailModal/EditThumbnailModal.tsx
cazala: The EditThumbnailModal is just for emotes I think (it is the one with the slider to pick the frame), so no support for WearableWithBlobs

src/components/Modals/CreateSingleItemModal/EditThumbnailStep/EditThumbnailStep.types.ts
```diff
@@ -18,7 +18,7 @@ export enum ControlOptionAction {
 export type Props = {
   title: string
   isLoading: boolean
-  blob?: WearableWithBlobs
+  blob?: EmoteWithBlobs
```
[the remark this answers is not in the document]
cazala: Same answer as above

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
juanmahidalgo: sorry, GitHub trolled me 😛 
image
Now that I'm seeing it correctly, we only need the `toEmoteWithBlobs` here cause it's only for emotes 👍

src/components/Modals/EditThumbnailModal/EditThumbnailModal.tsx
juanmahidalgo: Yes, correct! perfect then 👍

juanmahidalgo (approved):

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
cyaiox: hey @juanmahidalgo!  the method renderWearablePreview is used in the import file step to generate the initial thumbnail for emotes and wearables.

PR #2322, opened 2022-09-21
feat: add play-mode to Emotes metadata

juanmahidalgo:


cyaiox (approved):

fzavalia (approved):

fzavalia:
Assuming this urn format is correct `${version}:${type}:${name}:${description}:${category}:${bodyShapeTypes}:${playMode}`

juanmahidalgo:
> Approved assuming this urn format is correct `${version}:${type}:${name}:${description}:${category}:${bodyShapeTypes}:${playMode}`

I believe it doesn't translate into an urn directly. It's parsed later on by the collections graph to create the entities. The urn does not support `playMode` at the end, do you believe that's an issue @fzavalia ?

fzavalia:
> > Approved assuming this urn format is correct `${version}:${type}:${name}:${description}:${category}:${bodyShapeTypes}:${playMode}`
> 
> I believe it doesn't translate into an urn directly. It's parsed later on by the collections graph to create the entities. The urn does not support `playMode` at the end, do you believe that's an issue @fzavalia ?

Seems ok 😄

cazala:
as discussed offline, i'd suggest not incorporating the concept of `playMode` into the system other than for the UI/UX, and keep the value as `loop: boolean` to match the ADR74 schema. So we could just store `0` or `1` in this slot in order to tell the subgraph the boolean value.

cazala (approved):

PR #2361, opened 2022-10-07
fix: thegraph 1000 limit for land

### LINKED ISSUE #2360: TheGraph 1000 limit making some users not see all their land
It has been reported by users who own +1000 lands that they can't see all their lands in the builder

cazala:
Fixes #2360 

I impersonated this address `0xfd649cd5a87fe803970cdfedf50efb7904dccc85` who owns 2815 LANDs and I was able to reproduce the issue, only 1000 land were loaded. This was the map:

Screen Shot 2022-10-07 at 17 05 15
Screen Shot 2022-10-07 at 17 05 25

What I did was adding a check to see if the results have hit the max amount of 1000, and if so, keep fetch the next 1000 using recursion, until the results of a single query are less than the maximum amount. Now I can see the 2815 LANDs of that user:

Screen Shot 2022-10-07 at 16 50 11

Screen Shot 2022-10-07 at 16 50 23

I also took the chance to add some missing traslations in the tooltip of the map page.


src/lib/api/manager.ts
```diff
@@ -121,12 +123,26 @@ const fromEstate = (estate: EstateFields, role: RoleType) => {
   return result
 }
 
+function isMax(result: LandQueryResult) {
+  // true if any result length matches the MAX_RESULTS constant
+  return (
+    result.tenantParcels.length >= MAX_RESULTS ||
```
fzavalia, line 129: Maybe refactor to something like `Object.values(result).some(x => x.length >= MAX_RESULTS)`

fzavalia: Afaik every property in the result is an array

cazala: done ✅

PR #2371, opened 2022-10-13
fix: Update event banner component

cyaiox:
This PR updates the component banner from the advertiser component to a plane div to avoid being blocked by ad blockers. 

Banner:
image


juanmahidalgo (approved):
LGTM

PR #2380, opened 2022-10-20
fix: eslint require-await rule

cyaiox:
This PR fixes the eslint require-await rule

### DIFF
--- a/src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
+++ b/src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
@@ -404,14 +404,16 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
     this.setState({ item: item, category: item.data.category, rarity: item.rarity })
   }
 
-  handleCategoryChange = async (_event: React.SyntheticEvent<HTMLElement, Event>, { value }: DropdownProps) => {
+  handleCategoryChange = (_event: React.SyntheticEvent<HTMLElement, Event>, { value }: DropdownProps) => {
     const category = value as WearableCategory
     if (this.state.category !== category) {
       this.setState({ category })
       if (this.state.type === ItemType.WEARABLE) {
-        await this.updateThumbnailByCategory(category)
+        return this.updateThumbnailByCategory(category)
       }
     }
+
+    return
   }
 
   handleRarityChange = (_event: React.SyntheticEvent<HTMLElement, Event>, { value }: DropdownProps) => {
@@ -586,8 +588,7 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
     const { weareblePreviewUpdated, type, model } = this.state
     // if model is an image, the wearable preview won't be needed
     if (model && isImageFile(model)) {
-      await this.getMetricsAndScreenshot()
-      return
+      return this.getMetricsAndScreenshot()
     }
 
     const controller = WearablePreview.createController('thumbnail-picker')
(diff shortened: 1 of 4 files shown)


src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
```diff
@@ -586,15 +586,15 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
     const { weareblePreviewUpdated, type, model } = this.state
     // if model is an image, the wearable preview won't be needed
     if (model && isImageFile(model)) {
-      this.getMetricsAndScreenshot()
+      await this.getMetricsAndScreenshot()
```
LautaroPetaccio, line 589: Returning the promise works as well:
```suggestion
      return this.getMetricsAndScreenshot()
```

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
```diff
@@ -404,13 +404,13 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
     this.setState({ item: item, category: item.data.category, rarity: item.rarity })
   }
 
-  handleCategoryChange = (_event: React.SyntheticEvent<HTMLElement, Event>, { value }: DropdownProps) => {
+  handleCategoryChange = async (_event: React.SyntheticEvent<HTMLElement, Event>, { value }: DropdownProps) => {
```
LautaroPetaccio, line 407: This function didn't have an async keyword on it. What was the reason behind its change?
There's also the possibility of removing the `async` keyword from it and returning `this.updateThumbnailByCategory(category)`

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
```diff
@@ -586,15 +586,15 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
...
+      await this.getMetricsAndScreenshot()
       return
     }
 
     const controller = WearablePreview.createController('thumbnail-picker')
     this.setState({ previewController: controller })
 
     if (weareblePreviewUpdated) {
-      this.getMetricsAndScreenshot()
+      await this.getMetricsAndScreenshot()
```
LautaroPetaccio, line 597: Was this await forgotten or was it placed on purpose to prevent blocking the UI?

LautaroPetaccio (changes requested):

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
cyaiox: the reason to add the async keyword was that the method `this.updateThumbnailByCategory`is async and broke an eslint rule type but returning it works without problems.

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
cyaiox: I think it was initially forgotten because this step occurs when loading the model for the first time

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
```diff
@@ -404,14 +404,16 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
...
     const category = value as WearableCategory
     if (this.state.category !== category) {
       this.setState({ category })
       if (this.state.type === ItemType.WEARABLE) {
-        await this.updateThumbnailByCategory(category)
+        return this.updateThumbnailByCategory(category)
       }
     }
+
+    return
```
LautaroPetaccio, line 416: Would you mind removing this return as it has no effect on the function?

LautaroPetaccio (changes requested):

PR #2384, opened 2022-10-24
fix: Update calculation of the max item size

### LINKED ISSUE #2382: [BUG][Creation] Error creating wearables by max size exceeds
## 🎉 Description

Error calculating the model and thumbnail size when creating a wearable.

* [x] 🐛 This is a bug report.

cyaiox:
This PR updates the max item size calculation, computing the model's contents and thumbnails with different limits.

Closes #2382

### DIFF
--- a/src/modules/item/export.ts
+++ b/src/modules/item/export.ts
@@ -70,7 +70,11 @@ function getUniqueFiles(hashes: Record<string, string>, blobs: Record<string, Bl
  * @param item - An item that contains the old and the new hashed content.
  * @param newContents - The new content that is going to be added to the item.
  */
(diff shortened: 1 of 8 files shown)


src/modules/item/export.ts
```diff
@@ -96,6 +100,10 @@ export async function calculateFinalSize(item: Item, newContents: Record<string,
   return imageSize + calculateFilesSize(uniqueFiles)
 }
 
+export function calculateThumbnailFinalSize(thumbnail: Blob): number {
```
LautaroPetaccio, line 103: As the calculating the thumbnail size implies calculating a file size, I would rename this function to `calculateFileZie` to compute the size of a single file or use it as is. What do you think?

cazala (approved):

src/modules/item/export.ts
cazala: +1

PR #2392, opened 2022-10-28
feat: Remove emote play mode ff

cyaiox:
This PR removes the emote-play-mode FF and other unused FF types related to the emote flow


juanmahidalgo (approved):
LGTM!

PR #2398, opened 2022-11-02
fix: Wrap tags in the item detail page

### LINKED ISSUE #2362: Having many tags messes up the UI
The tags don't wrap down, they just continue to the right but the page cannot be scrolled.

kevinszuchet:
Closes #2362

Screen Shot 2022-11-02 at 13 15 10


juanmahidalgo (approved):
👏👏👏

PR #2402, opened 2022-11-03
feat: Add filter by type in the curation page

### LINKED ISSUE #1724: Add type filters to Curation page
The current filters look like this:

We need to add a TYPE filter, and it should work like this:
- The TYPE filter should be `ALL`, `Standard`, `Third Party`
- If no TYPE filter is `ALL` or `Third Party`, both of the current selectors should be disabled
- If the TYPE is `Standard`, re-enable the other filters

If the TYPE changes, reset the other filters too

Depends on https://github.com/decentraland/builder-server/pull/382 for creating TP CollectionCurations
Might be related to #1723

This needs the server to support the filter as well. Create the task if necessary.

kevinszuchet:
Closes #1724 

image

### DIFF
--- a/src/components/CurationPage/CurationPage.css
+++ b/src/components/CurationPage/CurationPage.css
@@ -8,6 +8,12 @@
 
 .CurationPage .filters .text-filter-row {
   width: 100%;
+  box-sizing: border-box;
+  height: 44px;
+  background: #242129;
+  border-radius: 10px;
+  margin-bottom: 24px;
+  padding: 14px 12px;
 }
 
 .CurationPage .pagination {
--- a/src/components/CurationPage/CurationPage.css
+++ b/src/components/CurationPage/CurationPage.css
@@ -10,7 +10,7 @@
   width: 100%;
   box-sizing: border-box;
   height: 44px;
-  background: #242129;
+  background: var(--shadows);
   border-radius: 10px;
   margin-bottom: 24px;
   padding: 14px 12px;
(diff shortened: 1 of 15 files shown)


src/components/CurationPage/CurationPage.css
```diff
@@ -8,6 +8,12 @@
 
 .CurationPage .filters .text-filter-row {
   width: 100%;
+  box-sizing: border-box;
+  height: 44px;
+  background: #242129;
```
juanmahidalgo, line 13: We have a CSS variable that you can use for this color!

image

src/components/CurationPage/CurationPage.tsx
```diff
@@ -142,6 +180,7 @@ export default class CurationPage extends React.PureComponent<Props, State> {
           { value: CurationFilterOptions.APPROVED, text: t('curation_page.filter.approved') },
           { value: CurationFilterOptions.REJECTED, text: t('curation_page.filter.rejected') }
         ]}
+        disabled={filterByType !== CollectionFilterOptions.STANDARD}
```
juanmahidalgo, line 183: Can we remove the _disabling_ logic? I think we support all filters for all types of collections now!

kevinszuchet: Yes, we can. We should also remove the `reset` of the filters after changing the type, should we?

juanmahidalgo (approved):
Looks great 🙌

mrbaguvix (approved):

juanmahidalgo (approved):
LGTM

PR #2416, opened 2022-11-10
chore: eslint disable next line

cyaiox:


src/modules/translation/languages/index.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
 const en = require('./en.json')
 const es = require('./es.json')
 const zh = require('./zh.json')
```
LautaroPetaccio, line 4: These should work with imports as well, would you mind trying?

cyaiox: @LautaroPetaccio to use `imports` here, it's required to update the [translation interface](https://github.com/decentraland/decentraland-dapps/blob/master/src/modules/translation/types.ts#L5). I'll add a TODO comment to migrate to `imports`

PR #2421, opened 2022-11-10
fix: Atlas not rendering tenants

### LINKED ISSUE #2404: Cannot select rented Land to deploy a scene
With address `0x2f89eC84e0413950d9ADF8e56dd56c2B2f5066cb` I have rented from `0x24e5F44999c151f08609F8e27b2238c773C4D020` a Land with token id `34368519059014784809800835350608589357057`

I have rented it with my same address as the operator.

When I go to the builder, I see that I am both the tenant and the operator.

But when I want to deploy a scene, I am not given the option to select the rented Land.

LautaroPetaccio:
This PR fixes the layering of the Atlas component so when being a tenant, the rented LAND is shown as orange.
It also fixes some missing keys that some maps were missings.
Closes #2404

Screen Shot 2022-11-10 at 18 25 37
Screen Shot 2022-11-10 at 18 23 46
Screen Shot 2022-11-10 at 18 23 41

### DIFF
--- a/src/components/Modals/DeployModal/DeployToLand/LandAtlas/LandAtlas.tsx
+++ b/src/components/Modals/DeployModal/DeployToLand/LandAtlas/LandAtlas.tsx
@@ -112,7 +112,7 @@ export default class LandAtlas extends React.PureComponent<Props, State> {
   handleSelectPlacement = () => {
     const { placement } = this.state
     const { onConfirmPlacement } = this.props
-    const overlappedDeployment = this.getOverlappedDeployemnt()
+    const overlappedDeployment = this.getOverlappedDeployment()
     if (placement) {
       overlappedDeployment ? onConfirmPlacement(placement, overlappedDeployment.id) : onConfirmPlacement(placement)
     }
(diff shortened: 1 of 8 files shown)


src/components/Modals/DeployModal/DeployToLand/LandAtlas/LandAtlas.tsx
```diff
@@ -238,6 +238,8 @@ export default class LandAtlas extends React.PureComponent<Props, State> {
         <div className="atlas-container">
           <Atlas
             showControls
+            showOperator
+            showTenant
```
cazala, line 242: isn't this missing the `showOwner` too?

cazala (commented):
I left a question

src/components/Modals/DeployModal/DeployToLand/LandAtlas/LandAtlas.tsx
LautaroPetaccio: I thought so, but the [Atlas component](https://github.com/decentraland/ui/blob/447d1ec42febbcee16dc407a62e4010f9825d624/src/components/Atlas/Atlas.tsx#L30) seems to be showing them by default with the default layer that it has.

cazala (approved):

PR #2439, opened 2022-11-21
feat: Update single items management

### LINKED ISSUE #2346: Collection Management - Single Items
Supporting single items means supporting some edgy use cases that don’t add much value for users. In order to avoid introducing complications without need we propose to avoid generating new single items.

The following actions are meant to stop creating new single items.

✏️ *Note: legacy single items will be kept in the Single Items section but as soon there are no more single items the whole section should disappear.*

Tasks:
- Disable ability to create new single items
- Replace "Remove from collection" action in dropdown with "Remove item"
- Add confirm to the Remove item action
- Remove Collections/Single Items tabs if the user has no single items

cyaiox:
This PR updates the single items section, avoid creating new single items and lets the users move the currents to existing collections or delete them.

* Disable ability to create new single items
image

* Replace "Remove from collection" action in dropdown with "Remove item"
image

* Add confirm to the Remove item action
image

* Remove Collections/Single Items tabs if the user has no single items
    * image
    * image

Closes #2346

### DIFF
--- a/src/components/CollectionDetailPage/CollectionItem/CollectionItem.tsx
+++ b/src/components/CollectionDetailPage/CollectionItem/CollectionItem.tsx
@@ -43,9 +43,9 @@ export default class CollectionItem extends React.PureComponent<Props> {
     onNavigate(locations.itemEditor({ itemId: item.id, collectionId: item.collectionId }))
   }
 
-  handleRemoveItem = () => {
-    const { item, onRemoveItem } = this.props
-    onRemoveItem(item)
+  handleDeleteItem = () => {
+    const { item, onDeleteItem } = this.props
+    onDeleteItem(item)
   }
 
   renderPrice() {
@@ -149,14 +149,17 @@ export default class CollectionItem extends React.PureComponent<Props> {
                     {item.price ? (
                       <Dropdown.Item text={t('collection_item.edit_price')} onClick={this.handleEditPriceAndBeneficiary} />
                     ) : null}
-                    {!item.isPublished ? (
-                      <ConfirmDelete
-                        name={item.name}
-                        onDelete={this.handleRemoveItem}
-                        trigger={<Dropdown.Item text={t('collection_item.remove_item')} />}
-                      />
-                    ) : null}
                     <ResetItemButton itemId={item.id} />
+                    {!item.isPublished ? (
+                      <>
+                        <Dropdown.Divider />
+                        <ConfirmDelete
+                          name={item.name}
+                          onDelete={this.handleDeleteItem}
+                          trigger={<Dropdown.Item text={t('collection_item.delete_item')} />}
+                        />
+                      </>
+                    ) : null}
                   </>
                 ) : null}
               </Dropdown.Menu>
--- a/src/components/CollectionDetailPage/CollectionItem/CollectionItem.tsx
+++ b/src/components/CollectionDetailPage/CollectionItem/CollectionItem.tsx
@@ -13,6 +13,7 @@ import { isEmoteData, ItemType, WearableData } from 'modules/item/types'
 import ItemBadge from 'components/ItemBadge'
 import RarityBadge from 'components/RarityBadge'
 import ItemImage from 'components/ItemImage'
+import ConfirmDelete from 'components/ConfirmDelete'
 import { Props } from './CollectionItem.types'
 import ResetItemButton from './ResetItemButton'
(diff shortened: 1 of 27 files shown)


src/modules/item/reducer.ts
```diff
@@ -124,6 +125,7 @@ export const INITIAL_STATE: ItemState = {
   data: {},
   rarities: [],
   loading: [],
+  hasUserOrphanItems: false,
```
juanmahidalgo, line 128: Quesdtion here! should we initialize this as `undefined` to avoid race conditions? if not, how we know if the user doesn't have orphans or if the request is still on flight?

cyaiox: hey @juanmahidalgo -- good catch! 

currently, the value is set to `true` or keep `false` when the action `FETCH_ITEMS_SUCCESS` is executed, but it's required to trigger in a few components the `FETCH_ITEMS_REQUEST`. I'll update a little bit the logic.

src/components/ItemEditorPage/CenterPanel/CenterPanel.tsx
```diff
@@ -43,8 +43,9 @@ export default class CenterPanel extends React.PureComponent<Props, State> {
       // The TP collections wouldn't have emotes soon, for this reason, we are fetching only standard collections to show in the Play Emote dropdown
       if (collection && !isTPCollection(collection)) {
         onFetchCollectionItems(collection.id)
-      } else {
-        onFetchOrphanItems(address!)
+      } else if (address) {
+        // TODO: Remove this call when there are no users with orphan items
+        onFetchOrphanItems(address)
```
LautaroPetaccio, line 48: You can use the `hasUserOrphanItems` value to check if it's undefined to fetch the orphan items or not as well, what do you think about that?

src/components/CollectionsPage/CollectionsPage.tsx
```diff
@@ -143,83 +145,90 @@ export default class CollectionsPage extends React.PureComponent<Props> {
...
-    const { collectionsPaginationData, itemsPaginationData, view, isThirdPartyManager, onSetView, isLoadingItems, isLoadingCollections } =
-      this.props
+    const { collectionsPaginationData, itemsPaginationData, view, isLoadingItems, isLoadingCollections, hasUserOrphanItems } = this.props
     const { page } = this.state
     const totalCollections = collectionsPaginationData?.total
     const totalItems = itemsPaginationData?.total
     const count = this.isCollectionTabActive() ? totalCollections : totalItems
     const totalPages = this.isCollectionTabActive() ? collectionsPaginationData?.totalPages : itemsPaginationData?.totalPages
 
+    if (isLoadingItems || isLoadingCollections || count === undefined) {
```
LautaroPetaccio, line 201: Should the loader wait for the UI to load if it has orphan items?

LautaroPetaccio (changes requested):

src/components/CollectionDetailPage/CollectionItem/CollectionItem.tsx
```diff
@@ -42,9 +43,9 @@ export default class CollectionItem extends React.PureComponent<Props> {
     onNavigate(locations.itemEditor({ itemId: item.id, collectionId: item.collectionId }))
   }
 
-  handleRemoveFromCollection = () => {
-    const { item, onRemoveFromCollection } = this.props
-    onRemoveFromCollection(item, null)
+  handleRemoveItem = () => {
+    const { item, onRemoveItem } = this.props
+    onRemoveItem(item)
```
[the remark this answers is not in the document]
cyaiox: thanks for the feedback @LautaroPetaccio! good observation. I agree with you, `Delete item` makes more sense for this feature.

src/modules/item/selectors.ts
```diff
@@ -34,6 +34,8 @@ export const getItem = (state: RootState, itemId: string) => {
   return items.find(item => item.id === itemId) || null
 }
 
+export const hasUserOrphanItems = (state: RootState): boolean | undefined => getState(state).hasUserOrphanItems
```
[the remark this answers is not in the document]
cyaiox: sure!

src/components/CollectionsPage/CollectionsPage.tsx
cyaiox: good catch! it should wait

LautaroPetaccio (approved):

PR #2444, opened 2022-11-23
fix: mobile homepage unity compatibility

### LINKED ISSUE #2441: [BUG] Couldn't load unity loader in Safari
## 🎉 Description

Error accessing the webpage using Safari

Console output:
```
[Log] Renderer version: 1.8.42643 (editor.js, line 40)
[Log] Invoking error handler due to (UnityLoader.js, line 5026)
Script error.
```

* [x] 🐛 This is a bug report.

cyaiox:
This PR avoids loading unity scripts in mobile browsers because the scenes editor is only available for the desktop experience.

Closes #2441

### DIFF
--- a/src/components/Preview/Preview.tsx
+++ b/src/components/Preview/Preview.tsx
@@ -77,9 +77,8 @@ class Preview extends React.Component<Props & CollectedProps> {
     try {
       isDCLInitialized = true
       window.devicePixelRatio = 1 // without this unity blows up majestically 💥🌈🦄🔥🤷🏼‍♂️
-      await injectScript(`${PUBLIC_URL}/unity/Build/hls.min.js`)
-      await injectScript(`${PUBLIC_URL}/editor.js`)
-      await injectScript(`${PUBLIC_URL}/UnityLoader.js`)
+      const scriptsURLs = ['unity/Build/hls.min.js', 'editor.js', 'UnityLoader.js']
+      await Promise.all<void>(scriptsURLs.map(script => injectScript(`${PUBLIC_URL}/${script}`)))
       await editorWindow.editor.initEngine(this.canvasContainer.current, PUBLIC_URL + '/unity/Build/unity.json')
       if (!unityDebugParams) {
         canvas = await editorWindow.editor.getDCLCanvas()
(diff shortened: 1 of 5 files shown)


src/lib/urn.ts
```diff
@@ -122,6 +118,9 @@ export function isThirdParty(urn?: string) {
 }
 
 export function decodeURN(urn: URN): DecodedURN {
+  const urnRegExp = new RegExp(
```
cyaiox, line 121: Moved this const `urnRegExp ` inside the method because `Safari` && `Safari on iOS` doesn't support `lookbehind` in JS regular expressions.

src/routing/utils.ts
```diff
@@ -105,3 +105,13 @@ export function getSortOrder(value: string | null | undefined, defaultValue: 'de
...
+
+export function injectScript(url: string) {
+  const script = document.createElement('script')
+
+  script.src = url
+  script.type = 'text/javascript'
+  script.async = true
+
+  document.body.appendChild(script)
+}
```
LautaroPetaccio, line 117: Can we, by any chance, make this function return a promise when the script finished loading?
The idea is to use this on the component that loads the scene editor so it gets loaded only when it's needed, improving the load times of the site. We would need to use the promise to show a loader before loading the scene editor and we'll need to set up a flag somewhere to know that the script ha already finished loading.

cyaiox: @LautaroPetaccio -- thanks for the feedback! sure, we could return a promise when the script finished loading.

great idea! let me move the load script logic to the scene editor.

LautaroPetaccio (approved):
Works great! It has reduced the JS load time and the size of JS downloaded. Great job!

fzavalia (approved):
LGTM!!!!

src/components/Preview/Preview.tsx
```diff
@@ -76,6 +77,9 @@ class Preview extends React.Component<Props & CollectedProps> {
     try {
       isDCLInitialized = true
       window.devicePixelRatio = 1 // without this unity blows up majestically 💥🌈🦄🔥🤷🏼‍♂️
+      await injectScript(`${PUBLIC_URL}/unity/Build/hls.min.js`)
```
juanmahidalgo, line 80: We can launch all of them at the same time, right? it doesn't need to be in a sequence I think 

```suggestion
      const scriptsURLs = ['unity/Build/hls.min.js', 'editor.js', 'UnityLoader.js']
      await Promise.all<void>(scriptsURLs.map(script => injectScript(`${PUBLIC_URL}/${script}`)))
```

wdyt?

juanmahidalgo (approved):
looks great to me, left a small suggestion

src/components/Preview/Preview.tsx
cyaiox: It works like a charm!

juanmahidalgo (approved):
🚀

PR #2457, opened 2022-12-02
fix: Load intercom when adblockers on

cyaiox:
This PR fixes intercom loading when adblockers are on. The current component implementation depends on the segment `anon_id` field, but if an adblocker is on, the segment library won't load, and the intercom component will be closed.


juanmahidalgo (approved):
LGTM!

PR #2462, opened 2022-12-07
feat: change operator description to better understand the behavior

### LINKED ISSUE #2419: Improve Operator description
Improve builder UI by adding a description that the `operator` is the `update operator` to avoid confusion.

Option: `Update operators can **only** deploy and update scenes and can be removed at any time.`

meelrossi:
This PR changes the description shown when assigning an operator to the land so that the user better understands the permissions this operator has.

Closes #2419 

#### en
image

#### es
image

#### zh
image


cyaiox (approved):
LGTM!

PR #2466, opened 2022-12-08
feat: add item added toast

### LINKED ISSUE #2350: [Creation][ItemEditor] Show a toast after creating an Item and being redirected to the ItemEditor
Show a `toast` in the `bottom-left` side after the user successfully creates an emote and is redirected to the ItemEditor.

This toast has a link to return to the selected collection details.

editor owner

meelrossi:
This PR adds a toast in editor to show that the new item was added to the collection and giving the user the posiblity to see the collection.

This toast should only appear on Editor view when there was a redirection from creating a new item (for now just emotes). If the user continues on the editor (items list) view they should see the toast until they dismiss it. When they go to another page then the toast should be removed.

Closes #2350 

https://user-images.githubusercontent.com/11800206/206569728-2ec2832f-611d-46dd-af8d-e49258582ac7.mov

image
image


src/components/ItemEditorPage/LeftPanel/Toasts/ItemAdded/ItemAddedToast.tsx
```diff
@@ -0,0 +1,67 @@
...
+const ItemAddedToast: React.FC<Props> = props => {
+  const { collectionId } = props
+  const [shouldShowToast, setShouldShowToast] = useState(false)
+  const [item, setItem] = useState<string | undefined>()
+  const search = useLocation().search
+  const history = useHistory()
+
+  useEffect(() => {
+    const searchParams = new URLSearchParams(search)
+    const newItem = searchParams.get('newItem')
```
cyaiox, line 21: could you please move this get `newItem` query param to `modules.locations.selectors` please? like this one: https://github.com/decentraland/builder/blob/master/src/modules/location/selectors.ts#L54

src/components/ItemEditorPage/LeftPanel/Toasts/ItemAdded/ItemAddedToast.tsx
```diff
@@ -0,0 +1,67 @@
...
+      setShouldShowToast(true)
+      setItem(newItem)
+    }
+  }, [search])
+
+  function handleRemoveItemAddedToast() {
+    setShouldShowToast(false)
+    const searchParams = new URLSearchParams(search)
+    searchParams.delete('newItem')
+    history.replace({ search: searchParams.toString() })
```
cyaiox, line 32: what do you think about moving this `replace` logic to an `ItemAddedToast.container.ts` and dispatching the replace like this: https://github.com/decentraland/builder/blob/master/src/components/LandDetailPage/LandDetailPage.container.ts#L25? would worth it? or maybe we should send the `newItem` in the props?

-- 
in a case worth it to create a container, we could use the previously created selector there

cyaiox (changes requested):

cyaiox (approved):
LGTM! great job 🎉

PR #2471, opened 2022-12-13
Feat/cache move items collections request

juanmahidalgo:
The Modal that shows the collections where the item can be moved, was fetching all the collections unpublished every single time it was opened, making the background change to a loading spinner and also re-fetching data unnecessary, because it could be re-used.

### DIFF
--- a/src/components/CollectionDropdown/CollectionDropdown.tsx
+++ b/src/components/CollectionDropdown/CollectionDropdown.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { Dropdown, DropdownItemProps, DropdownProps, Row } from 'decentraland-ui'
+import { Dropdown, DropdownItemProps, DropdownProps, Loader, Row } from 'decentraland-ui'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
 import { Props } from './CollectionDropdown.types'
 import './CollectionDropdown.css'
@@ -47,9 +47,11 @@ export default class CollectionDropdown extends React.PureComponent<Props> {
   }
 
   render() {
-    const { isDisabled } = this.props
+    const { isDisabled, isLoading } = this.props
     const options = this.getOptions()
-    return (
+    return isLoading ? (
+      <Loader active size="mini" />
+    ) : (
       <Dropdown
         className="CollectionDropdown"
         trigger={this.renderTrigger(options)}
--- a/src/modules/collection/reducer.spec.ts
+++ b/src/modules/collection/reducer.spec.ts
@@ -52,30 +52,40 @@ describe('when FETCH_TRANSACTION_SUCCESS', () => {
   })
 })
 
-describe('when FETCH_COLLECTIONS_REQUEST', () => {
-  const anExistingCollectionId = 'id'
-  let mockedFetchCollectionParams: FetchCollectionsParams
+describe('when reducing the FETCH_COLLECTIONS_REQUEST action', () => {
   let initialState: CollectionState
+  let anExistingCollectionId
+  let mockedFetchCollectionParams: FetchCollectionsParams
   beforeEach(() => {
-    mockedFetchCollectionParams = {
-      assignee: 'anAssignee',
-      isPublished: false
-    } as FetchCollectionsParams
+    anExistingCollectionId = 'id'
     initialState = {
       data: {
         [anExistingCollectionId]: {
           id: 'anExistingCollectionId'
         }
-      },
-      lastFetchParams: mockedFetchCollectionParams,
-      loading: []
-    } as any
+      }
+    } as CollectionState
+    mockedFetchCollectionParams = {
+      assignee: 'anAssignee',
+      isPublished: false
+    } as FetchCollectionsParams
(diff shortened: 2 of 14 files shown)


src/modules/collection/reducer.spec.ts
```diff
@@ -51,6 +52,44 @@ describe('when FETCH_TRANSACTION_SUCCESS', () => {
   })
 })
 
+describe('when FETCH_COLLECTIONS_REQUEST', () => {
```
LautaroPetaccio, line 55: Just a suggestion:
```suggestion
describe('when reducing the FETCH_COLLECTIONS_REQUEST action', () => {
```

src/modules/collection/reducer.ts
```diff
@@ -187,7 +197,8 @@ export function collectionReducer(state: CollectionState = INITIAL_STATE, action
               }
             }
           : {}),
-        error: null
+        error: null,
+        lastFetchParams: params
```
LautaroPetaccio, line 201: Would you mind adding a test / changing tests to ensure that this variable is set when the success action is called?

src/modules/collection/reducer.spec.ts
```diff
@@ -51,6 +52,44 @@ describe('when FETCH_TRANSACTION_SUCCESS', () => {
...
+      lastFetchParams: mockedFetchCollectionParams,
+      loading: []
+    } as any
+  })
+
+  describe('and it sends the flag to re-use existing results and same parameters', () => {
+    it('should not set the loading state', () => {
+      const state = reducer(initialState, fetchCollectionsRequest(undefined, mockedFetchCollectionParams, true))
+      expect(reducer(initialState, fetchCollectionsRequest(undefined, mockedFetchCollectionParams, true))).toEqual(state)
+    })
```
LautaroPetaccio, line 79: If I'm not wrong, you are missing the case where the actions has the re-use flag but the parameters don't match, would you mind adding a test for this case?

src/modules/collection/sagas.spec.ts
```diff
@@ -1844,11 +1845,42 @@ describe('when handling the fetch of collections', () => {
...
+        .provide([
+          [select(getWalletItems), []],
+          [select(getLastFetchParams), mockedFetchParameters],
+          [select(getData), alreadyFetchedCollections],
+          [select(getCollectionPaginationData), undefined]
+        ])
+        .put(fetchCollectionsSuccess([collection], undefined, mockedFetchParameters))
+        .dispatch(fetchCollectionsRequest(undefined, mockedFetchParameters, true))
+        .run({ silenceTimeout: true })
+    })
```
LautaroPetaccio, line 1882: I think that it might be missing the case where the pagination data is required, which puts the success with pagination data, would you mind adding this case?

src/modules/collection/sagas.ts
```diff
@@ -180,23 +183,43 @@ export function* collectionSaga(legacyBuilderClient: BuilderAPI, client: Builder
...
+              },
+              params
+            )
+          )
+        } else {
+          yield put(fetchCollectionsSuccess(response, undefined, params))
+        }
       }
     } catch (error) {
+      console.log('error: ', error)
```
LautaroPetaccio, line 222: I think you might have left a console log here, right?

LautaroPetaccio (changes requested):

src/components/CollectionDropdown/CollectionDropdown.tsx
```diff
@@ -47,9 +47,11 @@ export default class CollectionDropdown extends React.PureComponent<Props> {
   }
 
   render() {
-    const { isDisabled } = this.props
+    const { isDisabled, isLoading } = this.props
     const options = this.getOptions()
-    return (
+    return isLoading ? (
+      <Loader active size="mini" />
+    ) : (
```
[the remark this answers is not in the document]
juanmahidalgo: We're overriding the default styles sadly :(, the loader icon has a `display: none`

src/modules/collection/reducer.spec.ts
juanmahidalgo: added!

src/modules/collection/sagas.spec.ts
juanmahidalgo: added!

src/modules/collection/sagas.ts
juanmahidalgo: removed!

src/modules/collection/sagas.spec.ts
```diff
@@ -1844,11 +1846,76 @@ describe('when handling the fetch of collections', () => {
...
+          totalPages: 1
+        }
+        paginationStats = {
+          limit: 1,
+          page: 1,
+          pages: 1,
+          total: 1
+        }
+      })
+      it('should put the success action with the data without pagination information', () => {
```
LautaroPetaccio, line 1890:
```suggestion
      it('should put the success action with the data with pagination information', () => {
```

LautaroPetaccio (approved):
Look great 👍

PR #2472, opened 2022-12-13
feat: add collections sorting

### LINKED ISSUE #1733: [CollectionsPage] Add a sort selector in the collections page
Add a sort selector in the collections page that allows sorting the collections by:
- Created Date
- Last Modified Date
- Name

It should look like this one:

Tasks:
- [ ] Adds a sort selector on the collections page that allows choosing: 
    - `Name ASC | DESC`
    - `Last Modified ASC | DESC`
    - `Created ASC | DESC`
- [ ] Send the sort choice using the query param `sort` to the `builder-api`

meelrossi:
- Add new sorting selector in collections list.
image

- This pr also fixes bug where items tab is being shown even though there are no single items. This is caused because total value is coming as null and not as 0 and we were checking `total !== 0`

Closes #1733 
### With FF on
image

### With FF off
image

### With tabs
image


src/components/CollectionsPage/CollectionsPage.tsx
```diff
@@ -29,20 +30,22 @@ import CollectionCard from './CollectionCard'
 import CollectionRow from './CollectionRow'
 import { Props, TABS } from './CollectionsPage.types'
 import './CollectionsPage.css'
+import { CurationSortOptions } from 'modules/curations/types'
```
juanmahidalgo, line 33: Can we move this import a few line above? 🙏

juanmahidalgo (approved):
just small comment left but looks awesome, great job 👏 👏👏

PR #2519, opened 2023-01-19
feat: Reclaim ENS name

### LINKED ISSUE #2510: Add reclaim operation procedure for unclaimed ENS names
## 🎉 Description

After transferring ENS names, the ENS registry isn't automatically updated to set the new owner as the owner of the subdomain, it must be done manually. Now, users have to do this operation by doing a manual transaction to the DCL Registry contract, making it difficult for non experimented users to use their recently transferred name.

Create a new step in the link assignment section where it is requested to the user to perform a reclaim before doing the next steps of the procedure if the user has not claimed the name before.

To check if the user needs to reclaim the ENS name, a call to the ENS Registry contract needs to be made, more specifically, to the [owner method](https://etherscan.io/address/0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e) which receives a `bytes32 node` as its parameter. To compute the node value, a keccak256 hash has to be built from the ABI encoded result of the domain name hash and the token id of DCL ENS name NFT:
```
keccak256(abi.encodePacked(domainNameHash, bytes32(_tokenId))
```
- The `domainNameHash` is `0x0058500cd570bacbce7dbb11497c280dea3f623a7479be7d8b01c2ae07b623fc` which can be checked at the [DCL registry](https://etherscan.io/address/0x2a187453064356c898cae034eaed119e1663acb8#readContract) if required.
- The token id is the id of the NFT that represents the DCL Name.

To perform the `reclaim` operation, the UI must send a transaction to the [reclaim method](https://github.com/decentraland/avatars-contract/blob/bd49fb20c71aa086351e544ffe0aaed6cce1c5ff/contracts/ens/DCLRegistrar.sol#L230) with the DCL ENS name token id and the address of the user.

* [ ] 🐛 This is a bug report.
* [x] 📈 This is a feature request.

LautaroPetaccio:
This PR adds the capability to reclaim a ENS name after being bought in the marketplace or being transferred. It also adds a new image for the ENS activity, the same one we have in the Marketplace.

Screenshot 2023-01-19 at 17 29 40
Screenshot 2023-01-19 at 17 17 43

Closes #2510


src/modules/ens/sagas.ts
```diff
@@ -308,23 +338,46 @@ export function* ensSaga(builderClient: BuilderClient) {
       const from = wallet.address
 
       const controllerContract = DCLController__factory.connect(CONTROLLER_ADDRESS, signer)
-      const transaction: ethers.ContractTransaction = yield call(() => controllerContract.register(name, from))
-
+      const dclRegistrarContract = DCLRegistrar__factory.connect(REGISTRAR_ADDRESS, signer)
+      console.log('About to register a name', name, from)
```
fzavalia, line 342: Remove the logs

fzavalia (changes requested):

fzavalia (approved):

PR #2524, opened 2023-01-23
feat: add full width to navbar and footer

meelrossi:
Add full width to builder
image


juanmahidalgo (approved):
LGTM! the builder looks really nice with the full width 🤩

PR #2529, opened 2023-01-27
feat: Item Editor navigation

### LINKED ISSUE #2349: Editor Navigation
When entering from the collection detail:

When entering from main dashboard:

Tasks:
- Control using one behavior or the other via a query param
- When the user should return to the CollectionDetail, the icon should be an "X", also when the user would return to the main Collections dashboard. There should be no hamburger icon anymore. When the user returns from the collection to the collections list within the editor, the icon remains an arrow.

cyaiox:
This PR allows going back to the collection detail page if the user came from one.

It also removes deprecated redirection code for single items creation logic.

UI when coming from a collection:
image

UI when wanting to exit the item editor:
image

Closes #2349

### DIFF
--- a/src/components/CollectionDetailPage/CollectionDetailPage.container.ts
+++ b/src/components/CollectionDetailPage/CollectionDetailPage.container.ts
@@ -30,7 +30,7 @@ const mapState = (state: RootState): MapStateProps => {
 }
 
 const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
-  onNavigate: path => dispatch(push(path)),
+  onNavigate: (path, prop) => dispatch(push(path, prop)),
   onOpenModal: (name, metadata) => dispatch(openModal(name, metadata))
 })
 
--- a/src/components/CollectionDetailPage/CollectionDetailPage.container.ts
+++ b/src/components/CollectionDetailPage/CollectionDetailPage.container.ts
@@ -30,7 +30,7 @@ const mapState = (state: RootState): MapStateProps => {
 }
 
 const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
-  onNavigate: (path, prop) => dispatch(push(path, prop)),
+  onNavigate: (path, locationState) => dispatch(push(path, locationState)),
   onOpenModal: (name, metadata) => dispatch(openModal(name, metadata))
 })
 
--- a/src/components/ItemEditorPage/LeftPanel/Header/Header.tsx
+++ b/src/components/ItemEditorPage/LeftPanel/Header/Header.tsx
@@ -15,8 +15,13 @@ export default class Header extends React.PureComponent<Props> {
   }
 
   handleBack = () => {
-    const { onNavigate } = this.props
-    onNavigate(locations.itemEditor())
+    const { collection, isFromCollections, onNavigate } = this.props
+    // If the user came from a collection details page, go back to the same page
+    if (isFromCollections) {
+      onNavigate(locations.collectionDetail(collection!.id))
+    } else {
+      onNavigate(locations.itemEditor())
+    }
   }
 
   handleNewItem = () => {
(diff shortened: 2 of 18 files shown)


src/modules/item/sagas.ts
```diff
@@ -437,19 +437,10 @@ export function* itemSaga(legacyBuilder: LegacyBuilderAPI, builder: BuilderClien
     if (ItemModals.some(modal => openModals[modal])) {
       yield put(closeAllModals())
     } else if (openModals['CreateSingleItemModal']) {
-      if (location.pathname === locations.collections()) {
```
cyaiox, line 440: This code was removed because creating single items is not allowed anymore

src/modules/item/sagas.ts
```diff
@@ -437,19 +437,10 @@ export function* itemSaga(legacyBuilder: LegacyBuilderAPI, builder: BuilderClien
...
-        if (item.type === ItemType.EMOTE) {
-          // Redirect to the item editor
-          yield put(setItems([item]))
-          yield put(push(locations.itemEditor({ itemId: item.id })))
-        } else {
-          // Redirect to the newly created item details
-          yield put(push(locations.itemDetail(item.id)))
-        }
```
cyaiox, line 448: This code was removed because creating single items is not allowed anymore

src/modules/item/sagas.spec.ts
```diff
@@ -1577,18 +1541,6 @@ describe('when handling the save item curation success action', () => {
...
-      .provide([
-        [select(getLocation), { pathname: locations.collections() }],
-        [select(getOpenModals), { CreateSingleItemModal: true }],
-        [select(getAddress), mockAddress]
-      ])
-      .put(push(locations.itemDetail(item.id)))
-      .dispatch(saveItemSuccess(item, {}))
-      .run({ silenceTimeout: true })
-  })
-
```
cyaiox, line 1591: This code was removed because creating single items is not allowed anymore

src/modules/item/sagas.spec.ts
```diff
@@ -528,42 +528,6 @@ describe('when handling the save item success action', () => {
...
-              [select(getAddress), mockAddress]
-            ])
-            .put(push(locations.itemEditor({ itemId: item.id })))
-            .dispatch(saveItemSuccess(item, {}))
-            .run({ silenceTimeout: true })
-        })
-      })
-    })
-  })
-
```
cyaiox, line 566: This code was removed because creating single items is not allowed anymore

src/modules/item/sagas.ts
```diff
@@ -437,19 +437,10 @@ export function* itemSaga(legacyBuilder: LegacyBuilderAPI, builder: BuilderClien
...
-      } else if (location.pathname === locations.collectionDetail(collectionId) && item.type === ItemType.EMOTE) {
+      if (location.pathname === locations.collectionDetail(collectionId) && item.type === ItemType.EMOTE) {
         // Redirect to the item editor
         yield put(setItems([item]))
-        yield put(push(locations.itemEditor({ collectionId, itemId: item.id, newItem: item.name })))
+        yield put(push(locations.itemEditor({ collectionId, itemId: item.id, newItem: item.name, fromParam: 'collections' })))
```
juanmahidalgo, line 443: What do you think if we send the `from` param as part of the router `state`? is the second parameter of the `push` method. It would be something like:

```
yield put(push(locations.itemEditor({ collectionId, itemId: item.id, newItem: item.name }), { fromParam: 'collections' }))
```

src/components/CollectionDetailPage/CollectionDetailPage.container.ts
```diff
@@ -30,7 +30,7 @@ const mapState = (state: RootState): MapStateProps => {
 }
 
 const mapDispatch = (dispatch: MapDispatch): MapDispatchProps => ({
-  onNavigate: path => dispatch(push(path)),
+  onNavigate: (path, prop) => dispatch(push(path, prop)),
```
fzavalia, line 33: What does prop stand for in this case? Can we rename it to something more representative?

fzavalia: like routerProps?

fzavalia (changes requested):

fzavalia (approved):
🚀

src/components/ItemEditorPage/LeftPanel/Header/Header.tsx
```diff
@@ -15,8 +15,13 @@ export default class Header extends React.PureComponent<Props> {
   }
 
   handleBack = () => {
-    const { onNavigate } = this.props
-    onNavigate(locations.itemEditor())
+    const { collection, isFromCollections, onNavigate } = this.props
+    // If the user came from a collection details page, go back to the same page
+    if (isFromCollections) {
+      onNavigate(locations.collectionDetail(collection!.id))
```
juanmahidalgo, line 21: can we avoid this `collection!` by adding `collection && isFromCollections` in the `if` above?

juanmahidalgo (approved):
Just a small question/suggestion but looks great 👏 , approving

PR #2533, opened 2023-01-31
feat: Errors creating items

### LINKED ISSUE #2351: Errors while configuring
There are some limitations that **depend on the wearable category.** For these cases we should show the error only if the user selects a category that cannot be supported.

This same error will show up in the Editor (the `Save` button should be disabled):

Note: we could create a component and reuse it on both sides, that computes the errors according to the category and the metrics. We could reuse a helper `isValid(metrics, category): boolean` to also disable the save buttons.

### LINKED ISSUE #2352: Multiple Files
Throw an error if there is more than 1 GLB per zip file. Only for emotes since wearables can have multiple representations or be smart wearables.

There's an edge case where multiple GLBs are uploaded and only one of them has animations. This is both an invalid wearable (because one GLB has animation) and an invalid emote (it has two GLBs). If we detect this we should throw an error of multiple files for an emote.

cyaiox:
This PR show errors when creating an item to give more context to the creators.

Closes #2351
Closes #2352

### DIFF
--- a/src/components/ItemEditorPage/RightPanel/RightPanel.tsx
+++ b/src/components/ItemEditorPage/RightPanel/RightPanel.tsx
@@ -36,6 +36,8 @@ import {
 import { dataURLToBlob } from 'modules/media/utils'
 import { areEmoteMetrics } from 'modules/models/types'
 import Collapsable from 'components/Collapsable'
+import ErrorMetrics from 'components/Modals/CreateSingleItemModal/ErrorMetrics/ErrorMetrics'
+import { isValidWearableMetrics } from 'components/Modals/CreateSingleItemModal/utils'
 import Input from './Input'
 import Select from './Select'
 import MultiSelect from './MultiSelect'
@@ -285,6 +287,16 @@ export default class RightPanel extends React.PureComponent<Props, State> {
     return hasItem ? this.hasStateItemChanged({ ...this.state, ...newState }, selectedItem!) : false
   }
 
+  isDisabled(item: Item | null) {
+    const { data } = this.state
+    let isValidWearable = true
+    if (!!item && !!data && item.type === ItemType.WEARABLE) {
+      isValidWearable = isValidWearableMetrics(item.metrics, data.category as WearableCategory)
+    }
+
+    return !isValidWearable
+  }
+
   hasSavedItem() {
     const { selectedItem } = this.props
     const { isDirty } = this.state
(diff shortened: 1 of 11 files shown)


src/components/ItemEditorPage/RightPanel/RightPanel.tsx
```diff
@@ -462,6 +474,11 @@ export default class RightPanel extends React.PureComponent<Props, State> {
                           disabled={!canEditItemMetadata}
                           onChange={this.handleChangeCategory}
                         />
+                        {!!data &&
```
juanmahidalgo, line 477: nit-pick, could we store this validation in a variable? 
something like:
```const showErrorMetrics = !!data && item.type === ItemType.WEARABLE && !isValidWearableMetrics(item.metrics, data.category as WearableCategory)```
so the JSX looks a bit cleaner, wdyt?

File: src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx [hunk 2]
```diff
@@ -703,10 +705,13 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
           required
           label={t('create_single_item_modal.category_label')}
           placeholder={t('create_single_item_modal.category_placeholder')}
-          value={categories.includes(category!) ? category : undefined}
+          value={isValidCategory ? category : undefined}
           options={categories.map(value => ({ value, text: t(`${type!}.category.${value}`) }))}
           onChange={this.handleCategoryChange}
         />
+        {type === ItemType.WEARABLE && isValidCategory && !isValidWearableMetrics(metrics as ModelMetrics, category as WearableCategory) ? (
```
juanmahidalgo, line 712: do we need this check `isValidWearableMetrics` here? The `ErrorMetrics` will also do it, right? looks like it will be validated twice

juanmahidalgo (approved):
LGTM! a few small comments but it's approved 👍

src/components/ItemEditorPage/RightPanel/RightPanel.tsx
cyaiox: sure! it will look cleaner 🙌

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
cyaiox: You're right, the `ErrorMetrics` do the same validation in parts to show `textures` errors or `triangles` errors. 
The reason for the validation here is to avoid rendering the component container.

PR #2537, opened 2023-02-01
fix: Collection details UI

cyaiox:


src/components/CollectionDetailPage/CollectionDetailPage.tsx
```diff
@@ -32,13 +30,33 @@ import CollectionItem from './CollectionItem'
...
+  componentDidMount(): void {
+    const { collection } = this.props
+    if (collection) {
+      this.fetchCollectionForumPostReply()
+    }
+  }
+
+  componentDidUpdate(prevProps: Props): void {
+    const { collection } = this.props
+    if (!prevProps.collection && collection) {
```
juanmahidalgo, line 47: Should we add `collection && !collection.forumPostReply` to avoid re-fetches?

src/components/CollectionDetailPage/CollectionDetailPage.tsx
```diff
@@ -32,13 +30,33 @@ import CollectionItem from './CollectionItem'
...
+      this.fetchCollectionForumPostReply()
+    }
+  }
+
+  fetchCollectionForumPostReply() {
+    const { collection, onFetchCollectionForumPostReply } = this.props
+    // Only fetch the forum post replies if the collection has a forum link and there's no other fetch process in progress
+    if (collection && collection.isPublished && !collection.isApproved && collection.forumLink) {
```
juanmahidalgo, line 55: or maybe here we could add the `&& !collection.forumPostReply` check

src/components/CollectionDetailPage/CollectionDetailPage.tsx
```diff
@@ -164,21 +181,37 @@ export default class CollectionDetailPage extends React.PureComponent<Props, Sta
...
             </span>
           </Button>
         )}
       </NetworkCheck>
     )
   }
 
+  renderForumRepliesBadge() {
+    const { collection } = this.props
+    if (collection && collection?.forumPostReply) {
```
juanmahidalgo, line 194:
```suggestion
    if (collection && collection.forumPostReply) {
```
or 
```suggestion
    if (collection?.forumPostReply) {
```
right?

juanmahidalgo (approved):
nice @cyaiox !! 🚀

src/components/CollectionDetailPage/CollectionDetailPage.tsx
cyaiox: This condition will be fulfilled only when the user enters the `CollectionDetails` directly without previously fetching all the collections in the `CollectionsPage` and the `CollectionProvider` updates the state.

Also, we want to fetch the forum post replies every time the user enters the `CollectionDetails` because we don't know when a curator left a new comment in the forum post.

src/components/CollectionDetailPage/CollectionDetailPage.tsx
cyaiox: I like the second one 😬

src/components/CollectionDetailPage/CollectionDetailPage.tsx
juanmahidalgo: makes sense! good catch

PR #2540, opened 2023-02-06
fix: Lazy loading unity scripts for scenes

cyaiox:
This PR fixes loading unity scripts for scenes when creating a new Scene.


juanmahidalgo (approved):
LGTM!

PR #2552, opened 2023-02-14
feat: Add underline to Collection Name in Item Editor page when hover

cyaiox:
This PR adds underling to the Collection Name in the Item Editor page when hovering.


juanmahidalgo (approved):
LGTM

PR #2622, opened 2023-05-24
feat: Show worlds scene details page

cyaiox:


src/components/Modals/DeployModal/DeployToWorld/DeployToWorld.tsx
```diff
@@ -85,6 +100,9 @@ export default function DeployToWorld({
   }, [])
 
   const getExplorerUrl = useMemo(() => {
+    if (isDevelopment) {
+      return `${EXPLORER_URL}/?realm=${WORLDS_CONTENT_SERVER_URL}/world/${world}&NETWORK=goerli`
```
meelrossi, line 104: Is it ok that this is super different from the one in prod?

meelrossi (approved):

src/components/Modals/DeployModal/DeployToWorld/DeployToWorld.tsx
cyaiox: for some reason, the explorer doesn't recognize the realm only with the name in goerli, so it ended up using the main content server instead of the worlds. This way, we're forcing the explorer to use the world's content server

PR #2655, opened 2023-06-09
fix: Loading Scene Editor

cyaiox:
This PR shows a Modal when the unity editor fails to load a scene.

The `Back button` and `Close Modal Action` redirect to the scene details page, forcing a fresh load to clean unity scripts.

The `Reload button` forces the browser to reload and load unity scripts again.

image

### DIFF
--- a/src/components/Preview/Preview.tsx
+++ b/src/components/Preview/Preview.tsx
@@ -4,6 +4,9 @@ import Lottie from 'lottie-react'
 import { config } from 'config'
 
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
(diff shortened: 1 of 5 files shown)


src/components/Preview/Preview.tsx
```diff
@@ -103,6 +119,23 @@ class Preview extends React.Component<Props & CollectedProps> {
     return null
   }
 
+  renderError() {
+    const forceBackReload = () => window.location.replace(locations.sceneDetail(this.props.project.id))
```
meelrossi, line 123: you can do this by using `as="a"` in the button and setting the url as the href

PR #2677, opened 2023-06-22
feat: Fetch scene templates

cyaiox:
This PR adds a new method to fetch scene templates from the `builder-api`.


src/components/ScenesPage/ScenesPage.css
```diff
@@ -6,8 +6,8 @@
   font-size: 17px;
 }
 
-.ScenesPage .button.inverted,
-.ScenesPage .button.inverted:hover {
+.ScenesPage .ui.button.inverted,
```
cyaiox, line 9: not related to this PR. Fixes secondary button color

src/modules/project/selectors.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+      project: {
+        data: {
+          ...mockTemplates,
+          [anotherMockTemplate.id]: anotherMockTemplate
+        }
+      }
+    } as any
+  })
+
+  describe('when getting the templates', () => {
```
meelrossi, line 37: can we add a project that is not a template and make sure its not included?

meelrossi (approved):

PR #2684, opened 2023-06-23
fix: templates feature design fixes

meelrossi:
Fix. multiple issues in templates feature


src/components/SceneViewPage/SceneViewPage.tsx
```diff
@@ -140,7 +140,12 @@ export default class SceneViewPage extends React.PureComponent<Props> {
             <Container className="back-container">
               <Back absolute onClick={onBack}></Back>
               {isTemplatesEnabled && (
-                <Button secondary onClick={this.handleExportProject.bind(this)}>
+                <Button
+                  secondary
+                  onClick={this.handleExportProject.bind(this)}
+                  loading={!this.props.currentProject}
```
cyaiox, line 146: should `currentProject` go here instead of `this.props.currentProject`?

src/components/Modals/CustomLayoutModal/CustomLayoutModal.tsx
```diff
@@ -9,6 +9,7 @@ import { ProjectLayout } from 'modules/project/types'
 import { Props, State } from './CustomLayoutModal.types'
 
 import './CustomLayoutModal.css'
+import { PreviewType } from 'modules/editor/types'
```
cyaiox, line 12: could you please move this import above `.css`?

src/components/SceneViewPage/SceneViewPage.tsx
meelrossi: currentProject is getting the project using the function `getCurrentProject` which value is different than the current project. from the props

PR #2691, opened 2023-06-27
fix: templates download

meelrossi:
Fix several issues in tempaltes

### DIFF
--- a/src/modules/deployment/sagas.ts
+++ b/src/modules/deployment/sagas.ts
@@ -102,7 +102,6 @@ export function* deploymentSaga(builder: BuilderAPI, catalystClient: CatalystCli
         yield put(takeScreenshot())
 
         yield put(setProgress(ProgressStage.NONE, 90))
-        console.log(additionalInfo)
         yield call(() => builder.deployToPool(project.id, additionalInfo))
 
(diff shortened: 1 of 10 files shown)


src/modules/deployment/sagas.ts
```diff
@@ -102,6 +102,7 @@ export function* deploymentSaga(builder: BuilderAPI, catalystClient: CatalystCli
         yield put(takeScreenshot())
 
         yield put(setProgress(ProgressStage.NONE, 90))
+        console.log(additionalInfo)
```
cyaiox, line 105: a wild `console.log` has appeared!

src/modules/scene/reducer.ts
```diff
@@ -77,7 +77,9 @@ const baseSceneReducer = (state: SceneState = INITIAL_STATE, action: SceneReduce
             },
             limits: {
               ...state.data[sceneId].limits,
-              ...limits
+              ...limits,
+              // INCREASE MATERIALS LIMIT FOR TEMPLATES
+              materials: (limits.materials || state.data[sceneId].limits.materials) * 1.1
```
cyaiox, line 82: is it possible for the `limits.materiales` to be undefined?

cyaiox (approved):

PR #2694, opened 2023-06-27
fix: Template cloning address

LautaroPetaccio:
This PR fixes an issue that showed a freshly published project that came up from a template as usynced.
The issue occurred when cloning the project that came from the template, where the address was not set, making the reducer store a project with a different address than the one the user has.
The PR also removes unnecessary values from some containers.

### DIFF
--- a/src/components/TemplateDetailPage/TemplateDetailPage.types.ts
+++ b/src/components/TemplateDetailPage/TemplateDetailPage.types.ts
@@ -2,16 +2,15 @@ import { Dispatch } from 'redux'
 import { CallHistoryMethodAction } from 'connected-react-router'
 import { Project } from 'modules/project/types'
 import { openModal, OpenModalAction } from 'modules/modal/actions'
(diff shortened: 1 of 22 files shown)


src/components/TemplateDetailPage/TemplateDetailPage.types.ts
```diff
@@ -2,16 +2,15 @@ import { Dispatch } from 'redux'
 import { CallHistoryMethodAction } from 'connected-react-router'
 import { Project } from 'modules/project/types'
 import { openModal, OpenModalAction } from 'modules/modal/actions'
-import { duplicateProject, DuplicateProjectAction } from 'modules/project/actions'
+import { DuplicateProjectAction } from 'modules/project/actions'
```
cyaiox, line 5: removing the `onDuplicate` dispatch, this action is no longer required

cyaiox (approved):

PR #2702, opened 2023-06-28
fix: handle login redirect to scenes when came from scens

flobarreto:


src/routing/Routes.tsx
```diff
@@ -15,6 +15,7 @@ import UnsupportedBrowserPage from 'components/UnsupportedBrowserPage'
 import { isDevelopment } from 'lib/environment'
 
 import { Props, State } from './Routes.types'
+import { ProtectedRoute } from 'modules/ProtectedRoute'
```
cyaiox, line 18: could you please move this import above?

cyaiox (approved):
greaat job! 🔥

PR #2706, opened 2023-06-28
fix: Duplicate project save thumbnail

cyaiox:
This PR fixes the save thumbnail error when duplicating projects.


cyaiox:
Fixed here: https://github.com/decentraland/builder/pull/2707/files

Issue #2730, opened 2023-07-06
[CreateSingleItemModal] Create a new step to upload files in the Smart Wearable Creation Flow


cyaiox:
Modify the `CreateSingleItemModal` and create a new step when creating a SmartWearable to upload a video preview.

Tasks:
- [x] Modify the `Item` type to add the new property: `video`
- [ ] Create a new step to upload a video preview
- [x] Modify the sagas handler: `handleSaveItemRequest` to send the video in the `Item.video` property
- [x] Modify the method: `saveItemContents` to send the video file to the new endpoint `/items/:id/video` created

PR #2737, opened 2023-07-11
feat: Handle video preview upload on single item creation

kevinszuchet:
#2730


src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
```diff
@@ -727,8 +818,8 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
             <Header sub>{t('create_single_item_modal.representation_label')}</Header>
             <Row>
               {this.renderRepresentation(BodyShapeType.BOTH)}
-              {this.renderRepresentation(BodyShapeType.MALE)}
-              {this.renderRepresentation(BodyShapeType.FEMALE)}
+              {!isSmartItem && this.renderRepresentation(BodyShapeType.MALE)}
```
cyaiox, line 821: wdyt about adding this validation in the condition above? 

`isAddingRepresentation || isSmartItem ? null : ( ... )`

kevinszuchet: This will be in the next PR with the real UI 😄

PR #2738, opened 2023-07-11
feat: add sepolia support

lauti7:
this shouldn't be merged until the migration is done


src/config/env/dev.json
```diff
@@ -1,5 +1,6 @@
 {
-  "CHAIN_ID": "5",
+  "NETWORK": "sepolia",
```
fzavalia, line 2: The network name can be inferred by the chainId. Can we avoid adding an extra variable?

fzavalia (changes requested):

fzavalia (changes requested):
Update the peer url  `"PEER_URL": "https://peer.decentraland.zone",` to the one in sepolia

src/components/WorldListPage/WorldListPage.tsx
```diff
@@ -24,6 +24,7 @@ import LoggedInDetailPage from 'components/LoggedInDetailPage'
 import { NavigationTab } from 'components/Navigation/Navigation.types'
 import { Props, SortBy } from './WorldListPage.types'
 import './WorldListPage.css'
+import { ChainId, getChainName } from '@dcl/schemas'
```
fzavalia, line 27: Move dependency from node_modules to the top of the other local dependencies.

Underneath decentraland-ui

fzavalia (changes requested):

fzavalia (approved):

PR #2758, opened 2023-07-17
feat: add removes hiding default

meelrossi:
Add new property in wearables that describes weather it should hide the hands by default or not

Any wearable that has removesDefaultHiding including HANDS_WEAR will be considered that it was created when the base mesh of male and female have already separate the hands from the upper body


cyaiox (approved):
🚀 🚀

PR #2764, opened 2023-07-19
feat: initial support for web editor

cazala:
Initial support for Web Editor
- Changed the `Scene` schema to support SDK6 and SDK7 scenes.
- Added migrations to convert outdated scenes into the new schema.
- Integrated `@dcl/inspector` into a new `/inspector` route and added ability to open a scene in it.
- Hidden everything behind a feature flag so it can be merged without releasing the feature.

Needs to be merged along with this https://github.com/decentraland/builder-server/pull/681


src/components/InspectorPage/InspectorPage.types.ts
```diff
@@ -0,0 +1,19 @@
...
+  onOpen: typeof openInspector
+  onConnect: typeof connectInspector
+}
+
+export type State = {
+  isLoaded: boolean
+}
+
+export type MapStateProps = Pick<Props, 'isLoggedIn' | 'scene'>
+// eslint-disable-next-line @typescript-eslint/ban-types
```
meelrossi, line 17: is this necessary?

src/modules/scene/reducer.ts
```diff
@@ -49,40 +48,78 @@ const INITIAL_STATE: SceneState = {
...
       const { scene } = action.payload
-
+      if (scene.sdk6) {
+        return {
+          ...state,
+          data: {
+            ...state.data,
+            [scene.sdk6.id]: {
+              ...scene,
+              sdk6: { ...scene.sdk6, components: { ...scene.sdk6.components }, entities: { ...scene.sdk6.entities } }
```
meelrossi, line 60: why do we need this rewrite?

meelrossi (approved):

src/components/InspectorPage/InspectorPage.types.ts
cazala: it's not! good catch :)

src/modules/scene/reducer.ts
cazala: it shouldn't, but I saw that the [original reducer](https://github.com/decentraland/builder/blob/master/src/modules/scene/reducer.ts#L62) does it that way, and I didn't want to risk that there's some mutation going on somewhere else that this cloning fixes 😅

PR #2789, opened 2023-07-25
feat: add sdk indicator to scenes

meelrossi:
Add sdk tag to scenes info

## Project list
image

## Project detail page
image

## Template detail page
image

## Scene pool page
image

## When loading scene
image


cazala:
All the indicators seem to be working perfect form me except for one 🤔 When I navigate from the homepage into a scene from the scene pool, the indicator never changes from loading state. But if I navigate directly to the pool via link, or if I refresh the page after navigating from the homepage, it does finish loading. Here's an example:


https://github.com/decentraland/builder/assets/2781777/ae36499b-0365-4803-8d9c-599c54c248f5

cazala (approved):

PR #2790, opened 2023-07-26
feat: add sdk7 creation flow

meelrossi:
Add new creation flow for sdk scenes

image


src/modules/project/sagas.ts
```diff
@@ -88,19 +88,32 @@ export function* projectSaga(builder: BuilderAPI) {
...
-      },
-      sdk7: null
+    let scene: Scene
+
+    if (sdk === SDKVersion.SDK7) {
+      scene = {
+        sdk6: null,
+        sdk7: {
+          id: uuidv4(),
+          composite: '',
```
cazala, line 100: The composite should not be an empty string otherwise the inspector would not work properly.. you can see how to generate a composite [here](https://github.com/decentraland/builder/blob/master/src/modules/inspector/utils.ts#L20-L56), you can skip the for loop part but at minimum the composite should have a Scene component with the amount of parcels that the user selected in the "size" step, you can copy that logic from the line 46 in the file on the link above

cazala (approved):

PR #2795, opened 2023-07-28
feat: Allow clicking in the dropzone for uploading files

kevinszuchet:
Let users tap in the middle of the dropzone opening the same window as when they click on "browse from your computer".


src/components/Modals/CreateSingleItemModal/ImportStep/ImportStep.tsx
```diff
@@ -252,12 +253,16 @@ export default class ImportStep extends React.PureComponent<Props, State> {
...
+    window.open('https://docs.decentraland.org/decentraland/creating-wearables/', '_blank', 'noopener noreferrer')
+  }
+
   renderMoreInformation() {
     return (
       <span>
         {t('create_single_item_modal.import_information', {
           link: (
-            <a href="https://docs.decentraland.org/decentraland/creating-wearables/" target="_blank" rel="noopener noreferrer">
+            <a href="javascript" onClick={preventDefault(this.handleOpenMoreInformation)}>
```
cyaiox, line 265: could `<Button basic onClick={preventDefault(this.handleOpenMoreInformation)}>` works in this scenario?

src/lib/event.ts
```diff
@@ -0,0 +1,23 @@
+import { SyntheticEvent } from 'react'
```
cyaiox, line 1: do you mind adding some tests for this implementation?

cyaiox (approved):

PR #2796, opened 2023-07-28
fix: Learn more about rarities link should go to the wearable docs

kevinszuchet:
The docs page with link: https://docs.decentraland.org/creator/wearables-and-amotes/manage-collections/uploading-smart-wearables doesn't exist.


kevinszuchet:
The missing page is being created [here](https://github.com/decentraland/builder/pull/2796)

PR #2798, opened 2023-07-28
feat: Upload video back takes the previous view into account

kevinszuchet:
It considers the previous view to navigate using the back button in the header.

When the user is viewing the item's details and wants to modify the video, if the user clicks the back button, the component will consider that the previous view was the Item Details and render that again.

### DIFF
--- a/src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
+++ b/src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
@@ -722,7 +722,6 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
     const keys = Object.keys(this.state)
     const { fromView } = this.state
 
-    console.log('fromView', fromView)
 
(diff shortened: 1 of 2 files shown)


src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
```diff
@@ -719,6 +720,15 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
 
   handleUploadVideoGoBack = () => {
     const keys = Object.keys(this.state)
+    const { fromView } = this.state
+
+    console.log('fromView', fromView)
```
cyaiox, line 725: `a wild console.log has appeared!`

kevinszuchet, line 725:
```suggestion
```

cyaiox (approved):

PR #2803, opened 2023-07-31
feat: upload assets

cazala:
This PR allows for the upload of assets to the Builder server, it is necessary not only for importing custom assets, but or using Builder assets from the catalog.

It needs this PR to be merged first: https://github.com/decentraland/builder-server/pull/685

### DIFF
--- a/src/modules/inspector/sagas.ts
+++ b/src/modules/inspector/sagas.ts
@@ -307,6 +307,8 @@ export function* inspectorSaga(builder: BuilderAPI, store: RootStore) {
       }
       yield put(updateScene(newScene))
     }
+    // remove from memory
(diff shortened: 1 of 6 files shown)


src/modules/inspector/sagas.ts
```diff
@@ -291,8 +298,6 @@ export function* inspectorSaga(store: RootStore) {
 
     const scene: SceneSDK7 = yield getScene()
 
-    FILES.delete(path)
```
meelrossi, line 294: should we remove it from memory at least? this delete is expected to delete it from the scene only or its supposed to delete the uploaded file too?

meelrossi (approved):

PR #2805, opened 2023-07-31
feat: add topbar to inspector

meelrossi:
- Add topbar to inspector view
- Rename old topbar to SDK6TopBar
- Move isInspectorEnabled condition to InspectorsPage so that the inspector can be refreshed without redirecting to `/` page (as the feature flags are not yet loaded)
image


cazala:
The layout editor does not really work on SDK7 projects.. it is used to initialize the Scene component once the scene is created, but afterwards what matters is the value of the Scene component.. maybe we should disable it when the project is SDK7?

Screenshot 2023-08-01 at 15 41 15

PR #2815, opened 2023-08-02
feat: Add profile dApp redirection

cyaiox:
This PR updates the SceneView author element to redirect to the profile dApp when the FF: `dapps-profile` is enabled.

image

Closes https://github.com/decentraland/profile/issues/74


LautaroPetaccio (approved):
🚀

PR #2817, opened 2023-08-04
feat: SDK7 thumbnails

cazala:
This PR adds support for SDK7 projects to have automatic thumbnails like the SDK6 ones, taking screenshots from the `@dcl/inspector`

Screenshot 2023-08-04 at 09 31 38


src/modules/inspector/sagas.ts
```diff
@@ -114,6 +122,21 @@ export function* inspectorSaga(builder: BuilderAPI, store: RootStore) {
         return promise
       })
     }
+
+    // wait for RPC to be idle (3 seconds)
+    yield waitForRpcIdle(3000)
```
cazala, line 127: There's no easy way of telling when the scene is fully loaded from the context of the builder.. and during my tests I found that sometimes screenshots could be taken too early, before the scene was loaded, because we are taking screenshots any time the composite is modified, and sometimes it happens at boot time, as soon as you open the inspector on an already exisiting scene. To avoid this I disabled the screenshots when the inspector is booted, and enabled them after a few seconds of idleness.

PR #2821, opened 2023-08-04
feat: Single Sign On

fzavalia:


package.json
```diff
@@ -129,7 +130,8 @@
       "^three/.*$": "identity-obj-proxy",
       "^@babylonjs.*$": "identity-obj-proxy",
       "^.*/modules/curations/collectionCuration/toasts.*$": "identity-obj-proxy",
-      "@dcl/ecs": "identity-obj-proxy"
+      "@dcl/ecs": "identity-obj-proxy",
+      "@dcl/single-sign-on-client": "identity-obj-proxy"
```
fzavalia, line 134: As the module is an ESM and not CommonJs, jest doesn't like it. Doing this will replace it in tests with a mockable stub.

src/modules/common/store.ts
```diff
@@ -80,8 +80,7 @@ const { storageMiddleware, loadStorageMiddleware } = createStorageMiddleware({
     ['scene', 'present'],
     ['ui', 'dashboard'],
     ['auth', 'data'],
-    ['sync', 'localProjectIds'],
-    ['identity', 'data']
```
fzavalia, line 84: By removing this, the identity is not persisted in local storage anymore. It is now persisted in the iframe so there is no reason for it.

cyaiox (approved):

PR #2828, opened 2023-08-09
feat: save crdt file

meelrossi:
Save crdt file on change and when project is migrated from sdk6 to sdk7. This file will be used on the preview feature

### DIFF
--- a/src/modules/scene/sagas.ts
+++ b/src/modules/scene/sagas.ts
@@ -69,14 +69,15 @@ import {
   DuplicateProjectSuccessAction,
   duplicateProjectRequest
 } from 'modules/project/actions'
-import { toComposite, toMappings } from 'modules/inspector/utils'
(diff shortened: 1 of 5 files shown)


src/modules/scene/sagas.ts
```diff
@@ -88,687 +89,689 @@ export function* sceneSaga() {
...
+        composite,
+        mappings
+      }
 
-    yield put(updateScene(newSDK7Scene))
-    yield put(push(locations.inspector(project.id)))
-    yield put(migrateToSDK7Success())
-  } catch (error) {
-    put(migrateToSDK7Failure(error))
+      yield builderApi.uploadCrdt(crdt, project.id)
```
cazala, line 769: mm shouldn't this be 
```
yield call([builderApi, 'uploadCrdt'], crdt, project.id)
```
?

cazala (approved):

PR #2924, opened 2023-09-28
feat: Add NameTabs component

### LINKED ISSUE #2910: In new Worlds view, Add tabs to separate DCL names view and ENS names view
When selecting one or the other, it should support displaying different views.

This will be used to display DCL names on one side and ENS names on the other.

fzavalia:
- Integrates it inside WorldListPage just to check it out. The page will be completely overhauled.
- The selected tab depends on the `tab` query param
- Defaults to dcl tab query param if the provided one is invalid

Closes #2910 

image

image

Based but not exact to figma given that it is not the final design.

image

### DIFF
--- a/src/components/WorldListPage_WorldsForEnsOwnersFeature/WorldListPage.tsx
+++ b/src/components/WorldListPage_WorldsForEnsOwnersFeature/WorldListPage.tsx
@@ -23,6 +23,7 @@ import Icon from 'components/Icon'
 import LoggedInDetailPage from 'components/LoggedInDetailPage'
 import { NavigationTab } from 'components/Navigation/Navigation.types'
 import { Props, SortBy } from './WorldListPage.types'
+import NameTabs from './NameTabs'
 import './WorldListPage.css'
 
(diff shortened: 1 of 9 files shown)


src/components/WorldListPage_WorldsForEnsOwnersFeature/WorldListPage.tsx
```diff
@@ -239,6 +240,8 @@ const WorldListPage: React.FC<Props> = props => {
       isLoading={isLoading}
       isPageFullscreen={true}
     >
+      <h1>Worlds</h1>
```
LautaroPetaccio, line 243: Is this a placeholder for something else?

LautaroPetaccio (approved):
🚀

src/components/WorldListPage_WorldsForEnsOwnersFeature/WorldListPage.tsx
fzavalia: It's actually part of the design to have that title.

PR #2935, opened 2023-10-03
fix: Show corresponding error

### LINKED ISSUE #2915: Show proper message when deploying to an ENS world and it fails because it is larger than 25mb

### LINKED ISSUE #2916: Show proper message when deploying to a DCL world and it fails because it is larger than the storage the user has available

fzavalia:
Closes #2915 
Closes #2916 

image

### DIFF
--- a/src/modules/translation/languages/es.json
+++ b/src/modules/translation/languages/es.json
@@ -484,7 +484,8 @@
       },
       "failure": {
         "title": "Algo salió mal",
-        "subtitle": "No pudimos publicar tu escena"
+        "subtitle": "No pudimos publicar tu escena",
+        "subtitle_size_error": "The scene is too big to be published in a World.<br></br>The maximum size is <b>{maxSizeMbs} mbs</b>, but the current scene is <b>{deployedSizedMbs} mbs</b> large."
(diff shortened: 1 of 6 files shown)


src/modules/translation/languages/es.json
```diff
@@ -484,7 +484,8 @@
       },
       "failure": {
         "title": "Algo salió mal",
-        "subtitle": "No pudimos publicar tu escena"
+        "subtitle": "No pudimos publicar tu escena",
+        "subtitle_size_error": "The scene is too big to be published in a World.<br></br>The maximum size is <b>{maxSizeMbs} mbs</b>, but the current scene is <b>{deployedSizedMbs} mbs</b> large."
```
LautaroPetaccio, line 488: What do you think about using the MB / Mb notation for the size?
```suggestion
        "subtitle_size_error": "The scene is too big to be published in a World.<br></br>The maximum size is <b>{maxSizeMbs} mbs</b>, but the current scene is <b>{deployedSizedMbs} MBs</b> large."
```

LautaroPetaccio (approved):

PR #2947, opened 2023-10-05
fix: Re-request wallet stats

fzavalia:
On clear deployment and deployment success

### DIFF
--- a/src/modules/worlds/sagas.spec.ts
+++ b/src/modules/worlds/sagas.spec.ts
@@ -1,11 +1,14 @@
 import { expectSaga } from 'redux-saga-test-plan'
-import { call } from 'redux-saga/effects'
+import { call, select } from 'redux-saga/effects'
 import { throwError } from 'redux-saga-test-plan/providers'
 import { connectWalletSuccess } from 'decentraland-dapps/dist/modules/wallet/actions'
 import { Wallet } from 'decentraland-dapps/dist/modules/wallet/types'
 import { WorldsWalletStats, content } from 'lib/api/worlds'
 import { fetchWorldsWalletStatsFailure, fetchWorldsWalletStatsRequest, fetchWorldsWalletStatsSuccess } from './actions'
 import { worldsSaga } from './sagas'
+import { clearDeploymentSuccess, deployToWorldSuccess } from 'modules/deployment/actions'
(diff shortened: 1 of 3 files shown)


src/modules/worlds/sagas.spec.ts
```diff
@@ -63,11 +66,68 @@ describe('when handling the request action to fetch worlds stats for a wallet',
...
+        .not.put(fetchWorldsWalletStatsRequest(address))
+        .dispatch(clearDeploymentSuccess(''))
+        .silentRun()
+    })
+  })
+
+  describe('when there is an address in the store', () => {
+    it('should put the request action to fetch worlds wallet stats with the stored address', () => {
+      return expectSaga(worldsSaga)
+        .provide([[select(getAddress), address]])
```
LautaroPetaccio, line 127: As `expectSaga` runs the saga until completion and, when testing `handleActionsThatTriggerFetchWorldsWalletStatsRequest`, the `fetchWorldsWalletStatsRequest` action is put and, taking into consideration that the sagas also handles that action, the `fetchWorldsWalletStatsRequest` should be ran as well. To prevent this, we can mock the put:
```suggestion
        .provide([[select(getAddress), address], [put(fetchWorldsWalletStatsRequest(address)), undefined]])
```

fzavalia: I'm certain that is unnecessary due to the nature of sagas.

`fetchWorldsWalletStatsRequest` will be handled by the saga yes, but it will not proceed on any undeclared effect.

For example, it will run and stop were said in the code:

```js
function* handlefetchWorldsWalletStatsRequest(action: FetchWalletWorldsStatsRequestAction) {
  const { address } = action.payload

  try {
    // Stops here given that the call is not declared in the test
    const stats: WorldsWalletStats | null = yield call([WorldsAPIContent, WorldsAPIContent.fetchWalletStats], address)
```

fzavalia: Besides if I set the request put in the .provide, it will not call the .put that comes afterwards, making the test not declaratively useful given that the things tested are what comes after the .provide

fzavalia: I'm mistaken with this, updated it per your suggestion

LautaroPetaccio (approved):

PR #2981, opened 2023-11-06
fix: collection name with emoji

meelrossi:
Change regex to remove any non ascii character from base64 wearable


fzavalia (commented):
Does this affect in anyway the publishing of the wearable or is it purely for the ui?

fzavalia (approved):

PR #2998, opened 2023-12-16
feat: SDK7 templates

cazala:
This PR does the following:
- Shows templates from `@dcl/builder-templates` package instead of loading them from the Builder Server
- Adds support for SDK7 templates
- Removes the creation flow of SDK6 scenes (if a user has SDK6 scenes they can still open them on the legacy builder).
- Everything is behind feature flags (one for enabling the SDK7 templates and one to enable the create from scratch only with SDK7)

### DIFF
--- a/src/components/Modals/CustomLayoutModal/CustomLayoutModal.tsx
+++ b/src/components/Modals/CustomLayoutModal/CustomLayoutModal.tsx
@@ -104,7 +104,7 @@ export default class CustomLayoutModal extends React.PureComponent<Props, State>
   }
 
   renderModalActions = () => {
-    const { isSDK7TemplatesEnabled } = this.props
+    const { isCreateSceneOnlySDK7Enabled } = this.props
     const { hasError, name, step } = this.state
 
     switch (step) {
@@ -128,7 +128,7 @@ export default class CustomLayoutModal extends React.PureComponent<Props, State>
             <Button
               primary
               disabled={hasError || !name}
-              onClick={isSDK7TemplatesEnabled ? this.handleSubmit.bind(this, SDKVersion.SDK7) : this.handleNext}
+              onClick={isCreateSceneOnlySDK7Enabled ? this.handleSubmit.bind(this, SDKVersion.SDK7) : this.handleNext}
             >
               {t('global.next')}
             </Button>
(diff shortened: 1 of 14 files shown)


src/components/Modals/CustomLayoutModal/CustomLayoutModal.tsx
```diff
@@ -27,6 +27,7 @@ export default class CustomLayoutModal extends React.PureComponent<Props, State>
   handleBack = () => {
     const { step } = this.state
     switch (step) {
+      // TODO: remove this after removing the SDK7_TEMPLATES feature flag
```
cazala, line 30: I left several of these comments through the codebase to remember that once we remove the `SDK7_TEMPLATES` (because we won't want SDK6 templates anymore) then these are parts of the code that don't need to exist anymore (like the SDK version selection step)

src/modules/features/selectors.ts
```diff
@@ -53,3 +53,19 @@ export const getIsWorldsForEnsOwnersEnabled = (state: RootState) => {
...
+
+export const getIsSDK7TemplatesEnabled = (state: RootState) => {
+  try {
+    return getIsFeatureEnabled(state, ApplicationName.BUILDER, FeatureName.SDK7_TEMPLATES)
+  } catch (e) {
+    return false
+  }
+}
+
+export const getIsCreateSceneOnlySDK7Enabled = (state: RootState) => {
```
LautaroPetaccio, line 65: Would you mind adding a simple test for these functions?

src/modules/project/selectors.ts
```diff
@@ -22,7 +22,9 @@ export const getUserProjects = createSelector(getAddress, getData, (address, pro
     const project = projects[projectId]
     const isOwnedByUser = !!project.ethAddress && !!address && isEqual(project.ethAddress, address)
     if (isOwnedByUser || project.ethAddress === null) {
-      record[projectId] = project
+      if (!project.isTemplate) {
```
LautaroPetaccio, line 25: What do you think about joining the conditions?
```
(isOwnedByUser || project.ethAddress === null) && !project.isTemplate
```

LautaroPetaccio (commented):
I've left two simple comments, but it looks good!

LautaroPetaccio (approved):

PR #3003, opened 2023-12-22
feat: add sign in thorugh auth dapp

meelrossi:
Add sign in flow with auth dapp


src/modules/features/selectors.ts
```diff
@@ -69,3 +69,11 @@ export const getIsCreateSceneOnlySDK7Enabled = (state: RootState) => {
     return false
   }
 }
+
+export const getIsAuthDappEnabled = (state: RootState) => {
```
LautaroPetaccio, line 73: Would you mind adding a test for this ff?

LautaroPetaccio (approved):

PR #3020, opened 2024-01-23
feat: add address field

meelrossi:
Add new address field to flows:
- Add minters
- Mint item
- Add collaborator
- Transfer land
- Set operator


src/components/Modals/ManageCollectionRoleModal/EmptyRole/EmptyRole.tsx
```diff
@@ -22,24 +23,29 @@ export default class EmptyRole extends React.PureComponent<Props, State> {
     onCancel()
   }
 
-  handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+  handleChange = (_: React.ChangeEvent<HTMLInputElement>, { value }: InputOnChangeData) => {
```
LautaroPetaccio, line 26: Isn't changing the `handleChange` method here breaking the old Field `onChange` handler?

src/components/LandOperatorPage/LandOperatorForm/LandOperatorForm.tsx
```diff
@@ -72,18 +73,34 @@ export default class LandOperatorForm extends React.PureComponent<Props, State>
...
+          <AddressField
+            placeholder="0x..."
+            label={t('operator_page.address')}
+            className={classes.join(' ')}
+            value={address}
+            onChange={this.handleChange}
+            loading={loading}
+            action={isRevokable ? (revoked ? t('operator_page.undo') : t('operator_page.revoke')) : undefined}
+            onAction={isRevokable ? (revoked ? this.handleUndo : this.handleRevoke) : undefined}
+            error={hasError}
```
LautaroPetaccio, line 86: Will `hasError` work with the new field?

src/components/Modals/ManageCollectionRoleModal/EmptyRole/EmptyRole.tsx
meelrossi: it shouldn't , its just changing where it takes the value from but the function signature is the samae one for previous Field component

src/components/LandOperatorPage/LandOperatorForm/LandOperatorForm.tsx
meelrossi: yess it will take the value coming for props or the value calculated on the field

LautaroPetaccio (approved):

PR #3021, opened 2024-01-23
feat: add empty state names

meelrossi:
image


LautaroPetaccio (approved):

src/components/ENSListPage/ENSListPage.tsx
```diff
@@ -421,7 +445,7 @@ export default class ENSListPage extends React.PureComponent<Props, State> {
               activePage={page}
               setPage={page => this.setState({ page })}
               totalPages={totalPages}
-              empty={() => null}
+              empty={() => this.renderEmptyEnsList()}
```
LautaroPetaccio, line 448: This might work as well if the typings are not strict:
```suggestion
              empty={this.renderEmptyEnsList}
```

meelrossi: The build failed with this 😕

PR #3026, opened 2024-01-27
feat: edit metadata

cazala:
Enable scene metadata inspector in web editor

Screenshot 2024-01-26 at 23 25 34


package.json
```diff
@@ -10,6 +10,7 @@
     "@dcl/content-hash-tree": "^1.1.3",
     "@dcl/crypto": "^3.4.5",
     "@dcl/hashing": "^3.0.4",
+    "@dcl/inspector": "^7.3.40",
```
cyaiox, line 13: is it required to install the `@dcl/inspector` package or we could upgrade the `@dcl/sdk` package?

cyaiox (approved):

package.json
cazala: ahh you are right, it’s the sdk we have to update

PR #3033, opened 2024-02-02
fix: Handle multiple representation for SmartWearables

cyaiox:
This PR fixes uploading both representations for smart wearables.


src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
```diff
@@ -181,20 +181,36 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
...
+      ...male,
+      ...(bodyShape === BodyShapeType.BOTH || bodyShape === BodyShapeType.MALE ? this.prefixContents(BodyShapeType.MALE, all) : {})
+    }
+
+    female = {
+      ...female,
+      ...(bodyShape === BodyShapeType.BOTH || bodyShape === BodyShapeType.FEMALE ? this.prefixContents(BodyShapeType.FEMALE, all) : {})
+    }
+
+    all = {
```
meelrossi, line 208: wont this now have repeated things? with and without the prefix?

cyaiox: it will have only the prefixed props.. I'll rename the var because it's confusing 😅

PR #3036, opened 2024-02-06
fix: Skip setup explorer url

fzavalia:
So users don't have to see the greeting page each time they want to preview.

### DIFF
--- a/src/components/InspectorPage/TopBar/TopBar.tsx
+++ b/src/components/InspectorPage/TopBar/TopBar.tsx
@@ -1,4 +1,4 @@
-import { useCallback } from 'react'
+import { useCallback, useMemo } from 'react'
 import { t } from 'decentraland-dapps/dist/modules/translation/utils'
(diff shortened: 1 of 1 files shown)


src/components/InspectorPage/TopBar/TopBar.tsx
```diff
@@ -29,11 +29,24 @@ export default function TopBar({ currentProject, isUploading, onBack, onOpenModa
...
+    url.searchParams.set('realm', `${BUILDER_SERVER_URL}/projects/${currentProject.id}`)
+
+    if (config.is(Env.DEVELOPMENT)) {
+      url.searchParams.set('NETWORK', 'sepolia')
+      url.searchParams.set('DEBUG_SCENE_LOG', '')
+    }
+
+    return url.toString()
+  }, [])
+
```
fzavalia, line 49: I think this way is more readable.

meelrossi (approved):

PR #3038, opened 2024-02-07
fix: Remove iframe from sso

fzavalia:


src/lib/api/builder.ts
```diff
@@ -558,7 +558,7 @@ export class BuilderAPI extends BaseAPI {
         headers = { ...config.headers }
       }
     }
-    const authHeaders = await this.authorization.createAuthHeaders(method, path)
+    const authHeaders = this.authorization.createAuthHeaders(method, path)
```
cyaiox, line 561: is the method `createAuthHeaders` no longer async?

fzavalia: No longer async :D

cyaiox (approved):

PR #3039, opened 2024-02-08
fix: Remove auth flag

fzavalia:


cazala (approved):

src/modules/identity/sagas.ts
```diff
@@ -168,77 +165,33 @@ function* handleLogout(_action: LogoutAction) {
...
-  // If not, proceed with the login in order to retrieve user's signature.
+  const identity = localStorageGetIdentity(address)
   if (identity) {
     yield put(generateIdentitySuccess(address, identity))
     yield put(loginRequest(providerType, true))
   } else {
-    yield put(loginRequest(providerType))
-  }
-
-  yield put(clearAssetPacks())
```
cyaiox, line 235: should we keep this `clearAssetPacks`?

src/modules/land/utils.ts
```diff
@@ -173,7 +173,7 @@ export function locateNextLand(landTiles: Record<string, LandTile>, currentLandI
   const nextIndex = (((index + 1) % landIds.length) + landIds.length) % landIds.length
 
   const nextLandId = landIds[nextIndex]
-  return landTiles[nextLandId]!.land
+  return landTiles[nextLandId].land
```
cyaiox, line 176: could `landTiles[nextLandId]` be undefined?

src/modules/identity/sagas.ts
fzavalia: nice catch 😱

fzavalia: The auth dapp flow returned before doing so, but it might have been a bug from before

src/modules/land/utils.ts
fzavalia: I'm not sure, but the result would be the same with or without the `!`. 
This change was required by the linter.

src/modules/identity/sagas.ts
cyaiox: you're right! with the new auth dapp flow, this is no longer required because when the user is redirected again to the builder, the state is cleared

cyaiox (approved):

src/modules/identity/sagas.ts
fzavalia: But the user might change account via the wallet, which event will end up calling the saga. If the changed wallet already has an identity, it ahould clear the assetpacks and whatever it did before.

cyaiox: good catch! this is the way 👍

PR #3066, opened 2024-03-28
fix: Wert enhancements

fzavalia:


src/modules/collection/sagas.spec.ts
```diff
@@ -2144,99 +2147,42 @@ describe('when publishing a collection with fiat', () => {
...
-                      ]
-                    })
-
-                    describe('when the wert env is dev', () => {
-                      beforeEach(() => {
-                        wertEnv = 'dev'
-                      })
-
-                      // TODO: Unskip once the success action can be reached.
-                      it.skip('should dispatch the action to open the fiat gateway widget with dev parameters', () => {
```
fzavalia, line 2172: Removed skip tests given that they have no reason to be.

cyaiox (approved):

PR #3073, opened 2024-04-05
feat: world permissions

braianj:


src/lib/api/worlds.ts
```diff
@@ -56,6 +98,68 @@ export class WorldsAPI extends BaseAPI {
...
+    }
+  }
+
+  public postPermissionType = async (
+    worldName: string,
+    worldPermissionNames: WorldPermissionNames,
+    worldPermissionType: WorldPermissionType
+  ) => {
+    if (!this.authorization) {
+      return null
```
LautaroPetaccio, line 125: If there's no authorization shouldn't it throw?

src/modules/worlds/reducer.ts
```diff
@@ -59,6 +117,107 @@ export function worldsReducer(state: WorldsState = INITIAL_STATE, action: Worlds
         }
       }
     }
+    case GET_WORLD_PERMISSIONS_SUCCESS: {
```
LautaroPetaccio, line 120: Would you mind adding tests for these actions?

src/modules/worlds/actions.ts
```diff
@@ -15,3 +16,114 @@ export const fetchWorldsWalletStatsFailure = (address: string, error: string) =>
...
+  worldName: string,
+  worldPermissionNames: WorldPermissionNames,
+  worldPermissionType: WorldPermissionType.AllowList,
+  newData: string
+) => action(PUT_WORLD_PERMISSIONS_SUCCESS, { worldName, worldPermissionNames, worldPermissionType, newData })
+export const putWorldPermissionsFailure = (
+  worldName: string,
+  worldPermissionNames: WorldPermissionNames,
+  worldPermissionType: WorldPermissionType.AllowList,
+  newData: string,
```
LautaroPetaccio, line 81: It seems that these parameters are not used in the reducer, would you mind removing them?
```suggestion
```
It seems that there are other actions with the same issue, would you mind checking them as well?

src/modules/worlds/reducer.ts
```diff
@@ -59,6 +117,107 @@ export function worldsReducer(state: WorldsState = INITIAL_STATE, action: Worlds
...
+        }
+      }
+
+      return {
+        ...state,
+        loading: loadingReducer(state.loading, action),
+        worldsPermissions: {
+          ...state.worldsPermissions,
+          [worldName]: {
+            ...state.worldsPermissions[worldName],
```
LautaroPetaccio, line 201: Isn't this already included in the `worldPermissions` variable?

src/modules/worlds/sagas.ts
```diff
@@ -2,46 +2,184 @@ import { call, put, select, takeEvery } from 'redux-saga/effects'
...
+          isErrorWithMessage(e) ? e.message : 'Unknown error'
+        )
+      )
+    }
+  }
+
+  function* handleDeleteWorldPermissionsRequest(action: DeleteWorldPermissionsRequestAction) {
+    const { worldName, worldPermissionNames, worldPermissionType, address } = action.payload
+    try {
+      const worldPremissions: boolean | null = yield call(WorldsAPIContent.deletePermissionType, worldName, worldPermissionNames, address)
```
LautaroPetaccio, line 155: Should this variable be named differently to represent the result of the api call?
```suggestion
      const hasDeletedPermissions: boolean | null = yield call(WorldsAPIContent.deletePermissionType, worldName, worldPermissionNames, address)
```
The same goes for the other sagas here.

src/modules/worlds/sagas.ts
```diff
@@ -2,46 +2,184 @@ import { call, put, select, takeEvery } from 'redux-saga/effects'
...
+
+      if (!worldPremissions) {
```
LautaroPetaccio, line 157: Why isn't considered a failure to not return true?
The same goes for other sagas here.

src/modules/worlds/sagas.ts
```diff
@@ -2,46 +2,184 @@ import { call, put, select, takeEvery } from 'redux-saga/effects'
...
+      return
+    }
+
+    yield put(fetchWorldsWalletStatsRequest(address))
   }
-}
 
-function* handleActionsThatTriggerFetchWorldsWalletStatsRequest() {
-  const address: string | undefined = yield select(getAddress)
+  function* handleGetWorldPermissionsRequest(action: GetWorldPermissionsRequestAction) {
```
LautaroPetaccio, line 75: There are no tests for the new sagas, would you mind adding them?

src/components/Modals/WorldPermissionsModal/WorldPermissionsModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+      const nextPermissionType = !data.checked ? WorldPermissionType.AllowList : WorldPermissionType.Unrestricted
+      if (metadata.worldName && worldPermissions?.access.type !== nextPermissionType) {
+        onPostWorldPermissionsRequest(metadata.worldName, WorldPermissionNames.Access, nextPermissionType)
+      }
+    },
+    [metadata.worldName, worldPermissions]
+  )
+
+  const handleRemoveCollaborators = useCallback(
+    (_event: React.MouseEvent<HTMLButtonElement, MouseEvent>, data: ButtonProps & { wallet: string }) => {
```
LautaroPetaccio, line 144: What do you think about returning if `metadata.worldName` doesn't exist? We would be covering the first two conditions (instead of repeating the check) and the `setCollaborUserList` that shouldn't be triggered if it doesn't exist:
```suggestion
    (_event: React.MouseEvent<HTMLButtonElement, MouseEvent>, data: ButtonProps & { wallet: string }) => {
    if(!metadata.worldName) {
      return
    }
```

src/components/Modals/WorldPermissionsModal/WorldPermissionsModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+      }
+
+      if (
+        metadata.worldName &&
+        worldPermissions?.streaming.type === WorldPermissionType.AllowList &&
+        worldPermissions.streaming.wallets.includes(data.wallet)
+      ) {
+        onDeleteWorldPermissionsRequest(metadata.worldName, WorldPermissionNames.Streaming, WorldPermissionType.AllowList, data.wallet)
+      }
+      setCollaboratorUserList(collaboratorUserList => {
```
LautaroPetaccio, line 160: This is deleting the permission optimistically. It doesn't say anything if there's a failure while doing so. We should re-think the solution so when deleting a permission, the collaborators get updated from the redux state.
The same seems to be happening to all other of the requests.

src/components/Modals/WorldPermissionsModal/WorldPermissionsModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+        onDeleteWorldPermissionsRequest(metadata.worldName, WorldPermissionNames.Deployment, WorldPermissionType.AllowList, data.wallet)
+      }
+
+      if (
+        metadata.worldName &&
+        worldPermissions?.streaming.type === WorldPermissionType.AllowList &&
+        worldPermissions.streaming.wallets.includes(data.wallet)
+      ) {
+        onDeleteWorldPermissionsRequest(metadata.worldName, WorldPermissionNames.Streaming, WorldPermissionType.AllowList, data.wallet)
+      }
```
LautaroPetaccio, line 159: If only one of these conditions should be executed I would consider using an else if, which describes that situation:
```suggestion
      if (
        metadata.worldName &&
        worldPermissions?.deployment.type === WorldPermissionType.AllowList &&
        worldPermissions.deployment.wallets.includes(data.wallet)
      ) {
        onDeleteWorldPermissionsRequest(metadata.worldName, WorldPermissionNames.Deployment, WorldPermissionType.AllowList, data.wallet)
      } else if (
        metadata.worldName &&
        worldPermissions?.streaming.type === WorldPermissionType.AllowList &&
        worldPermissions.streaming.wallets.includes(data.wallet)
      ) {
        onDeleteWorldPermissionsRequest(metadata.worldName, WorldPermissionNames.Streaming, WorldPermissionType.AllowList, data.wallet)
      }
```
Check `handleUserPermissionList` as well.

src/components/Modals/WorldPermissionsModal/WorldPermissionsModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+  }, [worldPermissions?.access.type])
+
+  useEffect(() => {
+    if (
+      tabSelected === TabsSection.Access &&
+      worldPermissions?.access.type === WorldPermissionType.AllowList &&
+      worldPermissions?.access.wallets.length > 0
+    ) {
+      const allowedWallets = new Set(worldPermissions.access.wallets)
+      allowedWallets.has(newAddress.toLowerCase()) && setNewAddress('')
```
LautaroPetaccio, line 53: We usually use `if` conditions when there's the need to execute code, would you mind changing it?
```suggestion
      if (allowedWallets.has(newAddress.toLowerCase())) {
        setNewAddress('')
      }
```

src/components/Modals/WorldPermissionsModal/WorldPermissionsModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+      if (worldPermissions?.streaming.type === WorldPermissionType.AllowList) {
+        worldPermissions.streaming.wallets.forEach(wallet => {
+          !newCollaboratorUserList.has(wallet) && newCollaboratorUserList.add(wallet.toLowerCase())
+        })
+      }
+      return [...newCollaboratorUserList]
+    })
+  }, [worldPermissions])
+
+  const profilesByWallets = useMemo(() => profiles, [profiles])
```
LautaroPetaccio, line 82: This shouldn't be necessary, would you mind removing it?
```suggestion
  const profilesByWallets = useMemo(() => profiles, [profiles])
```

src/lib/api/worlds.ts
braianj: WDYT about `throw new Error('Unauthorized')`

src/modules/worlds/reducer.ts
braianj: nop, the POST changes the type if the permission. For this cases we accept only unrestricted and allow-list, but in the future we could add also ntf and secret.

src/components/Modals/WorldPermissionsModal/ModelTabs/Layouts/WorldPermissionsAddUserForm/WorldPermissionsAddUserForm.tsx
```diff
@@ -0,0 +1,67 @@
...
+      {!showAddUserForm ? (
+        <div className={styles.addUserButtonContainer}>
+          <Button data-testid={WORLD_PERMISSIONS_ADD_USER_FORM_SHOW_FORM_BUTTON_DATA_TEST_ID} onClick={onShowAddUserForm}>
+            <Icon name="plus" />
+            {addButtonLabel}
+          </Button>
+        </div>
+      ) : (
+        <div className={styles.addUserFormContainer}>
+          <Field
```
LautaroPetaccio, line 40: Please, check the `AddressField` component (from the decentraland-dapps repo) that should be more suitable for this task. It does some address checking and it has an auto-name resolver.

src/components/Modals/WorldPermissionsModal/ModelTabs/Layouts/WorldPermissionsAddUserForm/WorldPermissionsAddUserForm.tsx
```diff
@@ -0,0 +1,67 @@
...
+            data-testid={WORLD_PERMISSIONS_ADD_USER_FORM_FIELD_DATA_TEST_ID}
+            placeholder="0x..."
+            value={newAddress}
+            onChange={onNewAddressChange}
+            kind="full"
+            loading={isLoadingNewUser}
+            disabled={isLoadingNewUser}
+            error={error}
+            message={error ? 'invalid address' : ''}
```
LautaroPetaccio, line 49: Any constants should be internationalized, would you mind internationalizing it?

src/components/Modals/WorldPermissionsModal/ModelTabs/Layouts/WorldPermissionsAvatarWithInfo.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <div className={styles.avatar}>
+        <AvatarFace size="small" inline />
+        <LoadingText className={styles.loadingtext} type="p" size="medium"></LoadingText>
+      </div>
+    )
+  }
+
+  return (
+    <div className={styles.avatar}>
+      <AvatarFace className={styles.avatarface} avatar={profiles[wallet]} size="small" inline />
```
LautaroPetaccio, line 24: We have a `Profile` componen in the `decentraland-dapps` repo, which is an enhanced version of the `Profile` component that comes from the UI, which automatically fetches the profiles.

src/components/Modals/WorldPermissionsModal/ModelTabs/Layouts/WorldPermissionsAvatarWithInfo.tsx
```diff
@@ -0,0 +1,33 @@
...
+import CopyToClipboard from 'components/CopyToClipboard/CopyToClipboard'
+import { getResumedAddress } from '../../utils'
+import { WorldPermissionsAvatarWithInfoProps } from './WorldPermissionsAvatarWithInfo.types'
+
+import styles from './WorldPermissionsAvatarWithInfo.module.css'
+
+export default React.memo(function WorldPermissionsAvatarWithInfo(props: WorldPermissionsAvatarWithInfoProps) {
+  const { wallet, profiles, loading } = props
+
+  if (loading || !profiles || !wallet) {
```
LautaroPetaccio, line 13: I think it will be better to make a container for this component. The component should receive only the address of the user and the container should get the profile from the state and if it's loading or not (both selectors are available as the profile module in the `decentraland-dapps` repo)

src/components/Modals/WorldPermissionsModal/ModelTabs/WorldPermissionsAccess/WorldPermissionsAccess.tsx
```diff
@@ -0,0 +1,126 @@
...
+            />
+          </div>
+          <span className={styles.approvedCount}>
+            {worldAccessPermissions?.type === WorldPermissionType.AllowList &&
+              t('world_permissions_modal.tab_access.approved_addresses', { number: `${worldAccessPermissions?.wallets.length}/100` })}
+          </span>
+        </div>
+      )}
+
+      {!isAccessUnrestricted && (
```
LautaroPetaccio, line 79: Should this be loaded if loading?

braianj: yes, but just get the loading prop also

src/components/Modals/WorldPermissionsModal/ModelTabs/Layouts/WorldPermissionsAvatarWithInfo.tsx
braianj: Making 100 fetch at the same time is not the better solution. I'll stick with this one if you don't mind

src/modules/worlds/reducer.ts
```diff
@@ -3,23 +3,52 @@ import {
...
-import { WorldsWalletStats } from 'lib/api/worlds'
+import { AllowListPermissionSetting, WorldPermissionType, WorldPermissions, WorldsWalletStats } from 'lib/api/worlds'
+import { Avatar } from '@dcl/schemas/dist/platform/profile/avatar'
 
 export type WorldsState = {
   // TODO: Find a use for the data object when there is something more relevant as the core data for the worlds module.
   data: Record<string, unknown>
   walletStats: Record<string, WorldsWalletStats>
+  worldsPermissions: Record<string, WorldPermissions>
+  profiles: Record<string, Avatar>
```
LautaroPetaccio, line 42: Would you mind removing the profiles here?

src/modules/worlds/reducer.spec.ts
```diff
@@ -78,3 +106,282 @@ describe('when handling the action to fetch worlds stats for a wallet', () => {
...
+    })
+  })
+
+  describe('when handling the success action to post permission type for a world name', () => {
+    beforeEach(() => {
+      state.loading = [postWorldPermissionsRequest(worldName, WorldPermissionNames.Deployment, WorldPermissionType.AllowList)]
+      state.worldsPermissions[worldName] = worldPermissionsMock
+    })
+
+    it('should remove the action from loading and set the stats', () => {
```
LautaroPetaccio, line 219:
```suggestion
    it('should remove the action from loading and set the new permission', () => {
```

src/modules/worlds/reducer.spec.ts
```diff
@@ -78,3 +106,282 @@ describe('when handling the action to fetch worlds stats for a wallet', () => {
...
+              wallets: []
+            }
+          }
+        }
+      })
+    })
+  })
+})
+
+describe('when handling the action to put an address in deployment permission for a world name', () => {
```
LautaroPetaccio, line 239:
```suggestion
describe('when handling the action to add an address to the deployment permission for a world name', () => {
```

src/modules/worlds/reducer.spec.ts
```diff
@@ -78,3 +106,282 @@ describe('when handling the action to fetch worlds stats for a wallet', () => {
...
+    })
+  })
+
+  describe('when handling the success action to put an address in deployment permission for a world name', () => {
+    beforeEach(() => {
+      state.loading = [putWorldPermissionsRequest(worldName, WorldPermissionNames.Deployment, WorldPermissionType.AllowList, addressWallet)]
+      state.worldsPermissions[worldName] = worldPermissionsMock
+    })
+
+    it('should remove the action from loading and set the stats', () => {
```
LautaroPetaccio, line 289:
```suggestion
    it('should remove the action from loading and set the new address in the wallets for the deployment permission', () => {
```

src/modules/worlds/reducer.spec.ts
```diff
@@ -78,3 +106,282 @@ describe('when handling the action to fetch worlds stats for a wallet', () => {
...
+      state.worldsPermissions[worldName] = {
+        ...worldPermissionsMock,
+        deployment: {
+          type: WorldPermissionType.AllowList,
+          wallets: [addressWallet]
+        }
+      }
+    })
+
+    it('should remove the action from loading and set the stats', () => {
```
LautaroPetaccio, line 369:
```suggestion
    it('should remove the action from loading and remove the wallet address from the wallets list of the given permission', () => {
```

LautaroPetaccio (approved):

PR #3087, opened 2024-04-19
fix: avoid counting models twice for the final size

juanmahidalgo:
Right now, when editing a representation, for the final size calculation, it will add the previous model + the new one, so if the old one has 1.5MB and the new one 1.6M, it will be clearly exceeding the max allowed (currently set at 2MB)


braianj (approved):
👍🏽

PR #3095, opened 2024-05-03
feat: add search for collections

meelrossi:
image


src/components/CollectionsPage/CollectionsPage.tsx
```diff
@@ -31,13 +34,15 @@ import CollectionRow from './CollectionRow'
 import { Props, TABS } from './CollectionsPage.types'
 import './CollectionsPage.css'
 
+let timeout: NodeJS.Timeout
```
LautaroPetaccio, line 37: As the `CollectionsPage` component lives in the web environment, we can ask for the return type of the setTimeout instead of typing it as the `NodeJS.Timeout` one:
```suggestion
let timeout: ReturnType<typeof setTimeout>
```

LautaroPetaccio: Would you mind defining the `timeout` variable as being part of the `CollectionsPage` component instance so it's not globally defined?

LautaroPetaccio (commented):
Great job! 👍 

There seem to be an issue with the search bar and the right buttons. The search bar seems to be stretching too much:
Screenshot 2024-05-03 at 15 05 54
And, I'm not sure if it's related to this change, but when there's no collection available, there's a big gap between the search bar and the nav bar. It seems that as there are no tabs, the space between the components remains:
Screenshot 2024-05-03 at 15 03 54
Would you mind taking a look at it?

LautaroPetaccio (approved):
Looks amaizing, great work 🚀

PR #3099, opened 2024-05-08
fix: builder templates local assets tab

cazala:
This PR fixes an issue where opening the Local Assets on a scene created from a template would crash the Web Editor.

The problem was that the `main.composite` file was being returned twice by the file system interface, and that resulted in the directory tree breaking.

I also took the chance to remove an usused feature flag on this PR: `sdk7-templates`

### DIFF
--- a/src/modules/inspector/sagas.ts
+++ b/src/modules/inspector/sagas.ts
@@ -291,7 +291,12 @@ export function* inspectorSaga(builder: BuilderAPI, store: RootStore) {
     const { path } = params
 
     const scene: SceneSDK7 = yield getScene()
-    const paths = [...Object.keys(scene.mappings), 'assets/scene/main.composite']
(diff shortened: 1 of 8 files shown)


src/modules/inspector/sagas.ts
```diff
@@ -291,7 +291,12 @@ export function* inspectorSaga(builder: BuilderAPI, store: RootStore) {
     const { path } = params
 
     const scene: SceneSDK7 = yield getScene()
-    const paths = [...Object.keys(scene.mappings), 'assets/scene/main.composite']
+    const paths = [...Object.keys(scene.mappings)]
+
+    // add main.composite if missing
+    if (!paths.includes('assets/scene/main.composite')) {
+      paths.push('assets/scene/main.composite')
```
cazala, line 298: this was the actual fix

cyaiox (approved):

PR #3158, opened 2024-08-20
fix: Update models file

cyaiox:
This PR validates the model size, including the thumbnail size, until there's a clear definition of how the catalyst will calculate it.

### DIFF
--- a/src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
+++ b/src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
@@ -67,7 +67,11 @@ import {
   getBodyShapeTypeFromContents,
   isSmart,
   isWearable,
-  buildItemMappings
+  buildItemMappings,
+  isEmoteFileSizeValid,
+  isSkinFileSizeValid,
+  isSmartWearableFileSizeValid,
+  isWearableFileSizeValid
 } from 'modules/item/utils'
 import { EngineType, getItemData, getModelData } from 'lib/getModelData'
 import { getExtension, toMB } from 'lib/file'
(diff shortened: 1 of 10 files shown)


src/modules/item/utils.ts
```diff
@@ -756,3 +762,19 @@ export const buildItemMappings = (mapping: Mapping, contract: LinkedContract): M
     }
   }
 }
+
+export const isWearableFileSizeValid = (fileSize: number): boolean => {
+  return fileSize > MAX_WEARABLE_FILE_SIZE
```
meelrossi, line 767: shouldn't this be `<`? 

Maybe change the name to invalid? I see that you are using it that if it returns true then its invalid

cyaiox: you're right! fixing it

cyaiox: fixed 🙌

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
```diff
@@ -1085,7 +1091,7 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
       return false
     }
 
-    if (isRequirementMet && !isSkin && !isSmartWearable && modelSize && modelSize > MAX_WEARABLE_FILE_SIZE) {
+    if (isRequirementMet && !isSkin && !isSmartWearable && finalSize && !isWearableFileSizeValid(finalSize)) {
```
meelrossi, line 1094: are we missing the !isEmote here?
```suggestion
    if (isRequirementMet && !isSkin && !isSmartWearable && !isEmote && finalSize && !isWearableFileSizeValid(finalSize)) {
```

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
```diff
@@ -1075,7 +1081,7 @@ export default class CreateSingleItemModal extends React.PureComponent<Props, St
       return false
     }
 
-    if (isRequirementMet && !isSkin && isSmartWearable && modelSize && modelSize > MAX_SMART_WEARABLE_FILE_SIZE) {
+    if (isRequirementMet && !isSkin && isSmartWearable && finalSize && !isSmartWearableFileSizeValid(finalSize)) {
```
meelrossi, line 1084: can it be a smart wearable and skin?

meelrossi (approved):

src/components/Modals/CreateSingleItemModal/CreateSingleItemModal.tsx
cyaiox: mmmm, it's a good point. the smart wearables can be skin but have the limitation for the `js` files of 3MB max. 

will update the condition to have a better validation

PR #3162, opened 2024-08-27
feat: Display collections by type

LautaroPetaccio:
This PR changes the collections page to display collections by its type, that is Linked Wearable's collections or Standard collections.

Screenshot 2024-08-27 at 09 51 32


src/components/CollectionsPage/CollectionsPage.tsx
```diff
@@ -87,17 +108,20 @@ export default function CollectionsPage(props: Props) {
...
-  const handleSearchChange = (_evt: React.ChangeEvent<HTMLInputElement>, data: InputOnChangeData) => {
-    setSearch(data.value)
-    if (timeout.current) {
-      clearTimeout(timeout.current)
-      timeout.current = null
-    }
+  const handleSearchChange = useCallback(
+    (_evt: React.ChangeEvent<HTMLInputElement>, data: InputOnChangeData) => {
+      setSearch(data.value)
+      if (timeout.current) {
```
cyaiox, line 114: is this something achievable using `import { debounce } from 'lib/debounce'`?

cyaiox (approved):

src/components/CollectionsPage/CollectionsPage.tsx
LautaroPetaccio: The `debounce` function allows you to create an unique function to be used in each debounce. The problem with this case is that the function will need the latest version of the `changeFilter` function, which changes each time the path changes, thus making it impossible to have an unique debounce function.

PR #3214, opened 2024-10-23
fix: Add FF to fetch collection items from marketplace server

cyaiox:
This PR adds a FF to fetch collection items from marketplace server


juanmahidalgo (approved):
LGTM!

PR #3223, opened 2024-11-18
fix: LW review view not showing wearables to review correctly

LautaroPetaccio:
This PR fixes:
- An issue where the LW review process did not work correctly when switching between the review tag and the all items tag due to the current page being reviewed changed between tab views.
- Having the possibility of seeing the same page when requesting a random page. This was due to not storing the already seen pages in the LeftPanel component
- A mismatching counter between reviewed wearables and the missing ones to complete the review. This was due to having two places where we stored the reviewed items, the state at the Items component and the state at the ItemEditorPage component.
- The random sample confirmation modal not saving that the user didn't want it to be shown again. This was due to the component being rendered each time a page was loaded, loosing the state in the process. To prevent this, the loaders and empty messages at the LeftPanel were moved to the corresponding components.


src/components/ItemEditorPage/LeftPanel/Items/Items.tsx
```diff
@@ -105,50 +81,70 @@ export default class Items extends React.PureComponent<Props, State> {
...
-    const { currentPages } = this.state
+    const { onLoadPage, onResetReviewedItems } = this.props
+    const { currentPages, currentTab } = this.state
+
     if (targetTab === ItemPanelTabs.ALL_ITEMS) {
       onLoadPage(currentPages[ItemPanelTabs.ALL_ITEMS])
     }
+
+    if (targetTab === ItemPanelTabs.ALL_ITEMS) {
+      this.setState({ showAllItemsTabChangeModal: false }, () => onLoadPage(currentPages[ItemPanelTabs.ALL_ITEMS]))
```
cyaiox, line 92: does this line replace the previous condition at line 87?

```
if (targetTab === ItemPanelTabs.ALL_ITEMS) {
  onLoadPage(currentPages[ItemPanelTabs.ALL_ITEMS])
}
```

LautaroPetaccio: It does! I'll remove the older condition.

src/components/ItemEditorPage/LeftPanel/Items/Items.types.ts
```diff
@@ -9,30 +10,32 @@ export enum ItemPanelTabs {
...
 export type Props = {
   items: Item[]
   isLoading: boolean
   totalItems: number | null
   selectedItemId: string | null
-  selectedCollectionId: string | null
+  collection: Collection | null
+  showSamplesModalAgain: boolean
+  onToggleShowSamplesModalAgain: () => unknown
+  onResetReviewedItems: () => unknown
```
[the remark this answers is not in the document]
LautaroPetaccio: Using void as the return type restricts the function to return nothing. The use of unknown here lets the component know that we don't know nor care if they return something.

PR #3232, opened 2024-12-29
fix: Close the sell collection modal when setting the store as minter

LautaroPetaccio:
This PR fixes a couple of small bugs, but the most important is that it fixes the experience of setting the store as minter, which was not closing the modal.


juanmahidalgo (approved):
LGTM!

PR #3233, opened 2025-01-05
fix: Prevent land unpublishing if lessor

LautaroPetaccio:
This PR changes the Scene component to not show the un-publish button on the scenes if the user is a lessor of the shown land.


juanmahidalgo (approved):
LGTM!!

PR #3240, opened 2025-01-21
feat: Add campaign module and banners

LautaroPetaccio:
This PR adds the new campaign module and the code to display the banners and use the campaign data from the module.
It also removes all of the old code that was used to display the banners.
The configurations are only set on dev, as the final entities and keys are to be set before the release.


vite.config.ts
```diff
@@ -62,7 +62,7 @@ export default defineConfig(({ mode }) => {
       commonjsOptions: {
         transformMixedEsModules: true
       },
-      sourcemap: true
+      sourcemap: false
```
LautaroPetaccio, line 65: Removing the sourcemaps is a tradeoff that we need to take at this moment, as building the distributable in vercel is quite impossible.

src/components/CollectionsPage/CollectionsPage.tsx
```diff
@@ -38,6 +38,7 @@ import CollectionRow from './CollectionRow'
 import { Props, TABS } from './CollectionsPage.types'
 import './CollectionsPage.css'
 
+const BUILDER_BANNER_ID = 'builderCampaignBanner'
```
cyaiox, line 41: should we move this const to `modules/common/types`?

cyaiox (approved):

PR #3244, opened 2025-01-27
Feat/paginate ens

juanmahidalgo:


src/components/WorldListPage/WorldListPage.container.ts
```diff
@@ -38,7 +39,8 @@ const mapState = (state: RootState): MapStateProps => ({
   isLoggedIn: isLoggedIn(state),
   worldsWalletStats: getConnectedWalletStats(state),
   isConnected: isConnected(state),
-  isWorldContributorEnabled: getIsWorldContributorEnabled(state)
+  isWorldContributorEnabled: getIsWorldContributorEnabled(state),
+  ensTotal: getTotalENS(state)
```
cyaiox, line 43:
```suggestion
  ensTotal: getExternalNamesForConnectedWallet(state).length
```

src/modules/ens/sagas.ts
```diff
@@ -365,7 +365,10 @@ export function* ensSaga(builderClient: BuilderClient, ensApi: ENSApi, worldsAPI
         REGISTRAR_ADDRESS,
         signer
       )
-      let domains: string[] = yield call([marketplace, 'fetchENSList'], address)
+      const defaultPageSize = 12
+      const { first = defaultPageSize, skip = 0 } = action.payload
+      let domains: string[] = yield call([marketplace, 'fetchENSList'], address, first, skip)
+      const totalDomains: number = yield call([marketplace, 'fetchENSListCount'], address)
       const bannedDomains: string[] = yield call(fetchBannedDomains)
```
LautaroPetaccio, line 372: These 3 requests can be done concurrently to speed up the fetch process, WDYT?

src/components/WorldListPage/WorldListPage.tsx
```diff
@@ -106,22 +118,20 @@ const WorldListPage: React.FC<Props> = props => {
...
-          case SortBy.ASC: {
-            return a.subdomain.toLowerCase() > b.subdomain.toLowerCase() ? 1 : -1
-          }
-          case SortBy.DESC: {
-            return a.subdomain.toLowerCase() < b.subdomain.toLowerCase() ? 1 : -1
-          }
-          default: {
-            return 0
-          }
+    return list.sort((a: ENS, b: ENS) => {
```
LautaroPetaccio, line 121: If we sort the domains here lexicographically, the sorting might break the pagination, as when new domains are fetched, they could be lexicographically before, after or in the same the page the user is seeing. The sorting should be done at the query level for it to work. We can also remove the sorting completely.

src/components/WorldListPage/WorldListPage.tsx
```diff
@@ -89,6 +97,10 @@ const WorldListPage: React.FC<Props> = props => {
     [deploymentsByWorlds, onUnpublishWorld]
   )
 
+  const handlePageChange = useCallback((_event: any, props: any) => {
```
LautaroPetaccio, line 100: If you find it easier, you could use [the `pagination` hook here](https://github.com/decentraland/builder/blob/f04929ed72e867c11ca18a06e9bf2c0837e4da07/src/lib/pagination.ts).

src/modules/ens/sagas.ts
juanmahidalgo: totally! good call!

src/modules/ens/sagas.ts
```diff
@@ -365,7 +365,10 @@ export function* ensSaga(builderClient: BuilderClient, ensApi: ENSApi, worldsAPI
         REGISTRAR_ADDRESS,
         signer
       )
-      let domains: string[] = yield call([marketplace, 'fetchENSList'], address)
+      const defaultPageSize = 12
```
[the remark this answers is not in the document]
juanmahidalgo: we could, yes!

PR #3272, opened 2025-05-20
fix: Map items with missing entities to force a push update

cyaiox:
This PR adds an extra mechanism for published collections that missed item entities, allowing push changes to redeploy items to the catalysts.


src/modules/item/selectors.ts
```diff
@@ -164,18 +172,27 @@ export const getStatusByItemId = createSelector<
...
     for (const item of items) {
+      const isMissingEntity = !!item.urn && missingEntities[item.urn]
+
       statusByItemId[item.id] = isThirdParty(item.urn)
         ? getStatusForTP(item, itemCurationByItemId[item.id])
-        : getStatusForStandard(item, item.collectionId ? curationsByCollectionId[item.collectionId] : null, entitiesByItemId[item.id])
+        : getStatusForStandard(
+            item,
+            item.collectionId ? curationsByCollectionId[item.collectionId] : null,
+            entitiesByItemId[item.id],
```
LautaroPetaccio, line 193: If the item does not have an entity, `entitiesByItemId[item.id]` should be undefined, right? So if the item does not have an entity but is approved, there's an entity missing.

src/modules/entity/reducer.ts
```diff
@@ -56,7 +58,33 @@ export function entityReducer(state: EntityState = INITIAL_STATE, action: Entity
...
+        loading: loadingReducer(state.loading, action),
+        error: null,
+        data: {
+          ...state.data,
+          ...entities.reduce((obj, entity) => {
+            obj[entity.id] = entity
+            return obj
+          }, {} as EntityState['data'])
+        },
+        missingEntities: newMissingEntities
```
LautaroPetaccio, line 85: What if instead of storing the missing entities, we create a new memoized selector, that by using the items and the entities, returns the newMissingEntities map?

src/components/Modals/DeployEntitiesModal/DeployEntitiesModal.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react'
+import { Modal } from 'decentraland-dapps/dist/containers'
+import { Button, ModalActions, ModalNavigation } from 'decentraland-ui'
+import { t } from 'decentraland-dapps/dist/modules/translation/utils'
+import { Props, DeployEntitiesModalView } from './DeployEntitiesModal.types'
+import './DeployEntitiesModal.css'
+
+export default class DeployEntitiesModal extends React.PureComponent<Props> {
```
LautaroPetaccio, line 8: If I'm not wrong, the `ApprovalModal` can be opened in different steps. We might be able to use the `ApprovalFlow` modal directly when clicking the deploy missing entities button, what do you think about it?

src/modules/item/selectors.ts
cyaiox: you're right! When `entitiesByItemId[item.id]` is undefined, the item doesn't have an entity yet.

it could happen that the entities are being fetched or don't exist... 

added an extra validation to know when we're fetching the entities to properly know if it's missing or loading

LautaroPetaccio (approved):

PR #3278, opened 2025-07-02
feat: refresh lands after success transactions

alejandralevy:


src/modules/land/sagas.ts
```diff
@@ -299,3 +312,17 @@ function* handleWallet(action: ConnectWalletSuccessAction | ChangeAccountAction)
 
   yield put(fetchLandsRequest(address))
 }
+
+function* refreshLandsAfterTransaction(from: string) {
+  const dissolveEstateTxn: {
```
cyaiox, line 317: as this method will be called in different kinds of actions, wdyt about renaming this var to something like: `txnStatus`?

cyaiox (approved):
awesome!! 🚀 🚀

PR #3299, opened 2025-11-11
feat: introduce using credits for collection fees

juanmahidalgo:
image

### DIFF
--- a/src/modules/collection/sagas.spec.ts
+++ b/src/modules/collection/sagas.spec.ts
@@ -13,7 +13,9 @@ import { t } from 'decentraland-dapps/dist/modules/translation/utils'
 import { getOpenModals } from 'decentraland-dapps/dist/modules/modal/selectors'
 import { getCredits } from 'decentraland-dapps/dist/modules/credits/selectors'
 import { CreditsResponse } from 'decentraland-dapps/dist/modules/credits/types'
-import { CreditsService } from 'decentraland-dapps/dist/lib/credits'
+import { CreditsService, CollectionManagerCreateCollectionArgs } from 'decentraland-dapps/dist/lib/credits'
+import { ContractName, getContract } from 'decentraland-transactions'
+import { toInitializeItems } from 'modules/item/utils'
 import { config } from 'config'
 import { locations } from 'routing/locations'
 import { ApprovalFlowModalMetadata, ApprovalFlowModalView } from 'components/Modals/ApprovalFlowModal/ApprovalFlowModal.types'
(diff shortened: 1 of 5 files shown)


src/modules/collection/sagas.ts
```diff
@@ -454,7 +457,62 @@ export function* collectionSaga(legacyBuilderClient: BuilderAPI, client: Builder
...
+
+        if (!creditsResponse || !creditsResponse.credits || creditsResponse.credits.length === 0) {
+          throw new Error('No credits available')
+        }
+
+        // Calculate total price
+        const totalPrice = ethers.BigNumber.from(rarity.prices.MANA).mul(items.length).toString()
+
+        // Get the credits server URL from config
+        const creditsServerUrl: string = yield call([config, config.get], 'CREDITS_SERVER_URL')
```
LautaroPetaccio: WDYT about instantiating the `CreditsService` where the builder client or the legacy builder client are instantiated and then pass it when building the sagas?

src/modules/collection/sagas.spec.ts
```diff
@@ -2377,3 +2380,163 @@ describe('when handling a deploy missing entities', () => {
...
+    let useCreditsCollectionManagerSpy: jest.SpyInstance
+
+    beforeEach(() => {
+      ;(mockBuilder.fetchCollectionItems as jest.Mock).mockResolvedValue(items)
+      // Mock fetch globally for credits server signature
+      originalFetch = global.fetch
+      global.fetch = jest.fn().mockResolvedValue({
+        ok: true,
+        json: () => Promise.resolve({ signature: '0xexternalCallSignature', messageHash: '0xhash' } as any)
+      } as any)
```
LautaroPetaccio: If we pass the `creditsService` to the sagas, we shouldn't need to mock the fetch globally, which would simplify our tests.

src/modules/collection/sagas.spec.ts
```diff
@@ -2377,3 +2380,163 @@ describe('when handling a deploy missing entities', () => {
...
+          [put(saveCollectionRequest(collection)), true],
+          [
+            race({ success: take(SAVE_COLLECTION_SUCCESS), failure: take(SAVE_COLLECTION_FAILURE) }),
+            { success: saveCollectionSuccess(collection) }
+          ],
+          [call([mockBuilder, 'fetchCollectionItems'], collection.id), items],
+          [select(getAddress), from],
+          [call(getChainIdByNetwork, Network.MATIC), ChainId.MATIC_AMOY],
+          [retry(10, 500, mockBuilder.saveTOS, TermsOfServiceEvent.PUBLISH_COLLECTION, collection, email), undefined],
+          [put(fetchRaritiesRequest()), undefined],
```
LautaroPetaccio: With the proposed changes, we can mock the `CreditsService.prototype, useCreditsCollectionManager` here and make it return what we want. We can also test that it's called with the correct parameters.

PR #3318, opened 2025-12-11
chore: fix component types for React 18

braianj:
- Add children?: React.ReactNode to Props interfaces
- Use React.PropsWithChildren where needed
- Update tsconfig.json settings

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,8 +1,8 @@
-import * as ReactDOM from 'react-dom'
+import { createRoot } from 'react-dom/client'
 import { Provider } from 'react-redux'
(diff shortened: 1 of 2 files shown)


src/components/SceneStats/Stats/Stats.types.ts
```diff
@@ -1,3 +1,4 @@
 export type Props = {
   label: string
+  children?: React.ReactNode
```
cyaiox: wdyt about using `export type Props = React.PropsWithChildren` in this component?

braianj: IMO both valid but React.ReactNode more explicit

Issue #3241, opened 2025-01-22
ENS list won't load while connected to the Polygon network


Vitsky4079:
### 🛠️ **Browser version:**

132.0.6834.84

### 📔 **Issue Description:**

The ENS list won't load while connected to the Polygon network.

### **STR:** 

1. connect to the Polygon network
2. go to "Your NAMEs"
3. click manage on the NAME

### ➗ **Repro Index:**

10/10

![Image](https://github.com/user-attachments/assets/d9370960-99d5-4695-9854-330bd93ca67b)

PR #3325, opened 2025-12-22
Fix: ENS list won't load while connected to the Polygon network

RocioCM:
# Fix: ENS list won't load while connected to the Polygon network

Closes https://github.com/decentraland/builder/issues/3241

## Problem statement

The ENS list won't load while connected to the Polygon network.

The reason is that we access the contract using a signer (binded to the connected network) instead of a provider explicitly connected to the desired network.

## Solution 

Replaced signer by provider on read-only operations (fetch ENS list & detail). Added an extra validation to show an user-friendly error in case that a write operation is about to be performed on the wrong network.

## Tests

- [ ] Go to builder then go to "Names" tab, sign in and connect Metamask to any network different from Sepolia (choose Ethereum Mainnet, Polygon, Base or any other) -> A modal should appear automatically prompting to connect to Sepolia.
- [ ] Go to builder then go to "Names" tab, sign in and connect Metamask to Amoy (Polygon testnet) -> Modal should not be shown or should dissapear automatically. Names list should be rendered successfully without any errors on UI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,38 +1,143 @@
 <img src="https://ui.decentraland.org/decentraland_256x256.png" height="128" width="128" />
 
-# Decentraland Builder [![Coverage Status](https://coveralls.io/repos/github/decentraland/builder/badge.svg?branch=chore/add-tests-and-ci)](https://coveralls.io/github/decentraland/builder?branch=chore/add-tests-and-ci)
+# Decentraland Builder
 
-You can create beautiful scenes for [Decentraland](https://decentraland.org) even if you don't own a parcel.
+[![Coverage Status](https://coveralls.io/repos/github/decentraland/builder/badge.svg?branch=master)](https://coveralls.io/github/decentraland/builder?branch=master)
+
+This UI allows users to create beautiful scenes for [Decentraland](https://decentraland.org), manage wearables and emotes, handle LAND parcels and Estates, and deploy content to Worlds.
 
 ![](https://github.com/decentraland/builder/blob/master/public/images/intro.gif)
 
-# How to run
+## Table of Contents
 
-The builder is a SPA or single page application built with [create-react-app](https://github.com/facebook/create-react-app). It uses an [`.env`](#environment) file as configuration for a few things, you'll need to create that first.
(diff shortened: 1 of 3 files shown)


cyaiox (approved):

src/modules/ens/sagas.ts
```diff
@@ -72,6 +78,28 @@ import { addWorldStatusToEachENS, getLandRedirectionHashes, isExternalName } fro
...
+  function* validateAndSwitchNetwork() {
+    const ethereumChainId: number = yield call(getChainIdByNetwork, Network.ETHEREUM)
+    const wallet: Wallet = yield getWallet()
+    if (wallet.chainId !== ethereumChainId) {
+      const signer: ethers.Signer = yield call(getSigner)
+      const signerNetwork: ethers.providers.Network = yield call([signer.provider as ethers.providers.Provider, 'getNetwork'])
+      if (signerNetwork.chainId !== ethereumChainId) {
+        yield put(switchNetworkRequest(ethereumChainId, signerNetwork.chainId))
+        yield take(SWITCH_NETWORK_SUCCESS)
+      }
```
cyaiox: I think the `wallet.chainId !== ethereumChainId` validation works as the `switchNetworkRequest` updates the wallet state.

so the `wallet.chainId` and `signerNetwork.chainId` should be equal

src/modules/ens/sagas.ts
```diff
@@ -397,7 +429,7 @@ export function* ensSaga(builderClient: BuilderClient, ensApi: ENSApi, worldsAPI
           const resolver = resolverAddress.toString()
 
           try {
-            const resolverContract = ENSResolver__factory.connect(ENS_RESOLVER_ADDRESS, signer)
+            const resolverContract = ENSResolver__factory.connect(ENS_RESOLVER_ADDRESS, provider)
```
cyaiox:
```suggestion
            const resolverContract = ENSResolver__factory.connect(ENS_RESOLVER_ADDRESS, ethereumProvider)
```

cyaiox (approved):

PR #3366, opened 2026-03-17
fix: Improve UX when minting more than 50 NFTs in a single transaction

### LINKED ISSUE #3365: Improve UX when minting more than 50 NFTs in a single transaction
## Problem

Currently, when a user tries to mint more than 50 NFTs in a single transaction, the `MAX_NFTS_PER_MINT` limit (defined in `src/modules/item/utils.ts`) silently blocks the action without showing a clear alert or warning. This makes it unclear for the user what's happening and why their mint didn't go through.

**Relevant code:**
- `src/modules/item/utils.ts:65` — `MAX_NFTS_PER_MINT = 50`
- `src/components/Modals/MintItemsModal/MintItemsModal.tsx:60-62` — limit check sets an error state, but lack of upfront visibility causes confusion

### Option 1: Show an alert/warning
Display a visible warning to the user **before** they attempt to submit, informing them of the 50 NFT per-transaction limit. This could be:
- A banner/notice in the `MintItemsModal` indicating the max per transaction
- Real-time validation as the user adds mint entries, disabling the submit button and showing a message when the limit is exceeded

### Option 2: Evaluate removing the limitation due to Polygon gas price
Since Polygon gas costs are significantly cheaper, evaluate whether the 50 NFT per-transaction cap is still necessary and if removing or increasing this limit is feasible.

JiwaniZakir:
Fixes #3365

Previously, users could configure mints exceeding `MAX_NFTS_PER_MINT` in `MintItemsModal` and attempt to submit the transaction, only receiving an error after the fact. The modal had no pre-submission validation against the per-transaction mint limit. `MintItemsModal.tsx` now computes `exceedsLimit` by comparing `totalMints` against `MAX_NFTS_PER_MINT` after the reduce, disables the submit button via `isDisabled`, and surfaces a localized warning via `t('mint_items_modal.limit_reached')` in the existing error row before the user attempts submission. Verified by manually entering mint amounts exceeding 50 and confirming the button disables and the error message renders inline.

### DIFF
--- a/src/components/Modals/MintItemsModal/MintItemsModal.tsx
+++ b/src/components/Modals/MintItemsModal/MintItemsModal.tsx
@@ -100,8 +100,6 @@ export default class MintItemsModal extends React.PureComponent<Props, State> {
 
     const isEmpty = items.length === 0
     const isFull = items.length === totalCollectionItems
-    const isDisabled = this.isDisabled()
-
     const totalMints = Object.values(itemMints).reduce((accumulator, mints) => {
       const totalMintsPerItem = mints.reduce((acc, mint) => {
         const amount = mint.amount || 0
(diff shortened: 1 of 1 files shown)


RocioCM:
Closing as the issue is already solved by https://github.com/decentraland/builder/pull/3367

JiwaniZakir:
The inline validation approach is solid — catching the limit before submission is much better UX than surfacing a transaction error after the fact. One thing worth confirming: does `exceedsLimit` also handle the edge case where individual item quantities are modified after an initial valid total, or does it only recompute on explicit user input changes? Also, since @cyaiox noted this is fixed in #3367, it may be worth checking whether that PR's implementation handles the same validation path before closing this one out.

PR #3404, opened 2026-05-12
feat: facial expressions support for image-based wearables

LautaroPetaccio:
## Why

For wearables in the image-only categories (`EYES`, `EYEBROWS`, `MOUTH`), creators currently ship one base PNG and (optionally) a same-named `_mask.png`. We want to let them ship an additional pair — `*_expressions.png` and `*_expressions_mask.png` — so the avatar can render facial expressions on top of those wearables. The base PNG is still the only required file; the rest are auxiliary.

The work also surfaces, both during upload and in the wearable editor, a small badge that tells creators (and reviewers) that a wearable carries facial-expression assets. Detection is derived from `item.contents`, so existing wearables get the indicator retroactively without a schema or backend change.

While threading this through, we also fix two pre-existing upload bugs that combined to make subfolder-organized image wearables and GLB-with-external-texture uploads silently broken. Details below.

### 1. File-name conventions and helpers (`src/modules/item/utils.ts`)

A small set of pure helpers identifies the four file shapes:

- `isMaskFile(name)` — matches `*_mask.png`, **excluding** `*_expressions_mask.png`
- `isExpressionsFile(name)` — matches `*_expressions.png` (which by construction does not end in `_mask.png`, so no extra guard needed)
- `isExpressionsMaskFile(name)` — matches `*_expressions_mask.png`
- `getRequiredCounterpartFile(name)` — for any auxiliary file, returns the path of the sibling that must exist for it to be valid (`mask → base`, `expressions → base`, `expressions_mask → expressions`); returns `null` for the base PNG
- `hasFacialExpressions(contents)` — true when any key matches `isExpressionsFile`. Used to drive the badge
- `findOrphanedAuxiliaryFiles(contents)` — returns the full list of `{ orphan, expected }` pairs. Counterpart lookup is case-insensitive so a base `Eyes.PNG` still satisfies an `Eyes_mask.png`

`isModelPath` is updated so all three auxiliary shapes are excluded from the main-file detector. Combined with `findOrphanedAuxiliaryFiles`, this means a zip with a stray `_mask.png` and no base will produce a specific, actionable error instead of a vague "couldn't find a model" message.

### 2. Validation hook on upload (`ImportStep.tsx`)

In `handleZippedModelFiles`, immediately after the zip is unpacked and a `modelPath` has been picked, we run the orphan check when the main file is a PNG:

```ts
if (isImageFile(modelPath)) {
  const orphans = findOrphanedAuxiliaryFiles(content)
  if (orphans.length > 0) {
    throw new OrphanedAuxiliaryFileError(orphans)
  }
}
```

`OrphanedAuxiliaryFileError` (`src/modules/item/errors.tsx`) extends `CustomErrorWithTitle`, renders a bullet list of every offending file in one shot (same pattern as `GLBValidationError`), and is mapped through the existing `handleErrorsOnFile` switch. Translations live in `en.json`, `es.json`, `zh.json`.

The validation runs regardless of whether the zip ships a `wearable.json`. The user spec wants this rule applied universally; the cost is that a creator who used `_mask` or `_expressions` as a non-mask naming convention will now get a clear error and have to rename. Trade-off discussed and accepted.

### 3. Auto-promote to BOTH for image wearables uploaded flat

When the zip has no `wearable.json` and no `male/`/`female/` subfolders, the legacy code left `bodyShape = null` and forced the creator to pick. For image wearables specifically, the user wants both representations built from the same files. So:

```ts
const shouldForceBoth = isImageFile(model) && !isRepresentation && !changeItemFile
acceptedFileProps.bodyShape = (detectedBodyShape ?? (shouldForceBoth ? BodyShapeType.BOTH : null)) as BodyShapeType
```

The `!isRepresentation && !changeItemFile` guard is important: those flows must preserve their original single-body-shape semantics. (Without the guard, the "add representation" flow would have thrown `InvalidModelFilesRepresentation` for image wearables — a regression caught during review.)

### 4. Visual indicator (`FacialExpressionsBadge`)

New stateless component that renders a `decentraland-ui` `Popup` wrapping a `smile outline` icon, with the tooltip text driven by translation key `facial_expressions_badge.tooltip`. Returns `null` when `hasFacialExpressions(contents)` is false, so the badge is purely opt-in based on the file shape.

Mounted in two places:

- **Upload modal — `WearableDetails`**: absolute-positioned at the top-right corner of the thumbnail, so the creator sees confirmation that their expression files were detected.
- **Wearable editor — `RightPanel`**: in the Properties header next to the title, only when the selected item is a wearable. Auto-detects from `item.contents`, so previously uploaded wearables get the badge for free as soon as a creator re-uploads with expression files.

### 5. Side-quest: two pre-existing upload bugs that needed fixing alongside

`cleanContentModelKeys` (`ImportStep.tsx`) and `buildRepresentationsZipBothBodyshape` (`CreateSingleItemModal/utils.ts`) combined to silently break two upload paths:

- `cleanContentModelKeys` unconditionally dropped any `*.png` inside a `male/`/`female/` subfolder. The function's stated intent (per its PR title back in 2022) was to *extract* images and models to the root level, but for images the implementation was a `return newContents` with nothing added. Image wearables organized in body-shape folders ended up with empty contents; GLB wearables with externally-referenced textures next to the model ended up with missing-texture renders.
- `buildRepresentationsZipBothBodyshape` selected `mainFile` via `content.includes('glb')`. For image wearables that returns `undefined`, masked by a `!` non-null assertion. Even if the previous bug were fixed, the saved representation would have no valid main file.

Both are needed to make image wearables in subfolders work end-to-end, so they're fixed in this PR:

- `cleanContentModelKeys` now treats image files the same way it treats model files: strip the prefix for single body shape, keep the prefix for BOTH. Empty entries are still dropped, non-image/non-model files are still passed through unchanged.
- `buildRepresentationsZipBothBodyshape` now picks `mainFile` via `isModelPath`, which returns the GLB if there is one and otherwise the base PNG (auto-excluding masks, expressions, and `thumbnail.png`).

Net effect: flat image-wearable uploads (the primary user case) work; `male/` `female/` subfolder image wearables work (was broken); GLB wearables with external textures in body-shape folders now render their textures correctly (was silently broken).

## Test plan

- [ ] `npm test` — 240 tests pass in the affected suites (`modules/item`, `FacialExpressionsBadge`, `CreateSingleItemModal`). New tests cover every helper, the orphan validator (case-insensitive matching + multiple orphans), and the `buildRepresentationsZipBothBodyshape` GLB and image paths.
- [ ] Run the builder locally and upload a zip containing `red_eyes.png` + `red_eyes_mask.png` + `red_eyes_expressions.png` + `red_eyes_expressions_mask.png` (no wearable.json). Verify category auto-detects to `Eyes`, both representations are built with all four files in each `contents`, and the facial-expressions badge appears on the thumbnail in the configuration step.
- [ ] Save the wearable, open it in the item editor, verify the badge appears next to the Properties title in the right panel.
- [ ] Negative: zip with only `red_eyes_mask.png` (no base) → upload should fail with the orphaned-auxiliary error pointing at the missing base PNG.
- [ ] Negative: zip with `red_eyes.png` + `red_eyes_expressions_mask.png` (no `red_eyes_expressions.png`) → same error class, pointing at the missing expressions PNG.
- [ ] Regression: existing image-wearable uploads (single PNG, no auxiliaries) still work and the badge does **not** appear.
- [ ] Regression: GLB wearable upload with embedded textures still works end-to-end.
- [ ] Regression: "Add representation" flow for an existing image wearable still works (does not trip `InvalidModelFilesRepresentation`).


LautaroPetaccio:
Thanks for the review! Addressed all four points:

**1. Double `.toLowerCase()` in `isModelPath`** — fixed in [b33e6883](https://github.com/decentraland/builder/pull/3404/commits/b33e6883). Inlined the suffix checks (`endsWith('_mask.png')` / `endsWith('_expressions.png')`) so the filename gets lowercased exactly once. The is\*File helpers stay self-sufficient for their own callers.

**2. Non-null assertion in `buildRepresentationsZipBothBodyshape`** — also in [b33e6883](https://github.com/decentraland/builder/pull/3404/commits/b33e6883). Replaced the `!` with an explicit `MissingModelFileError` throw when no main file is found in a per-body-shape bucket. Upstream orphan validation should prevent this path, but now the failure mode is a clear error rather than an undefined `mainFile` propagating to the saved representation. Covered by a new test in `CreateSingleItemModal/utils.spec.ts`.

**3. `OrphanedAuxiliaryFileError` coverage** — added focused tests in `modules/item/errors.spec.tsx` for single-orphan, multi-orphan, and empty-list construction, verifying the rendered `<ul>` has one `<li>` per orphan and that the class still extends `CustomErrorWithTitle`.

**4. Cross-repo `isModelPath` in builder-client** — opened [decentraland/builder-client#81](https://github.com/decentraland/builder-client/pull/81) to align the auxiliary-detection rule there. The builder doesn't currently consume `loadedFile.mainModel`, so this PR isn't blocked by it, but the two codebases now share the same definition of "auxiliary PNG".

RocioCM (approved):

nicoecheza (approved):

PR #3412, opened 2026-05-28
feat: open desktop explorer via deep-link in see-in-world modal

braianj:
When users click Jump In in the See-in-World modal, the builder now tries to launch the desktop client via
  `decentraland://?self-preview-builder-collections=<id>&dclenv=<env>&position=x,y` instead of opening /play in a new tab. Detection uses focus/visibility
  loss within 750ms (consistent with the ui2 `launchDesktopApp` pattern). When the deep link does not resolve, a `DownloadModal` from
  `decentraland-ui2` opens linking to decentraland.org/download. The legacy `getExplorerURL` is untouched so consumers that still need the /play URL keep
   working.

  ## Test plan
  - [ ] With the desktop explorer installed → desktop opens with the unreleased collection injected.
  - [ ] Without it installed → DownloadModal opens; Download button goes to decentraland.org/download.
  - [ ] Item-only invocations (no collectionId) close the modal cleanly without firing the deep link.

### DIFF
--- a/src/components/Modals/SeeInWorldModal/SeeInWorldModal.tsx
+++ b/src/components/Modals/SeeInWorldModal/SeeInWorldModal.tsx
@@ -1,21 +1,55 @@
 import React from 'react'
 import classNames from 'classnames'
 import { Button, Header, ModalContent, ModalNavigation } from 'decentraland-ui'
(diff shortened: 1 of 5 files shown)


juanmahidalgo (approved):
LGTM

PR #3454, opened 2026-07-30
feat: pose wearables for auto-generated thumbnails by category

RocioCM:
## Problem

Auto-generated thumbnails capture wearables in their authored bind pose (T-pose) from a 45° camera, which hides the details of upper body wearables and skins.

## Solution

Capture categories that benefit from it in a **static pose, from the front camera**, using the same offscreen renderer mechanism that already gives skin-like categories their front-facing camera (`getThumbnailType`). No wearable-preview involvement: the offscreen path is synchronous and deterministic.

| Before | After |
|---|---|
| Bind pose (arms out), 45° camera | Natural pose, front camera |

### How it works

- **`src/poses/pose_nicer_idle.glb`** — a mesh-less GLB with a single-keyframe animation on the standard `Avatar_*` armature (new top-level `poses/` folder, aliased like `icons/` and `images/`).
- **`lib/getScreenshot.ts`** — new optional `pose` option: loads the pose GLB as an asset container and writes each keyframe value directly onto the model's matching transform nodes (suffix-stripped name matching), renders once so bone matrices update, then screenshots. No animation system involved.
- **`lib/getModelData.ts`** — `getItemData` renders the import-time thumbnail of posed categories with the offscreen engine instead of screenshotting the wearable-preview iframe.
- **`CreateSingleItemModal/utils.ts`** — `THUMBNAIL_POSE_BY_CATEGORY` maps `upper_body` and `skin` to the pose; `upper_body` joins the `FRONT` camera bucket; shared `getThumbnailRenderOptions` helper deduplicates the render options (and removes a legacy `as any`).
- **`CommonFields.tsx`** — changing the category in the details step regenerates the thumbnail when the camera angle **or pose** differs, exactly like the existing skin behavior.

Adding a pose for another category = one GLB in `src/poses/` + one `THUMBNAIL_POSE_BY_CATEGORY` entry.

## Testing

- Pose retargeting validated against a replica of the offscreen pipeline (Babylon 4.2.2, same ortho camera/lights/scaling): all 186 pose channels match the standard armature; posed + front camera renders the garment dead-on.
- Manually verified in the Create Item modal: upper body and skin GLBs get posed front-facing thumbnails at import and when switching the category in the details step; other categories are unchanged.

## Screenshots

https://github.com/user-attachments/assets/e7684aed-9b0c-4b21-a972-7d6335b8302b


src/lib/getScreenshot.ts
```diff
@@ -35,11 +35,49 @@ async function refreshBoundingInfo(parent: Mesh) {
...
 
+const POSE_PROPERTIES = new Set(['position', 'rotationQuaternion', 'rotation', 'scaling'])
+
+/**
+ * Applies a static pose to the loaded model's skeleton. The pose file is a mesh-less GLB with a
+ * single-keyframe animation targeting the standard Avatar_* armature: instead of running the
+ * animation system, the keyframe values are written directly into the matching transform nodes.
+ */
+async function applyPose(scene: import('@babylonjs/core').Scene, poseUrl: string) {
+  const { SceneLoader } = await import('@babylonjs/core')
```
nicoecheza, line 46: is this dinacamically load intended?

src/lib/getModelData.ts
```diff
@@ -434,7 +436,21 @@ export async function getItemData({
...
+    // apply the pose deterministically) instead of screenshotting the wearable preview iframe
+    const renderOptions =
+      type === ItemType.WEARABLE && category ? getThumbnailRenderOptions(category as WearableCategory, model) : undefined
+    if (renderOptions?.pose) {
+      const modelUrl = URL.createObjectURL(contents[model])
+      try {
+        data.image = await getScreenshot(modelUrl, renderOptions)
+      } finally {
+        URL.revokeObjectURL(modelUrl)
+      }
```
nicoecheza, line 453: there is no `catch` here. should we add one?

src/lib/getModelData.ts
```diff
@@ -434,7 +436,21 @@ export async function getItemData({
...
+    } else {
+      data.image = await wearablePreviewController.scene.getScreenshot(THUMBNAIL_WIDTH, THUMBNAIL_HEIGHT)
+    }
```
nicoecheza, line 456: and no try/catch here

nicoecheza (changes requested):

src/lib/getScreenshot.ts
RocioCM: Yes, babylon.js is always loaded dynamically in this file, so the package (around 4MB) only loads when a thumbnail with a pose is actually rendered. There's no runtime cost, applyPose is only called from getScreenshot, which has already awaited the same import, so the second call resolves from the cache

src/lib/getModelData.ts
RocioCM: This `wearablePreviewController.scene.getScreenshot` was pre-existing without the try-catch, this line was just indented on this PR. We added a try-catch to `getScreenshot` as it throws errors intentionally, but  `wearablePreviewController.scene.getScreenshot` is not intended to throw errors and in case it does, those are just propagated to the caller, like the error explicitly thrown upper on the same function `getItemData`.

src/lib/getScreenshot.ts
```diff
@@ -35,11 +35,49 @@ async function refreshBoundingInfo(parent: Mesh) {
...
 const hideMaterialList = ['hair_mat', 'avatarskin_mat']
 
+const POSE_PROPERTIES = new Set(['position', 'rotationQuaternion', 'rotation', 'scaling'])
+
+/**
+ * Applies a static pose to the loaded model's skeleton. The pose file is a mesh-less GLB with a
+ * single-keyframe animation targeting the standard Avatar_* armature: instead of running the
+ * animation system, the keyframe values are written directly into the matching transform nodes.
+ */
+async function applyPose(scene: import('@babylonjs/core').Scene, poseUrl: string) {
```
cyaiox, line 45: As we load Babylon in the builder to generate the screenshot, can we use the wearable-preview in Unity to take the screenshot and handle the poses? That way, we could remove Babylon from the builder and centralize the screenshot logic

RocioCM: We could migrate wearable/emotes thumbnail generation to unity wearable-preview, but it's not already done as unity doesn't support yet some of the props/functionality required to render the thumbnails correctly. It's possible, but it's a bigger refactor that requires changes in unity too, I could ship it as a follow-up PR, wdyt?

nicoecheza (approved):

src/lib/getScreenshot.ts
cyaiox: makes sense, go ahead!

cyaiox (approved):

PR #3460, opened 2026-08-05
feat: show a credits listing's price and link it to the shop

juanmahidalgo:
A creator whose item is on sale in the shop sees the wrong price on their own collection page — and not
slightly wrong.

`modules/item/sagas.ts` overwrites the builder item's `price` with the marketplace's `/v1/items` price
(`price: publishedItem?.price || item.price`) and keeps the `tradeId`. When that listing is USD-pegged, the
amount is **USD wei** (1e18 = $1), not MANA wei — but `CollectionItem` rendered
`ethers.utils.formatEther(item.price)` behind the MANA glyph regardless. A $0.60 listing, which the shop
charges as 6 credits, reads as **"◈ 0.6"**.

The builder only ever signs `TradeAssetType.ERC20` trades itself, so a pegged trade on a builder item always
came from the shop — which is exactly the flow creators are being moved onto.

## How a pegged item is recognised

`received[0].assetType` on the trade is the only client-side discriminator: the price field carries no unit,
and neither `Item` nor `Order` in `@dcl/schemas` has a denomination flag. `modules/trade/denomination.ts` reads
it per `tradeId`, cached (a trade is signature-pinned, so immutable), **failures not cached** and falling back
to MANA — every listing this app creates is MANA, so MANA is the safe unknown, and a transient error must not
pin the wrong unit for the session.

## The peg

`lib/credits.ts` mirrors the shop's rules deliberately, because a price that disagrees between the two apps is
the bug this closes: `USD_CENTS_PER_CREDIT = 10`, `USD_WEI_PER_CREDIT` derived from the peg rather than
restated, BigInt **ceil** so the shown price never sits below what checkout charges, floored at 1 credit,
`null` on a malformed amount (`BigInt('')` is `0n`, which would otherwise show a confident "1 credit" for no
price at all).

Duplicated rather than shared — the builder and the shop do not depend on each other, and adding that
dependency to render a price would be worse than 80 lines. The credits glyph is the shop's own SVG recoloured
to `currentColor` and applied as a CSS mask, since there is no svgr here and `currentColor` does not cross an
`` boundary.

## Test plan

- [x] `npm run lint`, `npm run check:code` — clean
- [x] `npm run prebuild && npm run build:website`
- [x] `npm test` — 118 suites / 1528 passing
- [x] `CollectionItem`'s three new assertions all fail without the change; an `assetType: 1` fixture keeps
      rendering MANA, and an item with no `tradeId` never reads a trade
- [x] `lib/credits.spec.ts` pins the peg, the ceil, the 1-credit floor, BigInt exactness past float range, and
      the production case `600000000000000000 → 6`
- [ ] A creator's pegged item checked by eye on `.zone`

## Note

`check:prettier` flags `src/lib/thumbnailOptions.spec.ts`, which is **already unformatted on `master`** (its
diff against `HEAD` is empty). Left alone rather than shipped as unrelated churn.


juanmahidalgo:
Pushed a second commit: the price now says **where** the item is on sale, not just how much.

## Why this belongs with the price fix

They are one statement to a creator. Before this, a pegged item showed a number in credits and a plain
"Published" — the price was here and the place it applies was nowhere. A credits listing lives in the **shop**,
a different site from the marketplace this app publishes to, and until now nothing in the builder could even
name it: there is not one reference to the shop anywhere in `src/`.

So when an item's active listing is USD-pegged, the row's published status becomes a link to the item's shop
page. It **replaces** the generic "Published" rather than adding a second badge — a pegged listing implies the
item is published, and "on sale in the shop" is the more specific true thing.

## What was added

- `SHOP_WEB_URL` in `config/env/{dev,stg,prod}.json`, next to `MARKETPLACE_WEB_URL` so the two sibling sites
  read together. Per-environment on purpose: a creator on `.zone` must never be handed a production link to
  their own item.
- `lib/shop.ts` — `shopItemUrl(contractAddress, itemId)`, keyed by the (collection, item) pair the shop's own
  route takes rather than by the trade, so the link stays valid when the listing is replaced. Returns `''` when
  `SHOP_WEB_URL` is unset, so a missing config hides the link instead of pointing at nowhere.
- The link uses `item.tokenId` — this app's name for the item's blockchain id, the same value the published-item
  merge keys on (`${contractAddress}-${item.tokenId}`) and the trade builder sends as `itemId`.
- `target="_blank"` with `rel="noopener noreferrer"`, and `stopPropagation` so opening the listing does not also
  trigger the row's own navigation.

## Tests

118 suites / 1531 passing. Three new cases: a pegged item links to `/shop/item/<contract>/<tokenId>`; a MANA
item keeps the plain published status and no link; and an item with no blockchain id yet renders no link — that
last one waits for the credits price to settle first, so the absence is a real absence rather than the link
merely not having rendered yet.

## Still to come, separately

The put-for-sale modal is where this gets genuinely useful: a **MANA / credits choice** explaining that one
sells on the marketplace and the other in the shop. I checked the cost — `createItemOrderTrade` already signs a
generic `TradeCreation`, and the pegged variant is two fields in `received` (`assetType: USD_PEGGED_MANA` and
the amount in USD wei), with `beneficiary` already pointing at the creator, which is correct for a primary
sale. No signing machinery needs duplicating. It changes what gets EIP-712-signed though, so it goes in its own
reviewed PR rather than riding along here.

Worth flagging while that is pending: today the builder can create a MANA order on an item **already selling
for credits in the shop**, leaving two live prices for the same item and letting a buyer take the cheaper. That
is this morning's incident — an uncancelled MANA order alongside a pegged one — except by design instead of by
accident. The warning at the moment of listing is the part that protects money; this link is the friendly half.

PR #3473, opened 2026-08-20
feat: list collections through the newest off-chain marketplace

LautaroPetaccio:
> **Stacked** on #3472 (the dependency bump) — review that first; this PR's diff is only the listing changes.
> **Blocked** on a `decentraland-dapps` release containing decentraland/decentraland-dapps#819. Draft for that reason. The `decentraland-transactions` 3.1.1 dependency is resolved — 3.1.1 is released and #3472 pins it.

## What

Builder grants minter rights to the off-chain marketplace so it can mint the listed item. That was pinned to V2; it now resolves the newest version deployed on the chain, so testnet listings go through V3 while mainnet keeps using V2 until V3 ships there.

- `enableSaleOffchain` grants the resolved version
- `PutForSaleOffchainModal` labels and authorizes the resolved version instead of hardcoding V2
- `isEnableForSaleOffchain` accepts **any** deployed version as the minter

## Why the check has to widen, not switch

The minter check is what makes a collection show as on sale. If it only accepted the newest version, every collection already listed through V1 or V2 would immediately look unlisted — which is exactly what the V1 → V2 rollout had to handle too (`37531cbb` widened it from V1 to "V1 or V2"). This widens it to "any deployed version", so the next version needs no change here.

## The ordering constraint that makes this blocked

The marketplace granted minter rights is the one that actually mints, so it has to be the same version the trade was signed against — otherwise the mint reverts. Signing happens in `decentraland-dapps`' `getTradeSignature`, which decentraland/decentraland-dapps#819 moves to the same "newest deployed version" resolution.

So this must not merge before a dapps release carrying that change. If Builder granted V3 rights while the shipped dapps still signed V2 trades, listings would break.

The resolution is duplicated here rather than imported from dapps to keep this PR verifiable on its own, and the code says so — the two definitions have to stay in step. Happy to collapse it onto a dapps export as a follow-up once #819 is released.

## Testing

`src/modules/collection/utils.spec.ts` passes with 36 tests, up from 31. New cases cover resolving to V3 on Amoy, reporting V3 as the modal's contract label, falling back to V2 on Polygon mainnet, and a collection minted by V2 still reading as on sale while V3 is current. `tsc` and `eslint` are clean.

I ran every spec that touches the changed helpers, not just this one — `utils.spec.ts` is the only one.


LautaroPetaccio (commented):
Good catch — that was a real bug and it was mine. Fixed in f413bbe5.

Widening `isEnableForSaleOffchain()` to accept any deployed version without widening the revoke left the toggle half-broken: a collection holding V2 minter rights kept them after disabling, so existing V2 listings stayed sellable, and `isEnableForSaleOffchain()` still reported the collection as on sale — so the toggle looked like it had done nothing.

Worth noting the original V1 → V2 rollout got this right: `37531cbb` revoked whichever version the collection actually held, and `d2da9a66` later trimmed it to the single-address one-liner I inherited.

Now:

```ts
// Enabling grants the newest version only: the minter has to be the same contract as the
// signature domain or the mint reverts.
if (isOnSale) {
  return [{ address: getLatestOffchainSale(chainId).address, hasAccess: true, collection }]
}

const authorized = getOffchainSaleAddresses(chainId).filter(address => includes(collection.minters, address))
const toRevoke = authorized.length ? authorized : [getLatestOffchainSale(chainId).address]

return toRevoke.map(address => ({ address, hasAccess: false, collection }))
```

The fallback is deliberate: `handleSetCollectionMintersRequest` turns this list into the `addresses`/`values` arrays for `setMinters`, so returning `[]` when a collection holds no off-chain minters would send the contract two empty arrays. Falling back to the newest keeps the previous single-entry shape for that edge case.

Two tests added — that both held versions are revoked, and that a version the collection does not hold is not. 38 passing, `tsc` and `eslint` clean.

PR #3479, opened 2026-08-24
fix: drop the local navbar height compensation

braianj:
Takes `decentraland-dapps` to `^29.6.2` and removes the local margin this repo was carrying to work around a library bug that is now fixed.

### What this removes

`AnnouncementBar.module.css` had:

```css
/* The ui2 navbar is fixed and 92px tall on desktop, but the container that
   wraps it only reserves 66px, so the first 26px of the flow sit behind it.
   Compensating here keeps the bar flush under the navbar instead of below it.
   Under 992px the navbar is 64px tall and already fits in those 66px. */
margin-top: 26px;
```

The comment described the bug exactly, and it is fixed at the source: decentraland/decentraland-dapps#822, shipped in 29.6.2, makes `Navbar2`'s container track the bar's real heights (64px below 992px, 92px above) instead of a flat 66px. Keeping the margin on top of the corrected container would push the bar 26px too far down.

marketplace carried its own version of the same patch, a 36px margin with a rule to cancel it while its announcement bar showed, and it is removed in decentraland/marketplace#2688.

### Measured, at 1200px wide

| | Before | After |
| --- | --- | --- |
| Container reserves | 66px | 92px |
| Announcement bar `margin-top` | 26px | 0 |
| Announcement bar top edge | 92px | 92px |

The bar lands in exactly the same place; the 26px it needed now comes from the container instead of from here. No visual change is expected from this PR, which is the point.

### How to test

`npm ci && npm run check:code && npm run build && npm test`, then in the running app at a desktop width:

```js
const nav = document.querySelector('nav')
nav.closest('div').getBoundingClientRect().height // 92, was 66
document.querySelector('aside[class*=_bar_]').getBoundingClientRect().top // still 92
```

Worth a look below 992px too, where the bar is 64px and the container used to over-reserve by 2px.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,7 +44,7 @@
         "decentraland-builder-scripts": "^0.24.0",
         "decentraland-connect": "^12.0.0",
         "decentraland-crypto-fetch": "^2.0.1",
-        "decentraland-dapps": "^29.6.1",
+        "decentraland-dapps": "^29.6.2",
         "decentraland-ecs": "6.12.4-7784644013.commit-f770b3e",
         "decentraland-experiments": "^1.0.2",
         "decentraland-transactions": "^3.0.2",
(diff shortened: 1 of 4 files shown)


braianj:
Both addressed in 30f6e25. The P1 was a real miss on my side, and the same one I made in marketplace: I grepped CSS files for the workaround, found one, and stopped looking. This repo had it twice too.

**P1** — the inline `style={{ marginBottom: isAnnouncementBarVisible ? 0 : 36 }}` in `Navbar.tsx` is gone, and with no styles left the wrapper is a plain `<div>`. Kept as a `<div>` rather than collapsed to a fragment, since the navbar container resolves through it and removing a level changes the DOM shape for no gain here.

**P2** — the `margin-top: 0` in the `max-width: 991px` block is gone.

Measured on this branch at 1200px, both announcement-bar states:

| | Bar showing | Bar dismissed |
| --- | --- | --- |
| Container reserves | 92px | 92px |
| Wrapper `margin-bottom` | 0 | 0 |
| Announcement bar top | 92px | not rendered |
| Gap navbar to bar | 0 | n/a |
| First content top | 138px | 92px |

With the bar dismissed that content top would have been 128px before this commit, which is the 36px of dead space you predicted.

Also swept both this repo and marketplace for any remaining margin near the navbar, announcement bar or page layout, inline or in CSS: nothing left. Re-ran `check:code`, `build` and 120 suites / 1568 tests.

